PR #12, opened 2020-06-17
Add github CI

### LINKED ISSUE #9: Disable direct API tests for non-master merges
Certain sanity tests are ran against the actual discord API to minimize the chance of bugs. While we still consider this a good strategy moving forward, we should minimize those tests to being ran **only before a merge to master**. This comes with a downside of not immediately being able to tell a bug on our part, but makes it many times easier for contributors to provide changes.

**Issue originally posted by @BartArys**

HopeBaron:
* This fixes #9 by enabling rest tests only on merge requests targeting master.
* Add a CI using GitHub Actions.


HopeBaron:
Note that the GitLab issue templates were removed as well as the CI file. We might need to write new templates.

Issue #5, opened 2020-06-16
Stop using java.awt.Color


HopeBaron:
Kord uses `java.awt.Color` to represent colors. This issue suggests moving to our own/a multiplatform implementation.

## Why switch from java's implementation

* People using jlink to reduce the size of the JRE are forced to include the `java.desktop` module for what in our use case is a glorified integer. It comes down to ~30 megabytes which would be nice to shave off.
* We'll move to multiplatform eventually, and will be required to implement a `Color` regardless (or risk depending on an npm module!). We might as well make one that fits all platforms.

## Isn't this outside of Kord's scope

It is, technically, outside of our scope. A color class is not related to the Discord API. A simple `Color(r,g,b)` should be sufficient though. The alternative would be to just accept an integer, which would be a dramatically worse user experience.

**Issue originally made by @BartArys**

PR #20, opened 2020-06-20
Add a common Color implementation #5

HopeBaron:
This PR adds a common color interface for Kord in the favor of MPP


HopeBaron:
On the point. Will do.

common/src/main/kotlin/com/gitlab/kordlib/common/Color.kt
```diff
@@ -9,12 +9,12 @@ class Color(val rgb: Int) {
...
 
     init {
-        if (rgb < 0 || rgb > 0xFFFFFF) throw IllegalArgumentException("Color is invalid.")
+        if (rgb !in 0..0xFFFFFF) throw IllegalArgumentException("Color is invalid.")
     }
 }
 
-
 private fun rgb(red: Int, green: Int, blue: Int): Int {
+    if(red !in 0..255 || green !in 0..255 || blue !in 0..255) throw IllegalArgumentException()
```
BartArys, line 17: This should probably throw a more informative exception, inform what the expected vs actual value is, which specific value was illegal.

PR #23, opened 2020-06-22
Invite core support in Kord

HopeBaron:
* Add support to `Invite`s in the core module.
* Fix incorrect usage of supplier's getGuild for getGuild in `PermissionOverwriteEntity`
* Add `PartialGuild`  to the core

### DIFF
--- a/core/src/main/kotlin/com/gitlab/kordlib/core/cache/data/PartialGuildData.kt
+++ b/core/src/main/kotlin/com/gitlab/kordlib/core/cache/data/PartialGuildData.kt
@@ -1,0 +1,26 @@
+package com.gitlab.kordlib.core.cache.data
+
+import com.gitlab.kordlib.cache.api.data.description
+import com.gitlab.kordlib.common.entity.DiscordPartialGuild
+import com.gitlab.kordlib.common.entity.Permissions
+import kotlinx.serialization.Serializable
+
+@Serializable
+class PartialGuildData(
+        val id: Long,
+        val name: String,
+        val icon: String? = null,
+        val owner: Boolean? = null,
+        val permissions: Permissions? = null
+) {
+    companion object {
+
+        fun from(partialGuild: DiscordPartialGuild) = with(partialGuild) {
+            PartialGuildData(id.toLong(), name, icon, owner, permissions)
+        }
+    }
+
+
+}
+
+val InviteData.guildId get() = guild?.id
--- a/core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
+++ b/core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
@@ -5,11 +5,10 @@ import com.gitlab.kordlib.common.entity.TargetUserType
 import com.gitlab.kordlib.common.exception.RequestException
 import com.gitlab.kordlib.core.Kord
 import com.gitlab.kordlib.core.KordObject
-import com.gitlab.kordlib.core.behavior.GuildBehavior
 import com.gitlab.kordlib.core.behavior.UserBehavior
-import com.gitlab.kordlib.core.behavior.channel.GuildChannelBehavior
+import com.gitlab.kordlib.core.behavior.channel.ChannelBehavior
 import com.gitlab.kordlib.core.cache.data.InviteData
-import com.gitlab.kordlib.core.entity.channel.GuildChannel
+import com.gitlab.kordlib.core.entity.channel.Channel
 import com.gitlab.kordlib.core.exception.EntityNotFoundException
 import com.gitlab.kordlib.core.supplier.EntitySupplier
 import com.gitlab.kordlib.core.supplier.EntitySupplyStrategy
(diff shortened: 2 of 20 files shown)


core/src/main/kotlin/com/gitlab/kordlib/core/cache/data/PartialGuildData.kt
```diff
@@ -0,0 +1,22 @@
...
+@Serializable
+class PartialGuildData(
+        val id: Long,
+        val name: String,
+        val icon: String? = null,
+        val owner: Boolean? = null,
+        val permissions: Permissions? = null
+) {
+    companion object {
+        val description = description(PartialGuildData::id) {
```
BartArys, line 16: Do we plan on storing PartialGuilds in cache?

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
```diff
@@ -108,22 +105,23 @@ data class Invite(
      * @throws [RequestException] if anything went wrong during the request.
      * @throws [EntityNotFoundException] if the [Guild] wasn't present.
      */
-    suspend fun getGuild(): Guild = supplier.getGuild(guildId)
+    suspend fun getGuild(): Guild = supplier.getGuild(guildId!!)
```
BartArys, line 108: I shouldn't this become nullable then? Generally when an id is nullable the `getEntity` variant also is nullable

core/src/main/kotlin/com/gitlab/kordlib/core/supplier/EntitySupplier.kt
```diff
@@ -344,6 +344,11 @@ interface EntitySupplier {
     suspend fun getWebhookWithToken(id: Snowflake, token: String): Webhook =
             getWebhookWithTokenOrNull(id, token) ?: EntityNotFoundException.webhookNotFound(id)
 
+    suspend fun getInviteOrNull(code: String, withCounts: Boolean = true): Invite?
```
BartArys, line 347: We're not caching invites right now, cache shouldn't implement this.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
```diff
@@ -146,4 +144,20 @@ data class Invite(
     override fun withStrategy(strategy: EntitySupplyStrategy<*>): Invite =
             Invite(data, kord, strategy.supply(kord))
 
+    companion object {
+        fun from(invite: InviteResponse) = with(invite) {
```
BartArys, line 148: Pretty sure this doesn't belong here

core/src/test/kotlin/InviteTests.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.jupiter.api.TestInstance
+import kotlin.test.assertEquals
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class InviteTests {
+
+    @Test
+    fun `get invite from rest and cache`() = runBlocking {
+        val kord = Kord(System.getenv("KORD_TEST_TOKEN"))
+        val invite = kord.defaultSupplier.getInvite("mpDQm5N")
```
BartArys, line 19: Should probably store the invite code in an ENV variable instead. Don't want to update our source code because our invite expired or whatever.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
HopeBaron: I'll work on this in my next branch regarding the null assertation.

BartArys: In that case this should throw an `EntityNotFoundException` instead of a KNP

core/src/main/kotlin/com/gitlab/kordlib/core/supplier/EntitySupplier.kt
HopeBaron: Still useful if someone is manually putting them in the cache. 
probably shall yeet docs there.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
HopeBaron: There are various functions that return KNP instead of `EntityNotFoundException`. I'm going to work on those on a separate branch.

core/src/main/kotlin/com/gitlab/kordlib/core/cache/data/PartialGuildData.kt
HopeBaron: We don't receive that from the gateway I made a mistake.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Invite.kt
HopeBaron: Yeah. I'm caching from rest. :sweat_smile:

PR #45, opened 2020-08-19
Bits and bobs

BartArys:
This PR contains some *fairly* small improvements I've worked on the past days:

* Add some more docs, fixed some outdated docs on Kord's builder as well
* Move non-cdn links from discordapp -> discord. The only non-doc change is `Route's` `baseUrl`
* Add some more docs, again
* Fix an issue with the recently introduced error reporting on faulty tokens when building Kord. 
Seems like discord has again introduced some undocumented fields, so I added the flag to ignore those.
* Added a no-op gateway, this one literally does nothing, its eventflow is empty as well.
* Added a shortcut for creating intents without... intents.
* Fix some formatting in the `Util` file.
* Remove the `rest.getSelf` call when building Kord. we can get the ID from the token itself.
* Introduced an experimental REST-only variant of the Kord builder, this one doesn't need to do any API calls when building and
disables gateway and cache related stuff.
* As a result, some code from the builders have been moved towards an internal util file to reduce duplication.
* Change the default dispatcher to `Dispatchers.Default` when building Kord instances.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,23 @@
 
 * `Member#kick` and `Guild#kick` now have an optional reason.
 * `KordBuilder` now throws a (more) useful exception when building a bot with an invalid token.
+* The REST module will now use `discord.com/api` instead of the deprecated `discordapp.com/api`.
+* Kord now uses `Dispatchers.default` as the default dispatcher.
 
 ## Fixes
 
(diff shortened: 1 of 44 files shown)


HopeBaron (approved):
Looks fine. Some of the `Flow` extension functions we defined now already exist in the coroutines library itself. :+1:

Issue #50, opened 2020-08-21
Communicate Unstable API


BartArys:
With Discord's ever changing API we will sometimes have to add or change properties in our own lower level API. 
Since this technically results in us breaking the API, it would mean that staying up to do date with Discord means a lot of minor version bumps. 

Most of the changes don't propagate themselves in a breaking way through our API though. The effects are mostly limited to the serializable classes and entity data classes, which shouldn't really be used by consumers of our API in the first place.

I propose introducing `@KordUnstableApi` to properly inform both Kord developers and users which APIs can and are allowed to change in minor versions.

BartArys:
This feature will not be implemented, see #56 for a detailed explanation.

PR #56, opened 2020-08-26
Annotate unstable API

BartArys:
This is a WIP PR to discuss the current progress on annotating parts of the API that would be considered unstable. #50

While implementing this I encountered some issues:

* In an ideal world, just annotating the constructors of unstable data classes would suffice. 
After all it will only be constructors gaining an extra argument that would cause breaking changes. 
This however isn't the full truth, data classes generate a `copy` method based on the primary constructor.
Changing the constructor would also break the `copy` method and there is no way to annotate generated methods.
This leads to the unfortunate conclusion that the entire class has to be annotated as unstable, even though it
is only the constructor and copy method that could break binary compatibility.

* Kord makes extensive use of inline functions when building requests, this causes the implementation of these 
functions to be copied directly into the user's code. These implementations often involve interacting with the
unstable API. Meaning that unless we mark those parts of our API unstable as well, we will eventually break
user's bytecode without changing our stable API. A long-term solution might be to un-inline our entire API, 
regaining control of this part of the API at the cost of performance.

* Kotlin allows exhaustive checks on enums and sealed classes. A scenario where Discord introduces a new value
to an exhaustive set would result in us adding that value to the enum or sealed class. 
All user code that depends on this set being exhaustive will then have the option to break.
I don't think there's a good long-term solution here that allows us to introduce these new fields to core in a non-breaking way.

Looking at these issues collectively, I don't think there's a easy way for us to negotiate a fast-moving and stable API.
The path with the least headaches might be to keep any breaking change to minor versions after all, 
anything else I fear would turn in too many compromises for too little gain.

### DIFF
--- a/common/src/main/kotlin/com/gitlab/kordlib/common/annotation/Annotations.kt
+++ b/common/src/main/kotlin/com/gitlab/kordlib/common/annotation/Annotations.kt
@@ -51,3 +51,20 @@ annotation class KordExperimental
 @Experimental(level = Experimental.Level.WARNING)
 @Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.TYPEALIAS, AnnotationTarget.PROPERTY)
 annotation class KordUnsafe
+
+/**
+ * Marks a Kord-related API as unstable.
+ *
+ * Kord marks targets as unstable if they are subject to changes outside the development process.
+ * They offer no binary compatibility between versions.
+ *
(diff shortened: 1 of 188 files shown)


HopeBaron:
Looking at the points you provided, I find it easier to look into a way to test the package itself locally?

BartArys:
Reflecting back on the problems encountered while trying to implement this, we've decided that the compromises which would have to be made would make this feature serve in few people's interest. Attempts at implementing this resulted in an API that's neither flexible to update nor enjoyable to use.

As such, moving forward, we will make a best-effort to maintain API stability between patch versions and move Discord-related changes to the minor version changes.

PR #57, opened 2020-08-29
New field builder

HopeBaron:
This introduces a new field builder function that allows empty field values and makes mandatory values such as `name` a part of its parameters.

### DIFF
--- a/rest/src/main/kotlin/com/gitlab/kordlib/rest/builder/message/EmbedBuilder.kt
+++ b/rest/src/main/kotlin/com/gitlab/kordlib/rest/builder/message/EmbedBuilder.kt
@@ -13,9 +13,14 @@ import java.time.format.DateTimeFormatter
  * Inline Markdown links are supported in in all description-like fields.
  */
 @KordDsl
+@Suppress("unused")
 class EmbedBuilder : RequestBuilder<EmbedRequest> {
     companion object {
-        val ZERO_WIDTH_SPACE = "\u200E"
+        /**
+         * Utility default value for fields.
+         * While blank values are not allowed, this value can be used to trick Discord into rendering a Field as empty.
+         */
+        const val ZERO_WIDTH_SPACE = "\u200E"
     }
 
     /**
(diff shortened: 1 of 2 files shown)


rest/src/main/kotlin/com/gitlab/kordlib/rest/builder/message/EmbedBuilder.kt
```diff
@@ -14,6 +14,10 @@ import java.time.format.DateTimeFormatter
  */
 @KordDsl
 class EmbedBuilder : RequestBuilder<EmbedRequest> {
+    companion object {
+        val ZERO_WIDTH_SPACE = "\u200E"
+    }
+
```
BartArys, line 20: 1) can be made a const.

2) Wouldn't this fit better in the `Field` class, as a non-companion value?

I'm thinking in lieu of ease of use and API discovery changing this
```kotlin
field {
    name = title
    value = if(condition) text else EmbedBuilder.ZERO_WIDTH_SPACE
}
```
to this
```kotlin
field {
    name = title
    value = if(condition) text else ZERO_WIDTH_SPACE
}
```
might be beneficial.

rest/src/main/kotlin/com/gitlab/kordlib/rest/builder/message/EmbedBuilder.kt
```diff
@@ -92,6 +96,23 @@ class EmbedBuilder : RequestBuilder<EmbedRequest> {
         fields += Field().apply(builder)
     }
 
+    /**
+     * Adds a new [Field] using the given [name] and [value].
+     *
+     * [inline] defines wither a field should be inlined with the default  value set to false.
```
BartArys, line 102: `wither` -> `whether`, extra space there near the end as well.

This should probably use the `@param` syntax instead, it's kind of in that format already.

rest/src/main/kotlin/com/gitlab/kordlib/rest/builder/message/EmbedBuilder.kt
```diff
@@ -92,6 +96,23 @@ class EmbedBuilder : RequestBuilder<EmbedRequest> {
...
+     *
+     * **[value] defaults to [ZERO_WIDTH_SPACE] which is an empty field value.**
```
BartArys, line 104: Same here, this looks like `@param` syntax to me.

PR #66, opened 2020-09-14
Add reason to member role add/remove

BartArys:


HopeBaron (approved):

core/src/main/kotlin/com/gitlab/kordlib/core/behavior/MemberBehavior.kt
```diff
@@ -71,10 +71,19 @@ interface MemberBehavior : Entity, UserBehavior {
      *
      * @throws [RestRequestException] if something went wrong during the request.
      */
-    suspend fun addRole(roleId: Snowflake) {
+    suspend fun addRole(roleId: Snowflake) {//TODO remove in Kord 0.7.0
```
Goregius, line 74: Add Depreciated annotation?

BartArys: Good point, I forgot about the hidden deprecation

PR #70, opened 2020-09-22
Remove live message events override

AllanWang:
Events come from `kord`, which is the same instance for a given bot. Disallowing overrides prevents us from accidentally skipping the live filter.


BartArys (approved):

BartArys:
merged https://github.com/kordlib/kord/commit/a427a616df1a33883dbfea6dc13ab9e8fde7cc29

PR #73, opened 2020-09-26
Add team field in addiotion to only the teamId to ApplicationInfo

DRSchlaubi:


BartArys:
This feature seems to have accidentally slipped in via another PR already, whoops!
Thanks for the PR either way.

PR #74, opened 2020-09-26
Add premium tier to Guild

DRSchlaubi:
- Add max emote count to Guild
- Update docs for Snowflake


BartArys:
Hey, thanks for the PR! Sorry it took such a long time for us to get to it. 

Before we can merge this, could you rebase your changes on top of 0.7.x?
After you've done that, you'll most likely encounter an error on build telling you the binary API has changed.
If so, run `gradle jar` and then `gradle dumpApi --tag=current` to create a new reference API from your current code, your build should now succeed.

common/src/main/kotlin/com/gitlab/kordlib/common/entity/Snowflake.kt
```diff
@@ -7,12 +7,20 @@ import kotlin.time.toKotlinDuration
 
 /**
  * A unique identifier for entities [used by discord](https://discord.com/developers/docs/reference#snowflakes).
+ *
+ * @property longValue this snowflake represented in a long
```
BartArys, line 11: use `as a Long` instead of  `represented in a long`

common/src/main/kotlin/com/gitlab/kordlib/common/entity/Snowflake.kt
```diff
@@ -7,12 +7,20 @@ import kotlin.time.toKotlinDuration
...
  */
 inline class Snowflake(val longValue: Long) : Comparable<Snowflake> {
     constructor(value: String) : this(value.toLong())
 
-    val value get() = longValue.toString()
+    /**
+     * This snowflake represented in a String.
```
BartArys, line 17: use `as a String` instead of  `represented in a String`

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Guild.kt
```diff
@@ -73,6 +73,29 @@ class Guild(
      */
     val bannerHash: String? get() = data.banner
 
+    /**
+     * The [PremiumTier] from Nitro boosts,
```
BartArys, line 77: `.` instead of `,`

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Guild.kt
```diff
@@ -73,6 +73,29 @@ class Guild(
...
+    /**
+     * Maximal amount of server emotes based on [premiumTier].
+     *
+     * [PremiumTier.Unknown] -> [Int.MIN_VALUE]
+     * [PremiumTier.None] -> 50
+     * [PremiumTier.One] -> 100
+     * [PremiumTier.Two] -> 150
+     * [PremiumTier.Three] -> 250
+     */
+    val maxEmotes: Int
```
BartArys, line 90: This property and the information it conveys aren't documented in the API anywhere. 

I'm aware it's visible from the client, but this does carry the risk of not in any way being stable. 
Discord has already taken liberties in changing the perks related to boost levels on a whim during the recent pandemic (not in negative ways, mind you), so I'm extra nervous of this becoming outdated *very* quickly, which makes it a lot less useful for users.

BartArys (changes requested):

core/src/main/kotlin/com/gitlab/kordlib/core/entity/Guild.kt
BartArys: Thinking about this some more, I'm not certain this is the right place to put this property either. If it is completely dependent on `PremiumTier` and not the `Guild` object, it should be a property of `PremiumTier`.

I'm not keen on using magic numbers for `Unknown` either. If we do not know the value for all tiers, the property should be nullable instead.

DRSchlaubi: What about adding it to PremiumTier and adding some annotation on top of it to mark it as not documented by discord?
Or just remove it completely?

BartArys: If we're adding it, it should definitely end up in `PremiumTier`. 

I'm not keen on marking it with any kind of annotation though, API annotated with Kord annotations have a goal of becoming stable in the long run. Something like this can never become stable until Discord documents it as such, so I'm not sure what we would be conveying through it.

Looking further into this, there *is* some form of [official documentation](https://support.discord.com/hc/en-us/articles/360028038352-Server-Boosting-) on the matter, which makes me not completely opposed to adding this anymore. Linking to this in the documentation of the property/enum would go a long way.

I'm not sure how @HopeBaron stands on this issue.

I guess the more general question is if we should aim to provide data that is only present in the client UI. If that data has value for the bot (which I'd argue is the case here), then I guess I should be fine with that, we'll take on the burden of keeping it up to date.

So the recap, I'm okay with adding the property to `PremiumTier` if the link gets added to the official docs.

HopeBaron:
Since all the data we know is constant and not won't be changed unless Discord changes them I'd say that those should be a part of the enum class rather than their own objects. I don't see any exact values/numbers. Regarding the added properties except the tier numbers in official docs, I couldn't find any except the article @BartArys linked. I'm fine as well if the link is added

DRSchlaubi:
These are objects because @BartArys told me to make it a sealed class

common/src/main/kotlin/com/gitlab/kordlib/common/entity/DiscordGuild.kt
```diff
@@ -234,28 +237,60 @@ data class DiscordVoiceState(
...
+    abstract val level: Int?
+    abstract val maxEmotes: Int?
+
     /** The default code for unknown values. */
-    Unknown(Int.MIN_VALUE),
-    None(0),
-    One(1),
-    Two(2),
-    Three(3);
+    object Unknown : PremiumTier() {
```
BartArys, line 252: I've come to the realization that every premium tier unknown to us would have a `level`, so making it `null` is nonsensical.
Since we know the level from the decoder, we can turn `Unknown` into a class with the level in its constructor instead, and remove the nullable type marker from the parent sealed class.

BartArys (changes requested):

common/src/main/kotlin/com/gitlab/kordlib/common/entity/DiscordGuild.kt
DRSchlaubi: What about the maxEmotes then? surely we won't have access to them

BartArys: Indeed, those should stay nullable.

BartArys (approved):

PR #79, opened 2020-09-28
Improve Kord.on

BartArys:
Provides a more detailed documentation on the behaviour of `Kord.on`, also changes the behaviour to run consumers in parallel and retry infinitely (not sure what even could throw) on errors.

### DIFF
--- a/core/src/main/kotlin/com/gitlab/kordlib/core/Kord.kt
+++ b/core/src/main/kotlin/com/gitlab/kordlib/core/Kord.kt
@@ -189,7 +189,7 @@ class Kord(
  * The events are buffered in an [unlimited][CoroutineChannel.UNLIMITED] [buffer][Flow.buffer] and
  * [launched][CoroutineScope.launch] in the supplied [scope], which is [Kord] by default.
  * Each event will be [launched][CoroutineScope.launch] inside the [scope] separately and
- * any thrown [Throwable] will be caught and logged. The flow will [retry][Flow.retry] infinitely on error.
+ * any thrown [Throwable] will be caught and logged.
  *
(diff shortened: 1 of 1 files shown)


AllanWang:
Should retry be infinite? I imagine we can get a reasonable max count (5?), after which the consumer should be ignored. Or if not, perhaps there should be a configurable max count

BartArys:
> Should retry be infinite? I imagine we can get a reasonable max count (5?), after which the consumer should be ignored. Or if not, perhaps there should be a configurable max count

To be fair, I don't think this code can even get to the retry, seeing as the only action that *could* throw is already caught.

On the topic of of defaults; any number would seem arbitrary to me. Is there any reason to prefer 5 over 4 or 3 or any other number?

As for configurability; `Kord#on` was originally introduced as a macro for quick prototyping, and as such not meant to be configurable. I fear the community might have adopted it differently though, especially those who were (initially) unfamiliar with kotlinx.coroutines.

AllanWang:
What I meant was that if kord had a max retry field, the extension could use it. No need to add optional args to the method.

Code wise, I don't think it makes sense for any request to potentially loop forever. If we think that retry won't even be called, it would be better to not add it in the first place. If it is callable, please considering adding max loop count, even if it's a very generous number.

BartArys:
I think there's a slight misunderstanding of what retry does here. If a flow operation throws without any error handling this results in the flow halting, and in this situation it would halt silently (safe for a log message). `retry` will intercept the error and attempt to consume the flow again. 
Since this is a hot flow it will drop the error-causing event and continue on processing new events. It won't try and handle an event more than once.

That said, it is unnecessary here, so I'll opt to remove it instead.

PR #88, opened 2020-10-14
Provide meaningful toString

BartArys:
This PR aims to add meaningful toString implementations for core classes to facilitate debugging. Most of these follow the data class-like approach of `ClassName(p1Name=p1Value, p2Name=p2Value, ...)`. 
Exceptions to these are behaviors (since they don't have constructors), who instead display the Snowflake IDs they are made up of,
and the bitshift classes like Permissions; those display their enum values instead.


HopeBaron (commented):
:+1:

PR #90, opened 2020-10-17
Fix a lot of compiler warnings

BartArys:
This PR went through the code and fixed (almost) all compiler warnings. Most notably:

* warnings originating from newly experimentalized(?) APIs from kotlinx.serialization, this mostly comes down to `@Serializer` being made experimental. Fixing these involved removing the annotation and referring to the custom serializers is `@Serializable`.

* warnings originating from shadowed variable names. Fixing this mostly meant a rename of variables.

* warnings in Test, these were fixed by suppressing/opting in to them for the most part. We don't expect a stable API in our tests after all.

* warnings due to using outdated `@Experimental` API, these were changed to use `@RequiresOptIn` instead.

A unique case was `GuildBehavior#getVanityUrl` which erroneously assumed the `code` from the REST response to be nullable.
In actuality, the HTTP Request will return with an error code and a specific `JsonErrorCode`. In an attempt to appease to Kord's contract of returning the invite URL or null if not present, I decided to change the `RestRequestException` to include the `DiscordErrorResponse` containing said `JsonErrorCode` (see e1b12a9). In an attempt to keep some semblance of backwards-compatible migration I decided to keep the old fields as deprecated and introduce a `DeprecatedSinceKord` annotation to inform when a field was deprecated. This to ease the next step into 0.7.x which will be to remove previously deprecated parts of our API.

Following this, I used the new fields in the exception to return null on the `InviteCodeInvalidOrTaken` error code (see 38b7a81).

Fixing this bug wasn't the intention of the PR, but since I was touching the code anyway I though I'd get that out of the way with this PR as well.


core/src/main/kotlin/com/gitlab/kordlib/core/Util.kt
```diff
@@ -48,11 +49,29 @@ internal inline fun <T> catchNotFound(block: () -> T): T?  {
...
+@OptIn(ExperimentalContracts::class)
+internal inline fun <T> catchDiscordError(vararg codes: JsonErrorCode, block: () -> T): T?  {
+    contract {
+        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
+    }
+    return try {
+        block()
+    } catch (exception: RestRequestException) {
+        when {
+            codes.isEmpty() -> null
```
HopeBaron, line 66: if codes is empty then we are not supposed to catch any json error.

common/src/test/kotlin/json/MessageTest.kt
```diff
@@ -76,7 +76,7 @@ fun `User serialization`() {
         id shouldBe "334385199974967042"
         pinned shouldBe false
         editedTimestamp shouldBe null
-        author?.let {
+        author.let {
```
HopeBaron, line 79: With `author` being null-safe there is no need for the let scope.

rest/src/main/kotlin/com/gitlab/kordlib/rest/request/HttpUtils.kt
```diff
@@ -50,8 +50,8 @@ fun HttpResponse.logString(body: String) = "[RESPONSE]:${status.value}:${call.re
 
 suspend fun HttpResponse.errorString(): String {
     val message = String(this.readBytes())
-    return if (message.isBlank()) "${logString(message)}"
-    else "${logString(message)}"
+    return if (message.isBlank()) logString(message)
```
HopeBaron, line 53: we can just return the log without checking the message content can't we?

core/src/main/kotlin/com/gitlab/kordlib/core/Util.kt
BartArys: That would make the function effectively useless though. 
I rationalized that not providing any codes implied catching them all, but I see how that is counter-intuitive. Ideally I'd have the parameter default to all enums (since that is what it implied), but varargs cannot have default values.

common/src/test/kotlin/json/MessageTest.kt
BartArys: Sure, I can turn that into a `with` statement instead.

rest/src/main/kotlin/com/gitlab/kordlib/rest/request/HttpUtils.kt
BartArys: yeah, this seems to be the result of simplifying a condition to the point it is no longer needed 😅

core/src/main/kotlin/com/gitlab/kordlib/core/Util.kt
HopeBaron: It's fine like this :+1: unless it's used we want to catch something we are not going to use this

HopeBaron (approved):

PR #96, opened 2020-10-24
Fix everyone mention

### LINKED ISSUE #94: Trying to mention the everyone role results in a malformed @@everyone mention in the client
```kotlin
channel.createMessage(guild.everyoneRole.mention)
```

results in:
`@@everyone`

This behaviour on is correct according the official docs, but it seems like the client stumbles over the special everyone type. Using the literal `@everyone` works fine though.

Proposed solution:
Introduce a check on `RoleBehavior#mention` that checks for `guildId.value == id.value` and returns the literal `@everyone` on true instead.

mrbarrientosg:
Fix #94

### DIFF
--- a/core/src/main/kotlin/com/gitlab/kordlib/core/behavior/RoleBehavior.kt
+++ b/core/src/main/kotlin/com/gitlab/kordlib/core/behavior/RoleBehavior.kt
@@ -37,7 +37,12 @@ interface RoleBehavior : Entity, Strategizable {
     /**
      * The raw mention of this entity.
      */
-    val mention: String get() = "<@&${id.value}>"
+    val mention: String get() {
+        return when(guildId.value == id.value) {
+            true -> "@everyone"
+            false -> "<@&${id.value}>"
+        }
+    }
 
     /**
      * Requests to change the [position] of this role.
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/com/gitlab/kordlib/core/behavior/RoleBehavior.kt
```diff
@@ -37,7 +37,12 @@ interface RoleBehavior : Entity, Strategizable {
     /**
      * The raw mention of this entity.
      */
-    val mention: String get() = "<@&${id.value}>"
+    val mention: String get() {
+        return when(guildId.value == id.value) {
+            true -> "@everyone"
+            false -> "<@&${id.value}>"
+        }
```
BartArys, line 44: Prefer using `if/else` for binary conditions: https://kotlinlang.org/docs/reference/coding-conventions.html#if-versus-when

```suggestion
        return if(guildId.value == id.value) "@everyone"
        else "<@&${id.value}>"
```

BartArys (changes requested):
Thanks for taking your time to fix this issue!
One small nitpick as per the Kotlin coding conventions.

core/src/main/kotlin/com/gitlab/kordlib/core/behavior/RoleBehavior.kt
mrbarrientosg: Hi, thanks you for the tip, I'm going to consider it next time :)

BartArys (approved):

PR #98, opened 2020-10-26
Add support to resolve cdn image to given resolution.

### LINKED ISSUE #97: Add library support to resolve cdn image to given resolution
As [documented in the docs](https://discord.com/developers/docs/reference#image-formatting), it is possible to request different resolution of the Images sent by discord by appending a get parameter `?size=desired_size` where desired_size can be between 16 and 4096 (under power of two like 128, 256, 512, etc.).

This issue is created to address this enchantment in the API.

Proposed solution:
Introduce a new enum class similar to Format under Image, which specify the image resolution.

Animeshz:
Fix #97


core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
```diff
@@ -91,22 +87,31 @@ open class User(
...
-            } ?: defaultUrl
+                isAnimated -> Image.Format.GIF
+                else -> Image.Format.PNG
+            }
+
+        /**
+         * Gets the avatar url in supported format with default behavior of [format] and size [Image.Size.SIZE_128].
+         */
+        val url: String
+            get() = getUrl()!!
```
BartArys, line 106: Consider solving the nullability issue without throwing an NPE.

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
```diff
@@ -18,19 +18,42 @@ class Image private constructor(val data: ByteArray, val format: Format) {
...
 
-            Image(bytes, Format.fromContentType(contentType))
+            Image(bytes, Format.fromContentType(contentType), size)
+        }
+    }
+
+    /**
+     * Represents size of the Image as specified by
+     */
+    enum class Size(val resolution: String) {
```
BartArys, line 41: Why represent resolution as a String if the values can only be integers?

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
```diff
@@ -18,19 +18,42 @@ class Image private constructor(val data: ByteArray, val format: Format) {
...
+        SIZE_16("16"),
+        SIZE_32("32"),
+        SIZE_64("64"),
+        SIZE_128("128"),
+        SIZE_256("256"),
+        SIZE_512("512"),
+        SIZE_1024("1024"),
+        SIZE_2048("2048"),
+        SIZE_4096("4096");
```
BartArys, line 50: We generally prefer PascalCase naming for enum and sealed class constants. `Size16` etc would be more in line with the project's conventions.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
```diff
@@ -91,22 +87,31 @@ open class User(
...
 
         /**
-         * Gets the avatar url in a supported [Image.Format], or returns null if the [format] is not supported.
+         * Gets the avatar url in given [format] and [size], or returns null if the [format] is not supported.
+         *
+         * @param format    The requested image format, defaults to the behavior of [Avatar.format] if not specified.
+         * @param size      The requested image resolution, defaults to [Image.Size.SIZE_128] if not specified.
          */
-        fun getUrl(format: Image.Format): String? {
+        fun getUrl(format: Image.Format = this.format, size: Image.Size = Image.Size.SIZE_128): String? {
```
BartArys, line 114: Is mandating a size a good idea? The endpoint seems to works fine without it as well.

We've run into a little issue with nullability here, if `format = this.format` then we can guarantee the return type to not be null. I wonder if a separate overload where only `size` could be specified would be a solution to this.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
```diff
@@ -115,25 +120,21 @@ open class User(
...
+        suspend fun getImage(format: Image.Format = this.format, size: Image.Size = Image.Size.SIZE_128): Image? {
+            val url = getUrl(format, size) ?: return null
 
             return Image.fromUrl(kord.resources.httpClient, url)
         }
 
-        /**
-         * Requests to get the avatar of the user as an [Image], prioritizing gif for animated avatars and png for others.
-         */
-        suspend fun getImage(): Image = Image.fromUrl(kord.resources.httpClient, url)
```
BartArys, line 131: Why remove this?

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
```diff
@@ -115,25 +120,21 @@ open class User(
...
+         * Requests to get the avatar of the user as an [Image] given [format] and [size], or returns null if the
+         * format is not supported.
          *
-         * @param format the requested image format, defaults to the behavior of [url] if null.
+         * @param format    The requested image format, defaults to the behavior of [format] if not specified.
+         * @param size      The requested image resolution, defaults to [Image.Size.SIZE_128] if not specified.
          */
-        suspend fun getImage(format: Image.Format): Image? {
-            val url = getUrl(format) ?: return null
+        suspend fun getImage(format: Image.Format = this.format, size: Image.Size = Image.Size.SIZE_128): Image? {
```
BartArys, line 133: Same here, should size be a mandatory argument?

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
```diff
@@ -18,19 +18,42 @@ class Image private constructor(val data: ByteArray, val format: Format) {
...
     companion object {
-        fun raw(data: ByteArray, format: Format): Image {
-            return Image(data, format)
+        fun raw(data: ByteArray, format: Format, size: Size): Image {
+            return Image(data, format, size)
         }
 
         suspend fun fromUrl(client: HttpClient, url: String): Image = with(Dispatchers.IO) {
             val call = client.request<HttpResponse>(url) { method = HttpMethod.Get }
+            val size = Size.fromResolution(call.request.url.parameters["size"])
```
BartArys, line 27: If the URL does not contain a size (which is possible), this function would end up lying about the actual size of the image.

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
```diff
@@ -18,19 +18,42 @@ class Image private constructor(val data: ByteArray, val format: Format) {
...
             @Suppress("EXPERIMENTAL_API_USAGE")
             val bytes = call.content.toByteArray()
 
-            Image(bytes, Format.fromContentType(contentType))
+            Image(bytes, Format.fromContentType(contentType), size)
+        }
+    }
+
+    /**
+     * Represents size of the Image as specified by
```
BartArys, line 39: Unfinished documentation?

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
```diff
@@ -18,19 +18,42 @@ class Image private constructor(val data: ByteArray, val format: Format) {
...
+        SIZE_512("512"),
+        SIZE_1024("1024"),
+        SIZE_2048("2048"),
+        SIZE_4096("4096");
+
+        companion object {
+            fun fromResolution(resolution: String?): Size {
+                return if (resolution == null) SIZE_128
+                else values().first { it.resolution == resolution }
+            }
```
BartArys, line 56: * Lack of documentation
* Why was 128 chosen as a default?

```kotlin
values().first { it.resolution == resolution }
```
This will throw an exception when `resolution` doesn't match any of the values.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
```diff
@@ -91,22 +87,31 @@ open class User(
...
             get() = when {
-                isAnimated -> getUrl(Image.Format.GIF)
-                else -> getUrl(Image.Format.PNG)
-            } ?: defaultUrl
+                isAnimated -> Image.Format.GIF
+                else -> Image.Format.PNG
+            }
+
+        /**
+         * Gets the avatar url in supported format with default behavior of [format] and size [Image.Size.SIZE_128].
```
BartArys, line 103: `in supported format` -> `in a supported format`

BartArys (changes requested):
Thank you for taking your time to create this issue and fix it as well. 💯 
There's some minor issues that need changing and some clarifications that need to be addressed.

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
Animeshz: Hi, I thought the same at first 😅, but everytime we need to encode the url it may consume time because Int to String conversion isn't really cheap (as process goes through creating a byte buffer, then processing each digit to char values, and then copying the buffer to String).

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
Animeshz: 1. Sorry for missing docs, I'll update it.

2. When no resolution is specified in the url, discord sends us 128x128 image (or smaller if original res is smaller) by default. So to be explicit if resolution was not set in the request, we can put 128 by default.
*Note that, response resolution is always equal to or smaller than requested resolution i.e. if original res of image was 60x60 then even if we request 128, 256, or even 4096 we'll receive 60x60.*

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
Animeshz: We can't guarentee the actual size, discord send us size <= desired size. And when not specified it uses desired size being equal to 128. We've handled the null case in Size.fromResolution().

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
Animeshz: Size when not specified discord send us by assuming desired_size = 128.

The default value of format when not specified is this.format, but user can pass its desired format as argument :)

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
Animeshz: Satisfied by optional argument of format, in the function above. So no extra need of method overload.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
Animeshz: Url won't be null with default values of format and size. (default behavior of Avatar.format will return GIF or PNG and default value of size 128)

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
BartArys: 1) While I'm glad to hear you take performance to such a high standard, this sounds like a case of premature optimization. I doubt concatenating ints to strings for Images will be in anyone's hot loop when using Kord.

2) Your  explanation isn't correct. String concatenation in Kotlin is delegated to a `Stringbuilder`, which uses [Integer.getChars](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Integer.java#L486-L518) to provide a zero-garbage implementation for adding numbers to its internal byte array.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
BartArys: > Size when not specified discord send us by assuming desired_size = 128.

I can't find any documented behavior for this. Additionally, this seems to not be correct.
To give some examples: emojis have a default size of 64, while splash screen have a default width of 128 but a non-standard height and are rendered at size 512 by default.

> The default value of format when not specified is this.format, but user can pass its desired format as argument 

What I was trying to convey is that cases where `format` is the default format but sizes are different should result in a non-null `String` return. 
`avatar.getUrl(size = Image.Size.SIZE_128)` should not return a nullable String, 
while `avatar.getUrl(format = Format.GIF, size = Image.Size.SIZE_128)` should return a nullable String.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
BartArys: It is not, this function's return type is not nullable, while the one with optional arguments is.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
BartArys: Apologies, The examples do not apply here since this is purely about User's avatars.
The point about it not being documented behavior stands though.

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
BartArys: > And when not specified it uses desired size being equal to 128.

This is exactly the problem. If someone downloads a discord emote from a url they will end up with an image of 64x64, yet the size property will be 128x128.

Note that an `Image` is not necessarily a Discord image either, this class can also be used to upload images to Discord, so irregular file image sizes should also be supported.

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
```diff
@@ -61,7 +53,7 @@ class Image private constructor(val data: ByteArray, val format: Format) {
 
     /**
      * Represents size of the [Image], for requesting different sizes from the discord.
```
BartArys, line 55: * Unfinished sentence.
* Prefer capitalizing Discord.

core/src/main/kotlin/com/gitlab/kordlib/core/entity/User.kt
```diff
@@ -130,7 +130,7 @@ open class User(
          * Gets the avatar url in given [format] and [size], or returns null if the [format] is not supported.
          *
          * @param format    The requested image format.
-         * @param size      The requested image resolution.
+         * @param size      The requested image size.
```
BartArys: I approved this before, but don't do this: 
https://kotlinlang.org/docs/reference/coding-conventions.html#documentation-comments
> Generally, avoid using @param and @return tags. Instead, incorporate the description of parameters and return values directly into the documentation comment, and add links to parameters wherever they are mentioned. Use @param and @return only when a lengthy description is required which doesn't fit into the flow of the main text.

https://kotlinlang.org/docs/reference/coding-conventions.html#horizontal-whitespace
> As a general rule, avoid horizontal alignment of any kind. Renaming an identifier to a name with a different length should not affect the formatting of either the declaration or any of the usages.

BartArys (changes requested):

rest/src/main/kotlin/com/gitlab/kordlib/rest/Image.kt
Animeshz: [discord.py sets default of 1024 :P](https://github.com/Rapptz/discord.py/blob/master/discord/user.py#L150), but as I observed (multiple avatars that were uploaded at different res) discord send us 128x128 by default so I just set it.

Ouch I still didn't removed this function (even if it is not used anywhere right now), sorry will do 👍

PR #99, opened 2020-10-28
Fix multipart message not sending json

BartArys:
Fixes multipart messages not sending the json body.

### DIFF
--- a/rest/src/main/kotlin/com/gitlab/kordlib/rest/request/KtorRequestHandler.kt
+++ b/rest/src/main/kotlin/com/gitlab/kordlib/rest/request/KtorRequestHandler.kt
@@ -1,20 +1,16 @@
 package com.gitlab.kordlib.rest.request
 
 import com.gitlab.kordlib.rest.ratelimit.*
-import io.ktor.client.HttpClient
-import io.ktor.client.engine.cio.CIO
-import io.ktor.client.features.defaultRequest
-import io.ktor.client.request.forms.MultiPartFormDataContent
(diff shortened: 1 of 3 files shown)


rest/src/main/kotlin/com/gitlab/kordlib/rest/request/KtorRequestHandler.kt
```diff
@@ -80,20 +76,25 @@ class KtorRequestHandler(
...
-            when (request) {
-                is MultipartRequest<*, *> -> {
-                    headers.append("payload_json", body)
-                    this.body = MultiPartFormDataContent(request.data)
+        when (request) {
+            is JsonRequest -> {
+                request.body?.let {
+                    val json = parser.encodeToString(it.strategy, it.body)
+                    logger.debug { request.logString(json) }
+                    this.body = TextContent(json, io.ktor.http.ContentType.Application.Json)
```
HopeBaron, line 84: We can do an early return here and remove the let scope, makes it easier to read. But we can change this later

HopeBaron (approved):

rest/src/main/kotlin/com/gitlab/kordlib/rest/request/KtorRequestHandler.kt
BartArys: I decided against it initially because you can't return to a when, and returning out of the method early seemed like a source of bugs when we're nested this deeply. I could wrap this branch in a `run` though and return to that.

PR #108, opened 2020-11-16
Update entities to V8

BartArys:
Updates Common entities, Gateway events and REST builders/responses to V8 and adds support for optionals.

* All entities have been moved to the common module, irrespective of their need in both modules. This to prevent issues where a gateway/rest entity would suddenly also be present in the other module after Discord changes their API.
* Reïmplemented Audit logs to be a bit more user friendly to interact with.
* Added new REST endpoints and modified those that changed. Provided a migratory `@Deprecated` where possible.
* Added the ability to create a Snowflake from an `Instant`. Useful when querying to before/after positions .
* Added delegates for optionals used in request builders.
* Fixed some minor issues along the way
* Removed the `lateinit` pattern from requests, these are now supplied as mandatory arguments instead i.e. `guild.createTextChannel { name = "mandatory" }` -> `guild.createTextChannel("mandatory") { }`. They are still configurable in the DSL itself. This just helps prevent accidental exceptions when users aren't informed of which fields are mandatory.


common/src/main/kotlin/com/gitlab/kordlib/common/entity/DiscordChannel.kt
```diff
@@ -1,75 +1,151 @@
...
+            get() = PrimitiveSerialDescriptor("type", PrimitiveKind.INT)
+
+        override fun deserialize(decoder: Decoder): ChannelType = when (val code = decoder.decodeInt()) {
+            0 -> GuildText
+            1 -> DM
+            2 -> GuildVoice
+            3 -> GroupDM
+            4 -> GuildCategory
+            5 -> GuildNews
+            96 -> GuildStore
```
HopeBaron, line 108: 96 should be 6

common/src/main/kotlin/com/gitlab/kordlib/common/entity/Presence.kt
```diff
@@ -1,62 +1,92 @@
...
 
-                override fun deserialize(decoder: Decoder): Status {
-                        val name = decoder.decodeString()
-                        return values().firstOrNull { it.name.toLowerCase() == name } ?: Unknown
-                }
+        override fun deserialize(decoder: Decoder): PresenceStatus = when (val value = decoder.decodeString()) {
+            "online" -> Online
+            "idle" -> Idle
+            "dnd" -> DoNotDisturb
+            else -> Unknown(value)
```
HopeBaron, line 85: invisble and offline are missing

HopeBaron (changes requested):

common/src/main/kotlin/com/gitlab/kordlib/common/entity/Presence.kt
BartArys: Technically so. Yet the gateway will not emit presences for those who appear 'offline' and does not reveal invisible statuses either. I'll add them for completeness though.

HopeBaron (approved):

PR #113, opened 2020-11-20
Readd values in intents

HopeBaron:
This adds back the values() function removed due to changing intents to objects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## Additions
 
-* Re-added `values()` in `Intent` to replace enum values 
-* Re-added `intents` in `Intents` to get of `Intents`
+* Re-added `values`  in `Intent` and `Intents`.
 
 # 0.7.0-M1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.7.0-M2
+
+## Additions
+
+* Re-added `values()` in `Intent` to replace enum values 
+* Re-added `intents` in `Intents` to get of `Intents`
+
 # 0.7.0-M1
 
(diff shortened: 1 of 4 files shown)


gateway/src/main/kotlin/com/gitlab/kordlib/gateway/Intent.kt
```diff
@@ -148,14 +149,38 @@ sealed class Intent(val code: DiscordBitSet) {
...
  * A set of [intents][Intent] to be used while [identifying][Identify] a [Gateway] connection to communicate the events the client wishes to receive.
  */
 @Serializable(with = IntentsSerializer::class)
 data class Intents internal constructor(val code: DiscordBitSet) {
-
+    /**
+     *  Returns this [Intents] as a [Set] of [Intent]
+     */
+    @OptIn(PrivilegedIntent::class)
+    val intents = Intent.values().filter { it.code in code }.toSet()
```
BartArys, line 183: Don't make this a backing field property. I expect people to not interact with this all too much.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.7.0-M2
+
+## Additions
+
+* Re-added `values()` in `Intent` to replace enum values 
+* Re-added `intents` in `Intents` to get of `Intents`
```
BartArys, line 6: https://en.wikipedia.org/wiki/Semantic_satiation

BartArys (changes requested):

CHANGELOG.md
HopeBaron: Ruined cause of Kotlin markdowns I'd say :P

BartArys (approved):

PR #114, opened 2020-11-22
Fix Color conversion from java.awt.Color.

Animeshz:
PR include:
  * Fix java.awt.Color to com.gitlab.kordlib.common.Color conversion (essentially dropping alpha values).
  * Change the order of color class string representation (from `.toString()`) in rgb standard order.
  * Added test for it.

### DIFF
--- a/common/src/main/kotlin/com/gitlab/kordlib/common/Color.kt
+++ b/common/src/main/kotlin/com/gitlab/kordlib/common/Color.kt
@@ -10,18 +10,20 @@ import kotlinx.serialization.encoding.Encoder
 
 
 @Serializable(with = Color.Serializer::class)
-class Color(val rgb: Int) {
+class Color(rgb: Int) {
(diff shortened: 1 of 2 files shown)


common/src/main/kotlin/com/gitlab/kordlib/common/Color.kt
```diff
@@ -39,4 +39,4 @@ private fun rgb(red: Int, green: Int, blue: Int): Int {
             (blue and 0xFF) shl 0
 }
 
-val java.awt.Color.kColor get() = Color(rgb)
\ No newline at end of file
+val java.awt.Color.kColor get() = Color(rgb and 0x00FFFFFF)
```
HopeBaron, line 42: This should be a part of the `rgb` function as we expect the returned color to have no alpha for both the conversion method and the constructor.

PR #118, opened 2020-11-23
Fix type coercion in cache queries

BartArys:


core/src/main/kotlin/com/gitlab/kordlib/core/behavior/GuildBehavior.kt
```diff
@@ -149,7 +150,7 @@ interface GuildBehavior : Entity, Strategizable {
      */
     val voiceStates: Flow<VoiceState>
         get() = kord.cache
-                .query<VoiceStateData> { VoiceStateData::guildId eq id.value }
+                .query<VoiceStateData> { idEq( VoiceStateData::guildI, id.value) }
```
HopeBaron, line 153: This is a typo

HopeBaron (changes requested):

core/src/main/kotlin/com/gitlab/kordlib/core/behavior/GuildBehavior.kt
BartArys: There were a few yes. Seems that I refactored faster than IDE analysis tool could display the errors 😅

HopeBaron (approved):

PR #119, opened 2020-11-28
Add missing features

BartArys:
Adds some features we missed when implementing V8 (stickers and role tags). 
As well as some stuff from Discord PR's that are a couple of months old but never merged. Reference to the are in the commit descriptions.


BartArys:
Once this gets merged/okayed I'll add an issue to keep track of Discord PR's that aren't merged yet. 
Most of them aren't old enough yet to warrant implementing now without green-lighting from Discord though, but it beats having to go through the entire PR/issue list every time to find out what we're missing.

HopeBaron (approved):

PR #127, opened 2020-11-30
Fixed getting user's avatar url

Noituri:
User.avatar.url wasn't returning a proper avatar url.
It looked like so:
`https://cdn.discordapp.com/avatars/Snowflake(value=246604909451935745)/2ea115febfda75a80ae0a7e81fd9adea.png`


Scotsguy:
toString strikes again

PR #135, opened 2020-12-06
Fix nullability in builders

BartArys:
Fixes a bunch of nullability issues in the builders that were missed during the original conversion.

### DIFF
--- a/rest/src/main/kotlin/json/request/GuildRequests.kt
+++ b/rest/src/main/kotlin/json/request/GuildRequests.kt
@@ -56,13 +56,13 @@ data class GuildChannelCreateRequest(
 @Serializable
 data class ChannelPositionSwapRequest(
         val id: Snowflake,
-        val position: Int?,
+        val position: OptionalInt? = OptionalInt.Missing,
         @KordExperimental
         @SerialName("lock_permissions")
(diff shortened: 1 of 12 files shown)


HopeBaron (approved):

rest/src/main/kotlin/json/request/GuildRequests.kt
```diff
@@ -56,13 +56,13 @@ data class GuildChannelCreateRequest(
 @Serializable
 data class ChannelPositionSwapRequest(
         val id: Snowflake,
-        val position: Int?,
+        val position: OptionalInt? = OptionalInt.Missing,
         @KordExperimental
         @SerialName("lock_permissions")
-        val lockPermissions: Boolean?,
+        val lockPermissions: Boolean? = null,
```
HopeBaron, line 62: `= null` signifies that the given value is optional
is it optional and nullable 
or just nullable?

BartArys: This change was in reference to this commit: https://github.com/discord/discord-api-docs/pull/1776/commits/01063b10e4f75db5ab9b45a2509483183981e36c

so the type is just nullable, providing defaults was a mistake on my part

PR #141, opened 2020-12-17
Add welcome screen

HopeBaron:


common/src/main/kotlin/entity/DiscordGuild.kt
```diff
@@ -535,3 +539,21 @@ sealed class VerificationLevel(val value: Int) {
...
     }
 }
+
+@Serializable
+data class WelcomeScreenChannel(
+    @SerialName("channel_id")
+    val channelId: Snowflake,
+    val description: String,
+    @SerialName("emoji_id")
+    val emojiId: String?,
```
BartArys, line 549: Isn't this a Snowflake?

core/src/main/kotlin/entity/Guild.kt
```diff
@@ -71,6 +71,11 @@ class Guild(
      */
     val approximatePresenceCount: Int? get() = data.approximatePresenceCount.value
 
+    /**
+     * The welcome screen of a Community guild, shown to new members.
+     */
+    val welcomeScreen: WelcomeScreen? get() = data.welcomeScreen.value
```
BartArys, line 77: Should we really be returning a JSON object here? It's not how we usually do this

BartArys (changes requested):

core/src/main/kotlin/entity/Guild.kt
HopeBaron: right I'll have to add a core representation of this now that it also has some sort of behavior

common/src/main/kotlin/entity/DiscordGuild.kt
HopeBaron: Correct

BartArys (approved):

PR #150, opened 2021-01-03
Fix kord.kord -> kord

AllanWang:


HopeBaron:
This doesn't look correct. Please make sure you use `0.7.x` for modifications my merge commits shouldn't be there.

PR #154, opened 2021-01-04
Add Sonatype publishing

BartArys:


.github/workflows/deployment-ci.yml
```diff
@@ -1,29 +1,81 @@
...
-  release:
-    types: [created]
+  push:
+    paths:
+      - '**.gradle'
+      - '**.properties'
+      - '**/src/**'
+    branches:
+      - 'master'
+      - '0.7.X'
```
HopeBaron, line 13: These should have small x's right?

BartArys: ah, right

HopeBaron (approved):

PR #155, opened 2021-01-04
Fix WebhookType.Incoming

MrPowerGamerBR:
Incoming is actually type `1`, not type `0` 😊

Found this out while trying to retrieve webhooks in my own server, but all webhooks were showing up as "Unknown"


MrPowerGamerBR:
Looks like I used a incorrect branch and trying to change the branch just fucks stuff up, so I will close this PR and open it again in the correct branch

PR #162, opened 2021-01-10
Add support for slash commands option errors

HopeBaron:


rest/src/main/kotlin/json/response/DiscordErrorResponse.kt
```diff
@@ -1,33 +1,73 @@
...
  */
-@Serializable
+@Serializable(DiscordFieldError.Serializer::class)
 class DiscordFieldError(
-        @SerialName("_errors")
-        val errors: List<DiscordErrorDetail>,
-)
+    @SerialName("_errors")
+    val errors: List<DiscordErrorDetail> = emptyList(),
+    val optionErrors: List<Map<String, DiscordFieldError>> = emptyList()
```
BartArys, line 35: I don't think there's a field consistently named `options`.
What I think is going on is that Discord provides the name of the failed property as a key, but I'm not certain, did you check this on multiple invalid structures?

HopeBaron: this is consistently named options for slash commands with options else it will contain a `_errors` field yeah.
I don't assume it to be named `options` I go through the elements looking for what we know exists which is `_errors`

BartArys (approved):

PR #165, opened 2021-01-11
Proper to string conversion

HopeBaron:


BartArys:
Not that I disapprove of the changes, but why were some `orNull` variants removed but not others? I've got my own conclusions but I'd like to hear yours

HopeBaron:
> Not that I disapprove of the changes, but why were some `orNull` variants removed but not others? I've got my own conclusions but I'd like to hear yours

if we changed to `value.toString()` in which case the returned string will always be present, we can't have both OrNull variants for snowflakes and String

BartArys (approved):

PR #171, opened 2021-01-19
Clean up Application Commands

BartArys:
* Fixes compiler warnings due to experimental usage
* Fixes (de)ser of unknown audit log keys
* calculates some properties instead of assigning them
* Merges Guild and Global application commands and their related builders where possible

### DIFF
--- a/core/src/main/kotlin/behavior/GlobalApplicationCommandBehavior.kt
+++ b/core/src/main/kotlin/behavior/GlobalApplicationCommandBehavior.kt
@@ -4,42 +4,46 @@ import dev.kord.common.annotation.KordPreview
 import dev.kord.common.entity.Snowflake
 import dev.kord.core.cache.data.ApplicationCommandData
 import dev.kord.core.entity.Entity
+import dev.kord.core.entity.interaction.ApplicationCommand
 import dev.kord.core.entity.interaction.GlobalApplicationCommand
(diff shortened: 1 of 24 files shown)


core/src/main/kotlin/behavior/GlobalApplicationCommandBehavior.kt
```diff
@@ -4,42 +4,46 @@ import dev.kord.common.annotation.KordPreview
...
+    val request = ApplicationCommandModifyBuilder().apply(builder).toRequest()
+    val response = when (this) { //sealed classes save us
+        is GlobalApplicationCommandBehavior ->
+            service.modifyGlobalApplicationCommand(applicationId, id, request)
+        is GuildApplicationCommandBehavior ->
+            service.modifyGuildApplicationCommand(applicationId, commandId = id, guildId = guildId, request = request)
+        else -> error("unexpected receiver type $this")
     }
+    val data = ApplicationCommandData.from(response)
+    return GlobalApplicationCommand(data, service)
```
HopeBaron, line 31: This should return a guild behavior if the edited commands belong to a command

BartArys: Good catch, I guess we can't put this into a single function then, since the return type won't match completely.

PR #175, opened 2021-01-24
Allow using the message behaviors in a followup

HopeBaron:
The message is now associated with the follow-up object allowing the use of its behaviors.


core/src/main/kotlin/entity/interaction/InteractionFollowup.kt
```diff
@@ -0,0 +1,42 @@
...
+import dev.kord.common.annotation.KordPreview
+import dev.kord.common.entity.Snowflake
+import dev.kord.core.Kord
+import dev.kord.core.behavior.InteractionFollowupBehavior
+import dev.kord.core.cache.data.MessageData
+import dev.kord.core.entity.*
+import dev.kord.core.entity.channel.MessageChannel
+
+/**
+ * A Message that's created to follow up an [Interaction Response][dev.kord.core.behavior.InteractionResponseBehavior]
```
BartArys, line 12: This doc isn't quite applicable anymore

core/src/main/kotlin/entity/interaction/InteractionFollowup.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val channelId: Snowflake
+        get() = data.channelId
+
+    /**
+     * The message created by this follow-up.
+     *
+     * To use the message behavior your application must be authorized as a bot.
+     */
+    val message: Message
+        get() = Message(data, kord)
```
BartArys, line 40: It's a minor thing, but why aren't we storing the message as is?

BartArys (approved):

BartArys (approved):

PR #176, opened 2021-01-29
Fix incorrect endpoint call

HopeBaron:


BartArys (approved):
We should think about finding a way to automate testing these REST calls sooner than later. Implementation is gonna be hit and miss until we do so.

HopeBaron:
I do believe so too. but what if we hit a rate limit just before acknowledging the interaction then the tests will fail for no reason as of the time being. especially that we already do other rest calls for other endpoints when it comes to testing.

Maybe if kordx.test becomes something sooner than later

PR #178, opened 2021-01-30
Do not throw when getting author member

BartArys:
Current implementation of `Message#getAuthorAsMember` throws when the message was not created in a guild. This change makes it return null instead, as it claims in the docs.


gdude2002 (approved):
Did this one line change need a review process?

HopeBaron (approved):

PR #179, opened 2021-01-30
Fix Groovy DSL example

DRSchlaubi:
The existing example doesn't actually work, because Groovy is dumb

```
Could not find method maven() for arguments [https://oss.sonatype.org/content/repositories/snapshots] on repository container of type org.gradle.api.internal.artifacts.dsl.DefaultRepositoryHandler.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ repositories {
     mavenCentral()
     // Kord Snapshots Repository (Optional):
     maven {
-        url = "https://oss.sonatype.org/content/repositories/snapshots"
+        url "https://oss.sonatype.org/content/repositories/snapshots"
     }
 
 }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +59,9 @@ e.g: `0.7.x-SNAPSHOT`
 repositories {
     mavenCentral()
     // Kord Snapshots Repository (Optional):
-    maven("https://oss.sonatype.org/content/repositories/snapshots")
+    maven {
+        url = "https://oss.sonatype.org/content/repositories/snapshots"
```
BartArys, line 63:
```suggestion
        url "https://oss.sonatype.org/content/repositories/snapshots"
```

BartArys (changes requested):
Nice catch. While we're changing this anyway, could you drop the assignment operator to make this more idiomatic?

BartArys (approved):

PR #180, opened 2021-02-06
Make slash command builders inline

DRSchlaubi:


HopeBaron:
I've opened a branch that addresses some other *supposed to be* inline functions. Please address those before I or Bart make a review also thank you for doing it :clap:

DRSchlaubi:
If that branch already addresses the other issues why don't we merge that branch in this branch or vice versa?

rest/src/main/kotlin/builder/interaction/InteractionsBuilder.kt
```diff
@@ -206,7 +222,6 @@ class InteractionApplicationCommandCallbackDataBuilder {
             _content,
             _embeds.mapList { it.toRequest() },
             _allowedMentions.map { it.build() })
-            _flags
```
BartArys, line 209: This doesn't look right to me, why was this removed?

DRSchlaubi: because it doesn't do anything. It's just a call to the getter of _flags after the call to the constructor of `InteractionApplicationCommandCallbackData`
 
 ```kotlin
return InteractionApplicationCommandCallbackData(
            _tts,
            _content,
            _embeds.mapList { it.toRequest() },
            _allowedMentions.map { it.build() }) // this ) is the closing ) of the constructor call so the code returns here
            _flags
 ```
 
 So either I am dumb and pressed some weird key that moved this line or it was wrong from the beginning
 
Anyways intellij complained about it beeing unreachable code. I thought i somehow  put it there so i removed it

Probabbly it was supposed to be in the constructor in the beginning

BartArys: Yeah, that should have been inside the constructor call.

DRSchlaubi: There we go I now also fixed a bug

HopeBaron (approved):
Looks fine :+1:

PR #183, opened 2021-02-08
[WIP] remove redundant map structures

HopeBaron:
Remove maps structures and replace them with single values for interactions

Since there will be only one group, or subcommand at a time. However, options remain unchanged


BartArys (commented):
Should probably document what these properties actually are and when they are null. 
I also don't like how the user has to check for this:

```kotlin
if(group != null) //group subcommand
else if (subcommand != null) //subcommand
else //command itself invoked
```

But I'm not really sure how to get around that. The group thing sort of breaks any common hierarchy we can introduce.

HopeBaron:
```kt

fun Command.group(block: (Group) -> Unit) {
    group?.let { block(it) }
}
fun Command.subCommand(block: (SubCommand) -> Unit) {
    subCommand?.let { block(it) }
}
```
Something like this?

BartArys:
> Something like this?

Getting rid of the if statement is the least of the issues. The logic behind figuring out what's being invoked is just really bad.
1. The command is always present, so you can't know if it was invoked or not. A command is invoked only if no sub command or group is invoked.
2. subcommand and group can both be null, but only one of them can be present, this exhaustiveness is not apparent.
3. Just getting the invoked command would be pointless as well, the context of what you're dealing with is equally important considering command frameworks use the 'full qualified name' of an invoked command to identify it.

Figuring out what's being invoked looks more like a case of a sealed class to me: 
| invoked command | invoked direct sub command | invoked group sub command |

This gets rid of the awkward nulls, allows for a much simpler and better communicates the intention of the check:

```kotlin 
when(invocation) {
    is invoked command -> //
    is invoked direct sub command -> //
    is invoked group sub command -> //
}
```
(not commenting on the naming scheme of course)

HopeBaron:
in such a case, isn't introducing three interactions for each type considered a valid solution?

BartArys (approved):
This looks good, do try to be consistent with how you spell subcommand/sub-command though.

PR #184, opened 2021-02-15
Allow providing a message request to create a message

HopeBaron:


BartArys (approved):
Good change, we should do this for every endpoint eventually

HopeBaron:
Sure thing, we missed the multi-part requests as far as I saw while going through files

PR #185, opened 2021-02-17
Bump dependencies

DRSchlaubi:
This PR bumps all dependencies to latest.
The two biggest changes are:
- kotlinx.atomicfu -> 0.15.1
that produced this errors:
```
DefaultGateway.kt:367: dev.kord.gateway.DefaultGateway::handleClose: Unsupported operation on atomic variable
Handler.kt:46: dev.kord.gateway.handler.HeartbeatHandler$start$$inlined$on$2::invokeSuspend: Unsupported branching/control within atomic operation
HeartbeatHandler.kt:61: dev.kord.gateway.handler.HeartbeatHandler$start$$inlined$on$4$lambda$1::invokeSuspend: Unsupported branching/control within atomic operation
```
I could fix them by extracting the update call into it's own method

- ktor to 1.5.1:
After upgrading to Ktor 1.5.1 this error occurred when connecting to the gateway
```
io.ktor.client.call.NoTransformationFoundException: No transformation found: class io.ktor.utils.io.ByteBufferChannel (Kotlin reflection is not available) -> class io.ktor.client.features.websocket.DefaultClientWebSocketSession (Kotlin reflection is not available)
with response from wss://gateway.discord.gg/?v=8&encoding=json&compress=zlib-stream:
status: 400 Bad Request
```

After a bit of debugging I found out that it worked with the OkHttp engine but didn't with CIO so i compared the two requests and saw that CIO sets this header `Sec-WebSocket-Extensions` to no value whilst okhttp doesn't set it all
I couldn't figure out a way to remove the header so I set it to this

Not sure if there is a better way to solve this though.


DRSchlaubi:
Replaced by #198, #199 and #186

Issue #121, opened 2020-11-29
Track open Discord PR's to implement


BartArys:
This issue exists to track PRs/issues on the discord-api-docs repository that document API changes/additions/removals. 
Implementing these isn't a priority until they get merged to the main branch.  Feel free to comment on this issue or notify one of the devs on [discord](https://discord.gg/6jcx5ev) If you find any open PRs/issues that are not tracked in this list.

* [ ] Document missing fields in Create and Modify Guild https://github.com/discord/discord-api-docs/pull/1764
* [x] Document message types 16 and 17 https://github.com/discord/discord-api-docs/pull/1779
* [x] Document missing fields in the Activity object https://github.com/discord/discord-api-docs/pull/2219
* [x] New webhook endpoints https://github.com/discord/discord-api-docs/pull/2243
* [x] New field `pending` in guild member object https://github.com/discord/discord-api-docs/pull/2299/
* [x] Document welcome_screen field on guild object https://github.com/discord/discord-api-docs/pull/1636/
* [ ] ~~Document Screen Sharing endpoints: https://github.com/discord/discord-api-docs/pull/2396 ~~ (Currently on hold, see #218 for more info)~~

HopeBaron:
Added: 
New field `pending` in guild member object https://github.com/discord/discord-api-docs/pull/2299/

HopeBaron:
Added: 
Document Screen Sharing endpoints: https://github.com/discord/discord-api-docs/pull/2396

DRSchlaubi:
> Added:
> Document Screen Sharing endpoints: [discord/discord-api-docs#2396](https://github.com/discord/discord-api-docs/pull/2396)

Pls update this to membership screening and not screensharing

DRSchlaubi:
discord/discord-api-docs#1762 should be added to this

DRSchlaubi:
discord/discord-api-docs#2855 new banger feature to implement

Luewd:
https://github.com/discord/discord-api-docs/pull/3524

There's a new `version` property on application command objects that needs to be added to `ApplicationCommand` and `ApplicationCommandData`. It's a snowflake that represents the "id" of the latest update to the slash command registration, so it's very useful if you want to track the latest version of a slash command registered with a guild (or globally).

HopeBaron:
Thanks for the heads up!

Luewd:
New feature User Commands and message Commands! <https://github.com/discord/discord-api-docs/pull/3614>

[...]

PR #189, opened 2021-02-26
Add bulk update commands via PUT

MrPowerGamerBR:
While not documented in the Discord docs, it was added in 26th of January: https://github.com/discord/discord-api-docs/issues/2490#issuecomment-767964749

Very useful to avoid getting ratelimited when trying to add new commands quickly

### DIFF
--- a/rest/src/main/kotlin/service/InteractionService.kt
+++ b/rest/src/main/kotlin/service/InteractionService.kt
@@ -57,8 +57,7 @@ class InteractionService(requestHandler: RequestHandler) : RestService(requestHa
         applicationId: Snowflake,
         guildId: Snowflake,
         request: ApplicationCommandCreateRequest
-    ) =
-        call(Route.GuildApplicationCommandCreate) {
+    ) = call(Route.GuildApplicationCommandCreate) {
             keys[Route.ApplicationId] = applicationId
             keys[Route.GuildId] = guildId
             body(ApplicationCommandCreateRequest.serializer(), request)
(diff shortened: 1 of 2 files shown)


rest/src/main/kotlin/service/InteractionService.kt
```diff
@@ -16,11 +18,19 @@ class InteractionService(requestHandler: RequestHandler) : RestService(requestHa
     suspend fun createGlobalApplicationCommand(
         applicationId: Snowflake,
         request: ApplicationCommandCreateRequest
-    ): DiscordApplicationCommand = call(Route.GlobalApplicationCommandCreate) {
+    ) = call(Route.GlobalApplicationCommandCreate) {
```
BartArys, line 21:
```suggestion
    ): DiscordApplicationCommand = call(Route.GlobalApplicationCommandCreate) {
```

Unnecessary removal of return type, we're gonna have to specify them all explicitly anyway once we switch to library mode.

BartArys (commented):
Not keen on merging new stuff that hasn't even been officially documented yet (see #121) . But commands are a big enough mess as-is that it might not matter anyhow.

rest/src/main/kotlin/service/InteractionService.kt
MrPowerGamerBR: Then shouldn't all types be explictly set in all functions in the `Interactions.kt` class? Most of them do not have any explict types so that's why I removed them from those two functions.

I can readd them if needed.

BartArys: > Then shouldn't all types be explictly set in all functions in the InteractionServices.kt class? 

Eventually, they should be set everywhere, I'm not going to force people to do the extra work though. It's not a big deal either way.

HopeBaron:
the changes have been applied to interaction-changes already

HopeBaron:
Added in #201

PR #192, opened 2021-02-26
Add support for DM slash commands

### LINKED ISSUE #190: "DiscordInteraction" doesn't work in Direct Messages
On a direct message, the `member` and `guild_id` are null.

**Guild:**
```json
{"channel_id":"513405772911345664","data":{"id":"812846584084955136","name":"online"},"guild_id":"268353819409252352","id":"814909719054647316","member":{"deaf":false,"is_pending":false,"joined_at":"2017-01-10T12:22:16.323000+00:00","mute":false,"nick":null,"pending":false,"permissions":"4294967295","premium_since":null,"roles":["297051132793061378","334711955736625185","401353261266763778","650869752524439563","693210200496144424"],"user":{"avatar":"8bd2b747f135c65fd2da873c34ba485c","discriminator":"4185","id":"123170274651668480","public_flags":131712,"username":"MrPowerGamerBR"}},"token":"-","type":2,"version":1}
```

**Direct Message:**
```json
{"channel_id":"637662290170216458","data":{"id":"812857383717699626","name":"online"},"id":"814909846746431539","token":"-","type":2,"user":{"avatar":"8bd2b747f135c65fd2da873c34ba485c","discriminator":"4185","id":"123170274651668480","public_flags":131712,"username":"MrPowerGamerBR"},"version":1}
```

This causes errors when trying to parse a command received via direct message.

DRSchlaubi:
Fixes #190 

- Make guild stuff nullable
- Add user properties

This is a draft as it takes 2 hrs to test this

### DIFF
--- a/common/src/main/kotlin/entity/Interactions.kt
+++ b/common/src/main/kotlin/entity/Interactions.kt
@@ -18,27 +18,27 @@ val kordLogger = KotlinLogging.logger { }
 @Serializable
 @KordPreview
 data class DiscordApplicationCommand(
-        val id: Snowflake,
-        @SerialName("application_id")
-        val applicationId: Snowflake,
-        val name: String,
-        val description: String,
-        @SerialName("guild_id")
-        val guildId: OptionalSnowflake = OptionalSnowflake.Missing,
-        val options: Optional<List<ApplicationCommandOption>> = Optional.Missing(),
+    val id: Snowflake,
+    @SerialName("application_id")
+    val applicationId: Snowflake,
+    val name: String,
+    val description: String,
+    @SerialName("guild_id")
+    val guildId: OptionalSnowflake = OptionalSnowflake.Missing,
+    val options: Optional<List<ApplicationCommandOption>> = Optional.Missing(),
 )
 
 @Serializable
 @KordPreview
 class ApplicationCommandOption(
-        val type: ApplicationCommandOptionType,
-        val name: String,
-        val description: String,
-        val default: OptionalBoolean = OptionalBoolean.Missing,
-        val required: OptionalBoolean = OptionalBoolean.Missing,
-        @OptIn(KordExperimental::class)
-        val choices: Optional<List<Choice<@Serializable(NotSerializable::class) Any?>>> = Optional.Missing(),
-        val options: Optional<List<ApplicationCommandOption>> = Optional.Missing(),
+    val type: ApplicationCommandOptionType,
+    val name: String,
+    val description: String,
+    val default: OptionalBoolean = OptionalBoolean.Missing,
+    val required: OptionalBoolean = OptionalBoolean.Missing,
+    @OptIn(KordExperimental::class)
+    val choices: Optional<List<Choice<@Serializable(NotSerializable::class) Any?>>> = Optional.Missing(),
+    val options: Optional<List<ApplicationCommandOption>> = Optional.Missing(),
 )
 
 /**
@@ -149,9 +149,9 @@ data class DiscordInteraction(
     val type: InteractionType,
     val data: DiscordApplicationCommandInteractionData,
     @SerialName("guild_id")
-        val guildId: Snowflake? = null,
+    val guildId: Snowflake? = null,
     @SerialName("channel_id")
-        val channelId: OptionalSnowflake = OptionalSnowflake.Missing,
+    val channelId: OptionalSnowflake = OptionalSnowflake.Missing,
     val member: Optional<DiscordInteractionGuildMember> = Optional.Missing(),
     val user: Optional<DiscordInteractionUser> = Optional.Missing(),
     val token: String,
@@ -176,7 +176,7 @@ sealed class InteractionType(val type: Int) {
     object ApplicationCommand : InteractionType(2)
     class Unknown(type: Int) : InteractionType(type)
 
-    override fun toString(): String = when(this){
+    override fun toString(): String = when (this) {
         Ping -> "InteractionType.Ping($type)"
         ApplicationCommand -> "InteractionType.ApplicationCommand($type)"
         is Unknown -> "InteractionType.Unknown($type)"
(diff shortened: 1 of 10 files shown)


HopeBaron:
We should make a distinction between DM and Guild interactions instead. As this helps against always-null properties
There should also be an abstract class for those who want to listen to both and cast depending on what they care about the common properties are meaningful enough to be shared through an abstract class

HopeBaron (changes requested):
Looks good, just address the Optional object and I'll merge it to the interactions changes branch to look more into it

common/src/main/kotlin/entity/Interactions.kt
```diff
@@ -149,14 +149,26 @@ data class DiscordInteraction(
...
         val data: DiscordApplicationCommandInteractionData,
         @SerialName("guild_id")
-        val guildId: Snowflake,
+        val guildId: Snowflake? = null,
         @SerialName("channel_id")
-        val channelId: Snowflake,
-        val member: DiscordInteractionGuildMember,
+        val channelId: Snowflake? = null,
+        val member: DiscordInteractionGuildMember? = null,
+        val user: DiscordInteractionUser? = null,
```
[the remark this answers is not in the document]
DRSchlaubi: they're actually optional here to since there is a default but i forgot that you have the cool Optional class

core/src/main/kotlin/cache/data/InteractionData.kt
```diff
@@ -19,11 +19,12 @@ data class InteractionData(
     val id: Snowflake,
     val type: InteractionType,
     val data: ApplicationCommandInteractionData,
-    val guildId: Snowflake,
-    val channelId: Snowflake,
-    val member: MemberData,
```
HopeBaron, line 24: use optionals here

core/src/main/kotlin/entity/interaction/Interaction.kt
```diff
@@ -41,27 +41,75 @@ class Interaction(
...
     val command: Command
         get() = Command(data.data)
 
     val version: Int get() = data.version
 
+    companion object {
+        operator fun invoke(data: InteractionData, applicationId: Snowflake, kord: Kord): Interaction {
+            return when {
+                data.user != null -> DMInteraction(data, applicationId, kord)
+                data.channelId != null && data.guildId != null && data.member != null ->
```
HopeBaron, line 59: I do believe channelIds are available in both cases cause a DM Channel has a id as well

HopeBaron:
```json
{
  "channel_id": "637662290170216458",
  "data": {
    "id": "812857383717699626",
    "name": "online"
  },
  "id": "814909846746431539",
  "token": "-",
  "type": 2,
  "user": {
    "avatar": "8bd2b747f135c65fd2da873c34ba485c",
    "discriminator": "4185",
    "id": "123170274651668480",
    "public_flags": 131712,
    "username": "MrPowerGamerBR"
  },
  "version": 1
}
```
confirmed with this sent by @MrPowerGamerBR

MrPowerGamerBR: Also, `channel_id` is nullable if you are looking at the docs: https://discord.com/developers/docs/interactions/slash-commands#interaction

*But* Mason says that it is a "future proof" thing and that for every use case that exists, you will receive a Channel ID: https://github.com/discord/discord-api-docs/pull/2568/files#r569025697

core/src/main/kotlin/cache/data/InteractionData.kt
DRSchlaubi: Did GitHub mess up here again? Because I don't see any optionals or nullables here

HopeBaron: They did, in your version those became nullables

MrPowerGamerBR: I think @DRSchlaubi said that "I don't see any optionals or nullables here" because (at least here) the code that had those `guildId`/`channelId`/`member` was removed.

HopeBaron: Yes I'm aware

HopeBaron:
channelId should be moved to the abstract class so they can access it for any kind of instance.

DRSchlaubi:
It is part of InteractionBehavior: https://github.com/kordlib/kord/pull/192/files#diff-fe733d6b39bad8153affc53fd16514ca507e76487ee9892503e3b57a0fa1a4c5R26 so you can already access it

HopeBaron:
Agreed, but it can be null safe since as we have seen the property is always there :+1:

DRSchlaubi:
> Agreed, but it can be null safe since as we have seen the property is always there 👍

https://github.com/kordlib/kord/pull/192#discussion_r584281053

HopeBaron:
> > Agreed, but it can be null safe since as we have seen the property is always there +1
> 
> [#192 (comment)](https://github.com/kordlib/kord/pull/192#discussion_r584281053)

I do understand, but until they do this we don't have to worry about it, at the time being it's always present and you would have to assert it's null. 
either way, this is not future proof for us, we will have to changes things if they provided a new interaction type

DRSchlaubi:
normally u would use the DM or guild subinterface anyways which makes it non-nullable

MrPowerGamerBR:
> either way, this is not future proof for us, we will have to changes things if they provided a new interaction type

> normally u would use the DM or guild subinterface anyways which makes it non-nullable

Then if you want to future proof Kord's code, then you could create a interface that has a non-nullable channel ID, and then guild/DM interfaces can implement that non-nullable channel ID interface.

DRSchlaubi:
I thought i did that maybe i lost it somewhere

DRSchlaubi:
Oh wow I did but I forgot to  add it to guild

PR #193, opened 2021-02-26
Fix Ephemeral Bit Flag

MrPowerGamerBR:


MrPowerGamerBR:
(Already fixed in https://github.com/kordlib/kord/commit/368bc92e9396191946219f0038aceff704c80d6f )

PR #194, opened 2021-03-01
Fix bug that will only start one shard

### LINKED ISSUE #195: Kord only starts one shard
Code to reproduce:
```kotlin
    val kord = Kord(args.firstOrNull() ?: error("token required")) {
        sharding { (0..1) }
    }

    kord.on<MessageCreateEvent> {
        if (message.author?.isBot == true) return@on
        if (message.content == "!ping") message.channel.createMessage("pong")
    }

    kord.login { playing("!ping to pong") }
```

See #194 for more

DRSchlaubi:
All shards will get started by this function:
```kotlin    
   /**
     * Calls [Gateway.start] on each Gateway in [gateways], changing the [GatewayConfiguration.shard] for each Gateway.
     */
    suspend fun start(configuration: GatewayConfiguration): Unit =
        gateways.entries.forEach { (shard, gateway) ->
            val config = configuration.copy(shard = configuration.shard.copy(index = shard))
             gateway.start(config)
        }
    
```

The problem is that Gateway.start() is a suspending functions which suspends as long as the gateway is active
As forEach executes all elements one by one on the same thread the first element calls gateway.start and blocks the coroutines
Therefore the other shards never get started.

I fixed this by wrapping start in coroutineScope {} and launch a new coroutine for every gateway.
This should still let the behavior remain, that Kord.login/MasterGateway.start blocks as long as the Gateway is active but each Gateway instance is in it's on child coroutine

Fixes #195

### DIFF
--- a/core/src/main/kotlin/gateway/MasterGateway.kt
+++ b/core/src/main/kotlin/gateway/MasterGateway.kt
@@ -57,7 +57,7 @@ class MasterGateway(
 
     suspend fun sendAll(command: Command) = gateways.values.forEach { it.send(command) }
 
-    suspend fun detachAll() = coroutineScope { gateways.values.forEach { launch { it.detach() } } }
+    suspend fun detachAll() = gateways.values.forEach { it.detach() }
 
     suspend fun stopAll() = coroutineScope {
         gateways.values.forEach { launch { it.stop() } }
--- a/core/src/main/kotlin/gateway/MasterGateway.kt
+++ b/core/src/main/kotlin/gateway/MasterGateway.kt
@@ -59,9 +59,7 @@ class MasterGateway(
 
     suspend fun detachAll() = gateways.values.forEach { it.detach() }
 
-    suspend fun stopAll() = coroutineScope {
-        gateways.values.forEach { launch { it.stop() } }
-    }
+    suspend fun stopAll() = gateways.values.forEach { it.stop() }
 
     override fun toString(): String {
         return "MasterGateway(gateways=$gateways)"
--- a/core/src/main/kotlin/gateway/MasterGateway.kt
+++ b/core/src/main/kotlin/gateway/MasterGateway.kt
@@ -2,17 +2,19 @@ package dev.kord.core.gateway
 
 import dev.kord.gateway.*
 import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.flattenMerge
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
 import kotlin.time.Duration
 import kotlin.time.microseconds
 
 data class ShardEvent(val event: Event, val gateway: Gateway, val shard: Int)
 
 class MasterGateway(
-        val gateways: Map<Int, Gateway>,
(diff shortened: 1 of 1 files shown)


DRSchlaubi:
All shards will get started by this function:
```kotlin    
   /**
     * Calls [Gateway.start] on each Gateway in [gateways], changing the [GatewayConfiguration.shard] for each Gateway.
     */
    suspend fun start(configuration: GatewayConfiguration): Unit =
        gateways.entries.forEach { (shard, gateway) ->
            val config = configuration.copy(shard = configuration.shard.copy(index = shard))
            launch {
                gateway.start(config)
            }
        }
    
```

The problem is that Gateway.start() is a suspending functions which suspends as long as the gateway is active
As forEach executes all elements one by one on the same thread the first element calls gateway.start and blocks the coroutines
Therefore the other shards never get started.

I fixed this by wrapping start in coroutineScope {} and launch a new coroutine for every gateway.
This should still let the behavior remain, that Kord.login/MasterGateway.start blocks as long as the Gateway is active but each Gateway instance is in it's on child coroutine

core/src/main/kotlin/gateway/MasterGateway.kt
```diff
@@ -32,15 +34,19 @@ class MasterGateway(
...
+        .map { (shard, gateway) -> gateway.events.map { ShardEvent(it, gateway, shard) } }
+        .flattenMerge(gateways.size.coerceAtLeast(1))
 
     /**
      * Calls [Gateway.start] on each Gateway in [gateways], changing the [GatewayConfiguration.shard] for each Gateway.
      */
-    suspend fun start(configuration: GatewayConfiguration) = gateways.entries.forEach { (shard, gateway) ->
-        val config = configuration.copy(shard = configuration.shard.copy(index = shard))
-        gateway.start(config)
+    suspend fun start(configuration: GatewayConfiguration): Unit = coroutineScope {
```
HopeBaron, line 43: launch and join jobs so the starting blocks instead, or we will end up exiting main after login happens.

```kt
    suspend fun start(configuration: GatewayConfiguration) = coroutineScope {
        val jobs = gateways.entries.map { (shard, gateway) ->
            val config = configuration.copy(shard = configuration.shard.copy(index = shard))
            launch { gateway.start(config) }
        }
        jobs.onEach { it.join() }
    }
    ```

HopeBaron (changes requested):

core/src/main/kotlin/gateway/MasterGateway.kt
DRSchlaubi: If you take a look at the docs of coroutineScope docs they say
> This function returns as soon as the given block and **all its children coroutines are completed**
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html

as all gateway coroutines run as long the connection is alive coroutineScope should suspend as long as any gateway connection is alive
meaning it doesn't exit main and also did not when trying this locally

HopeBaron: Hmm good point, this would be redundant

DRSchlaubi:
I reverted the changes to shutdown and logout as discussed that it's not necessary: https://discord.com/channels/556525343595298817/631147109311053844/815904781653770272

BartArys (approved):

HopeBaron (approved):

PR #196, opened 2021-03-01
Allow controlling the number of total shards

HopeBaron:
preferably merge after #194

### DIFF
--- a/core/src/main/kotlin/ClientResources.kt
+++ b/core/src/main/kotlin/ClientResources.kt
@@ -11,6 +11,7 @@ class ClientResources(
     val httpClient: HttpClient,
     val defaultStrategy: EntitySupplyStrategy<*>,
     val intents: Intents,
+    val totalShards: Int = shardCount
 ) {
     override fun toString(): String {
         return "ClientResources(shardCount=$shardCount, httpClient=$httpClient, defaultStrategy=$defaultStrategy, intents=$intents)"
--- a/core/src/main/kotlin/ClientResources.kt
+++ b/core/src/main/kotlin/ClientResources.kt
@@ -1,19 +1,18 @@
 package dev.kord.core
 
-import dev.kord.common.entity.Snowflake
+import dev.kord.core.builder.kord.Shards
 import dev.kord.core.supplier.EntitySupplyStrategy
 import dev.kord.gateway.Intents
-import io.ktor.client.HttpClient
+import io.ktor.client.*
 
 class ClientResources(
     val token: String,
-    val shardCount: Int,
+    val shards: Shards,
     val httpClient: HttpClient,
     val defaultStrategy: EntitySupplyStrategy<*>,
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/ClientResources.kt
```diff
@@ -11,6 +11,7 @@ class ClientResources(
     val httpClient: HttpClient,
     val defaultStrategy: EntitySupplyStrategy<*>,
     val intents: Intents,
+    val totalShards: Int = shardCount
```
BartArys, line 14: We might as well store the complete shard info in the resources, right?

core/src/main/kotlin/behavior/GuildBehavior.kt
```diff
@@ -192,7 +192,7 @@ interface GuildBehavior : KordEntity, Strategizable {
      */
     val gateway: Gateway?
         get() {
-            val shard = id.value.shr(22) % kord.resources.shardCount.coerceAtLeast(1)
+            val shard = id.value.shr(22) % kord.resources.shards.indices.count().coerceAtLeast(1)
```
hakiobo, line 195: Shouldn't we be calculating the shard based on the mod by totalShards instead? 
```suggestion
            val shard = id.value.shr(22) % kord.resources.shards.totalShards.coerceAtLeast(1)
```

core/src/main/kotlin/ClientResources.kt
```diff
@@ -1,18 +1,18 @@
...
     val token: String,
-    val shardCount: Int,
+    val shards: Shards,
     val httpClient: HttpClient,
     val defaultStrategy: EntitySupplyStrategy<*>,
     val intents: Intents,
 ) {
     override fun toString(): String {
-        return "ClientResources(shardCount=$shardCount, httpClient=$httpClient, defaultStrategy=$defaultStrategy, intents=$intents)"
+        return "ClientResources(shardCount=$shards, httpClient=$httpClient, defaultStrategy=$defaultStrategy, intents=$intents)"
```
hakiobo, line 16:
```suggestion
        return "ClientResources(shards=$shards, httpClient=$httpClient, defaultStrategy=$defaultStrategy, intents=$intents)"
```

core/src/main/kotlin/behavior/GuildBehavior.kt
HopeBaron: Good catch

core/src/main/kotlin/ClientResources.kt
HopeBaron: There is no need for this, sharding is a data class.

BartArys (approved):

core/src/main/kotlin/ClientResources.kt
HopeBaron: mb, I mis-understood the point of the change

PR #201, opened 2021-03-04
Interaction changes

HopeBaron:
*  Introduce Ephemeral for hidden interactions
* Rename and remove interaction types based on https://github.com/discord/discord-api-docs/pull/2615
* Support resolvable  entities https://gist.github.com/msciotti/b1dabde51ebe2f3e875faf73a7a3509b
* Adds a core OptionValue type
depends on #183


common/src/main/kotlin/entity/optional/Optional.kt
```diff
@@ -251,6 +251,16 @@ inline fun <E: Any, T> Optional<E?>.mapNotNull(mapper: (E) -> T): Optional<T?> =
     is Value -> Optional(mapper(value!!))
 }
 
+inline fun <E> Optional<List<E>>.firstOrNull(mapper: (E) -> Boolean) : E? = when(this){
```
BartArys, line 261: Argument should be called `predicate` to be more idiomatic.

common/src/main/kotlin/entity/optional/Optional.kt
```diff
@@ -251,6 +251,16 @@ inline fun <E: Any, T> Optional<E?>.mapNotNull(mapper: (E) -> T): Optional<T?> =
...
+    is Missing, is Null<*> -> null
+    is Value -> value.firstOrNull(mapper)
+}
+
+
+inline fun <E> Optional<List<E>>.first(mapper: (E) -> Boolean = {true}) : E = firstOrNull(mapper)!!
```
BartArys: Argument should be called `predicate` to be more idiomatic.

core/src/main/kotlin/entity/interaction/Interaction.kt
```diff
@@ -80,11 +80,28 @@ class Interaction(
...
     abstract val rootId: Snowflake
+
+    /**
+     * The root command name
+     */
     abstract val rootName: String
+
+    /**
+     * Options passed to the command.
+     * Type-check your command to against the sub-types of this class for the right context of the execution
```
BartArys, line 98: 'to against'?

core/src/main/kotlin/entity/interaction/Interaction.kt
```diff
@@ -80,11 +80,28 @@ class Interaction(
...
+     * * [RootCommand] - Context of main command execution.
+     * * [GroupCommand] - Context of a subcommand executed in a group.
+     * * [SubCommand] - Context of a subcommand executed under the root command.
```
BartArys: Should this be documented here? I'm not sure how this is relevant to the options?

core/src/main/kotlin/entity/interaction/Interaction.kt
```diff
@@ -109,13 +109,11 @@ sealed class InteractionCommand {
...
             val rootPredicate = firstLevelOptions.isEmpty() || firstLevelOptions.any { it.value.value != null }
             val groupPredicate = firstLevelOptions.any { it.subCommands.orEmpty().isNotEmpty() }
-            val subCommandPredicate = firstLevelOptions.any { it.values.orEmpty().isNotEmpty() }
 
             return when {
                 rootPredicate -> RootCommand(data)
                 groupPredicate -> GroupCommand(data)
-                subCommandPredicate -> SubCommand(data)
-                else -> error("Undefined data structure")
+                else -> SubCommand(data) // if not root, or group, it's a sub-command
```
BartArys: While throwing an error isn't great, assuming that the structure is a sub-command because it didn't match the previous predicates is going to come back to haunt us when Discord decides to add another structure type.

common/src/main/kotlin/entity/DiscordMessage.kt
```diff
@@ -264,7 +264,7 @@ data class DiscordMentionedChannel(
         val name: String,
 )
 
-enum class MessageFlag(val code: Long) {
+enum class MessageFlag(val code: Int) {
```
BartArys: Why the change? Why longs in the first place?

rest/src/main/kotlin/builder/interaction/InteractionsBuilder.kt
```diff
@@ -265,11 +286,13 @@ class FollowupMessageCreateBuilder : RequestBuilder<MultipartFollowupMessageCrea
 
     override fun toRequest(): MultipartFollowupMessageCreateRequest =
         MultipartFollowupMessageCreateRequest(
-            FollowupMessageCreateRequest(_content,
+            FollowupMessageCreateRequest(
+                _content,
                 _username,
                 _avatarUrl,
```
BartArys, line 289: Should probably use named arguments here for safety

common/src/main/kotlin/entity/Member.kt
```diff
@@ -18,8 +18,8 @@ data class DiscordGuildMember(
         val joinedAt: String,
         @SerialName("premium_since")
         val premiumSince: Optional<String?> = Optional.Missing(),
-        val deaf: Boolean,
-        val mute: Boolean,
+        val deaf: OptionalBoolean = OptionalBoolean.Missing,
+        val mute: OptionalBoolean = OptionalBoolean.Missing,
```
BartArys, line 22: You didn't change this in the data objects, does this compile?

HopeBaron: We don't expose the field anywhere in the Member object as of the time being. That's something we have to address

core/src/main/kotlin/entity/interaction/Interaction.kt
HopeBaron: I'm aware, but I assumed if they add another structure type it won't have the same structure as the Sub-command it will fail.
The other way to do it was to do exhaustive checks on the data structures which are redundant if the structure will fail at the serialization level

BartArys: > I'm aware, but I assumed if they add another structure type it won't have the same structure as the Sub-command it will fail.

I'd rather it fail-fast with a descriptive error though.

> The other way to do it was to do exhaustive checks on the data structures which are redundant if the structure will fail at the serialization level

we don't make the distinction at the ser level right now though

common/src/main/kotlin/entity/DiscordMessage.kt
HopeBaron: This was due to a typo in  https://github.com/discord/discord-api-docs/pull/2615
which had Empheral flag set as (1 << 64) but I was actually (1 << 6). (The post has been edited since we pointed this out)
The commit was reverted later on in the branch back to int

BartArys: This sounds like something that might grow in the future, like the permission flags

core/src/main/kotlin/entity/interaction/Interaction.kt
```diff
@@ -252,10 +263,26 @@ sealed class OptionValue<T>(val value: T) {
...
+                        resolvedObjects.channels?.get(snowflake) != null -> ChannelOptionValue(
+                            resolvedObjects.channels?.get(
+                                snowflake
+                            )!!
+                        )
+                        resolvedObjects.roles?.get(snowflake) != null -> RoleOptionValue(
+                            resolvedObjects.roles?.get(
+                                snowflake
+                            )!!
+                        )
```
BartArys: What's up with this janky formatting

core/src/test/kotlin/interaction/CommandsTypeTests.kt
```diff
@@ -1,86 +0,0 @@
...
-        val data = ApplicationCommandInteractionData.from(grouping)
-        val command = InteractionCommand(data)
-        assert(command is GroupCommand)
-        command as GroupCommand
-        assertEquals(1, command.options["argument"]?.int())
-        assertEquals("root", command.rootName)
-        assertEquals("group", command.groupName)
-        assertEquals("subCommand", command.name)
-    }
-}
```
BartArys: This got deleted because?

BartArys (approved):

PR #203, opened 2021-03-05
Provide the same functions of Permissions to MessageFlags

HopeBaron:


BartArys (commented):
I missed this during the initial review, but why is `MessageFlags` a data class? We hide the constructor behind an `internal` keyword but we still expose the autogenerated `copy` method.

Additionally, the autogenerated tostring isn't really useful here either. Returning a tostring with the `flags` would be of more use to those trying to debug.

HopeBaron:
> I missed this during the initial review, but why is `MessageFlags` a data class? We hide the constructor behind an `internal` keyword but we still expose the autogenerated `copy` method.
> 
> Additionally, the autogenerated tostring isn't really useful here either. Returning a tostring with the `flags` would be of more use to those trying to debug.

Good catch, will work on it

BartArys (approved):

PR #208, opened 2021-03-15
Add USE_SLASH_COMMANDS permission

DRSchlaubi:
Permission does already exist but is not documented. See: discord/discord-api-docs#2632


DRSchlaubi:
wow that was fast

HopeBaron:
it's a single line

PR #217, opened 2021-03-25
Add New webhook endpoints

DRSchlaubi:
Relates to #121
Adds:  discord/discord-api-docs#2243


core/src/main/kotlin/behavior/MessageBehavior.kt
```diff
@@ -212,13 +235,51 @@ interface MessageBehavior : KordEntity, Strategizable {
...
+@OptIn(ExperimentalContracts::class)
+suspend inline fun MessageBehavior.editWebhookMessage(
+    token: String,
+    builder: EditWebhookMessageBuilder.() -> Unit
+): Message {
+    contract {
+        callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+    }
+
+    require(webhookId != null) { "Cannot perform edits on  non webhook messages use edit() instead" }
```
HopeBaron, line 274: Why is this nullable in the first place?

DRSchlaubi: because both bot messages and webhook messasges are the same object and only one of them has a webhook ID

rest/src/main/kotlin/builder/webhook/EditWebhookMessageBuilder.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    private var _allowedMentions: Optional<AllowedMentions> = Optional.Missing()
+    var allowedMentions: AllowedMentions? by ::_allowedMentions.delegate()
+
+    @OptIn(ExperimentalContracts::class)
+    inline fun embed(builder: EmbedBuilder.() -> Unit) {
+        contract {
+            callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+        }
+        embeds.add(EmbedBuilder().apply(builder).toRequest())
```
HopeBaron, line 29: Make the embeds a list of EmbedBuilder, and transform it to a request down in the toRequest. this will allow the user to modify the list after done adding or perform whatever they please

HopeBaron (changes requested):

core/src/main/kotlin/behavior/MessageBehavior.kt
HopeBaron: this shouldn't be here then, this should be in the webhook service(???)

HopeBaron: nvm, I misunderstood, but you do expect a webhook id in the behavior it can't be nullable since it's a requirement to call the endpoint

DRSchlaubi: > nvm, I misunderstood, but you do expect a webhook id in the behavior it can't be nullable since it's a requirement to call the endpoint

The problem is we have no WebhookMessage or similar we only have Message and MessageBehvaior

Currently, you can call MessageBehavior.edit{} and it will give you a HTTP error, because it calls the normal message edit endpoint

HopeBaron (approved):
Whilst I don't like the idea that we are mixing a Message that's supposed to be sent by a webhook with a normal message. This might lead to an error because someone decided to call editWebhook on a normal message and should be looked into. However, this is not something that relates to this PR.

BartArys (changes requested):
Implementing these endpoints seem to be a rather big hassle for our current structure. Message behaviours cannot reliably know when they are a webhook message or not (because either Discord won't tell us, or we've made assumptions about knowing enough that have now come back to bite us in the butt).

Because of this lack of information, I think the checks you introduced are brittle and will cause more problems than the might solve. 

Looking at this from a user's perspective, no one should be trying to edit a random message and the distance between creating a message and editing it is relatively small in code. I don't think confusing whether a message is a webhook or not would a frequent occurrence. (certainly not without also having a full message object in hand).

Ideally, these checks would be covered by the type system instead. But even if we were to discern a user message from a webhook message with type system, we can't do that for the behaviors.

core/src/main/kotlin/behavior/MessageBehavior.kt
```diff
@@ -178,20 +199,22 @@ interface MessageBehavior : KordEntity, Strategizable {
...
+    ): MessageBehavior = MessageBehavior(channelId, id, kord, webhookId, strategy)
 
 }
 
- fun MessageBehavior(
+fun MessageBehavior(
     channelId: Snowflake,
     messageId: Snowflake,
     kord: Kord,
+    webhookId: Snowflake? = null,
```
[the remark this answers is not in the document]
DRSchlaubi: What else should I do then?
Editing webhook messages require a different endpoint, which also requires the webhook id, hence it's part of the endpoint path. If the message is a webhook message but does not have a webhook id because discord didn't give it to us the edit would fail either way. Ideally we would have a webhookmessage entity but that raises the issue again that we don't always know whether something is a webhook message or not

HopeBaron:
> Implementing these endpoints seem to be a rather big hassle for our current structure. Message behaviours cannot reliably know when they are a webhook message or not (because either Discord won't tell us, or we've made assumptions about knowing enough that have now come back to bite us in the butt).
> 
> Because of this lack of information, I think the checks you introduced are brittle and will cause more problems than the might solve.
> 
> Looking at this from a user's perspective, no one should be trying to edit a random message and the distance between creating a message and editing it is relatively small in code. I don't think confusing whether a message is a webhook or not would a frequent occurrence. (certainly not without also having a full message object in hand).
> 
> Ideally, these checks would be covered by the type system instead. But even if we were to discern a user message from a webhook message with type system, we can't do that for the behaviors.

I've gone through multiple libraries and checked how they identify a webhook, it seems to be through the webook_id field.
But the fact that both representations and behaviors are mixed up in those implementations is something that I don't like

core/src/main/kotlin/behavior/MessageBehavior.kt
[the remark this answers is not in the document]
BartArys: > What else should I do then?

Make it a mandatory argument when editing a message from a behavior, drop the field from the behavior interface.

Discord has different behavior based on message types, so it might be time to implement Discord's message types as concrete types ourselves. If we do, we can create an overload for webhook messages that doesn't require the id argument. That's something for another PR though.

DRSchlaubi:
> seems to be through the webook_id field.
> But the fact that both representations and behaviors are mixed up in those implementations is something that I don't like

I didn't like that too, when I implemented it that way but I didn't want to change everything related to webhooks in this PR. It is something we need to change though

core/src/main/kotlin/behavior/MessageBehavior.kt
```diff
@@ -212,13 +231,48 @@ interface MessageBehavior : KordEntity, Strategizable {
  * @return The edited [Message].
  *
  * @throws [RestRequestException] if something went wrong during the request.
+ * @throws [IllegalStateException] if this message is a webhook message (See [editWebhookMessage]).
+ * @see editWebhookMessage
```
BartArys, line 235: This is no longer appropriate, is it?

core/src/main/kotlin/behavior/MessageBehavior.kt
```diff
@@ -212,13 +231,48 @@ interface MessageBehavior : KordEntity, Strategizable {
...
+    return Message(data, kord)
+}
+
+/**
+ * Requests to edit this message.
+ *
+ * @return The edited [Message].
+ *
+ * @throws [RestRequestException] if something went wrong during the request.
+ * @throws [IllegalStateException] if this message is a normal message (see [edit]).
```
BartArys, line 256: This isn't appropriate anymore either.

BartArys (changes requested):

BartArys (approved):

BartArys (commented):
In the `RestServiceTest` in core there's a test dedicated to some webhook behaviour. See if you can fit the new endpoints in there.

BartArys (approved):

PR #216, opened 2021-03-25
Member Screening core support

HopeBaron:

PR #218, opened 2021-03-25
Add membership screening

DRSchlaubi:
- Adds discord/discord-api-docs#2396
- Add models
- Add rest requests

Related to #121, #216

### DIFF
--- a/common/src/main/kotlin/entity/DiscordGuild.kt
+++ b/common/src/main/kotlin/entity/DiscordGuild.kt
@@ -151,7 +151,7 @@ data class DiscordGuild(
     @SerialName("approximate_presence_count")
     val approximatePresenceCount: OptionalInt = OptionalInt.Missing,
 
-)
+    )
 
 /**
  * A partial representation of a [Discord Guild structure](https://discord.com/developers/docs/resources/guild#guild-object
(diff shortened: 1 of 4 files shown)


common/src/main/kotlin/entity/DiscordGuild.kt
```diff
@@ -557,4 +574,28 @@ data class DiscordWelcomeScreen(
...
+    @Serializable
+    data class FormField(
+        @SerialName("field_type")
+        val type: Type,
+        val label: String,
+        val values: Optional<List<String>> = Optional.Missing(),
+        val required: Boolean
+    ) {
+        @Serializable
+        enum class Type {
```
HopeBaron, line 596: Use a more expresive name than `Type`, maybe `FormFieldType` is a better. name
Also use objects so we can have `Unknown` type in there

HopeBaron (changes requested):
There is no core representation for member screening

DRSchlaubi:
actually, they removed the docs for membership screening so I stopped implementing them until they do a PR adding them again

HopeBaron:
Any updates on this?

DRSchlaubi:
https://discord.com/developers/docs/resources/guild#membership-screening-object still has the warning and I don't see any PR changeing it so I guess no updates

HopeBaron:
Understood :+1:

PR #186, opened 2021-02-17
Add .editorconfig file

DRSchlaubi:
Currently Kord uses the deprecated JetBrains code style guidelines. This causes quite a lot of reformatting noise in PR (totally not trying to distract from my pr noise)
In order to solve this problem I added an [.editorconfig](https://editorconfig.org/) file, which is just a generated copy of the intellij codeStyle.xml for the [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html)
This PR also reformats the whole project according to the new style guide

PR #224, opened 2021-03-27
Implement slash command permissions

DRSchlaubi:
This implements discord/discord-api-docs#2737
I will update this PR in case the Discord PR get's updated


DRSchlaubi:
Welp i totally missed that they merged it

BartArys:
Your commit contains over 300 purely cosmetic changes (some in otherwise untouched files). Could you please on PR changes that are relevant to the PR? I don't have the mental capacity to go over all these changes in detail and I might end up missing stuff like this: 
image

DRSchlaubi:
> Your commit contains over 300 purely cosmetic changes (some in otherwise untouched files). Could you please on PR changes that are relevant to the PR? I don't have the mental capacity to go over all these changes in detail and I might end up missing stuff like this:
> image

Thats why we need #186 because If I format my file it formats everything else as well. Also I might screwed up something whilst switching branches and inherited some changes that i didn't want to put in here

HopeBaron:
You should be using Kord's standard formatting until we merge the styling branch. which would preferably be after everything important has been merged

common/src/main/kotlin/entity/Interactions.kt
```diff
@@ -391,6 +410,51 @@ sealed class InteractionResponseType(val type: Int) {
...
+data class DiscordApplicationCommandPermission(
+    val id: Snowflake,
+    val type: Type,
+    val permission: Boolean
+) {
+    @Serializable(with = Type.Serializer::class)
+    enum class Type(val value: Int) {
+        ROLE(1),
+        USER(2);
+
```
HopeBaron, line 445: I'd say a class with `object`s would be more useful than an enum in case Discord, somewhat introduced a new type of permissions

HopeBaron (changes requested):

DRSchlaubi:
> You should be using Kord's standard formatting until we merge the styling branch. which would preferably be after everything important has been merged

That's why I made that PR what even is the Style Kord uses

DRSchlaubi:
Superseeded by #253

PR #225, opened 2021-03-29
Add ResolvedChannel

HopeBaron:
adds support to resolved channels by application commands

### DIFF
--- a/core/src/main/kotlin/entity/channel/ResolvedChannel.kt
+++ b/core/src/main/kotlin/entity/channel/ResolvedChannel.kt
@@ -1,0 +1,21 @@
+package dev.kord.core.entity.channel
+
+import dev.kord.common.annotation.KordPreview
+import dev.kord.common.entity.Permissions
+import dev.kord.core.Kord
+import dev.kord.core.cache.data.ChannelData
+import dev.kord.core.supplier.EntitySupplier
+import dev.kord.core.supplier.EntitySupplyStrategy
+
+@KordPreview
+class ResolvedChannel(override val data: ChannelData, override val kord: Kord, val strategy: EntitySupplyStrategy<*>): Channel {
+
+
+    val name: String get() = data.name.value!!
+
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/entity/channel/ResolvedChannel.kt
```diff
@@ -0,0 +1,25 @@
...
+) : Channel {
+
+
+    val name: String get() = data.name.value!!
+
+    val permissions: Permissions get() = data.permissions.value!!
+
+    override val supplier: EntitySupplier
+        get() = strategy.supply(kord)
+
```
BartArys, line 24: Should add a meaningful toString

BartArys (approved):

core/src/main/kotlin/entity/channel/ResolvedChannel.kt
HopeBaron: Should we add the object reference for the type, or it's value (Discord value) Although the object gives better idea of the actual type. However the value is less complicated than this: 
```kt
Type: dev.kord.common.entity.ChannelType$GuildCategory@47e3a62b
```

BartArys: We should probably just give those types a better toString as well, though that's not the scope of this PR

core/src/main/kotlin/entity/channel/ResolvedChannel.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun toString(): String {
+        return "ResolvedChannel(id=$id,type=$type, name=$name, permissions=$permissions)"
```
BartArys, line 26: minor nitpick

```suggestion
        return "ResolvedChannel(id=$id, type=$type, name=$name, permissions=$permissions)"
```

BartArys (approved):

PR #228, opened 2021-03-29
Add support for editing the original interaction with files

MrPowerGamerBR:
Related to #226

I made it similar to how Kord handles "normal" message file sending, currently it works fine.

Keep in mind that this only changes stuff in the REST module, I haven't changed anything in the core module. (maybe someone could do this later?)

image

```kotlin
            val kordMessage = rest.interaction.modifyInteractionResponse(
                applicationId,
                request.token
            ) {
                this.content = message.content
                this.addFile("ayaya.png", File("L:\\Pictures\\ExNJki9U8AAzsQe.jpg").inputStream())
                this.addFile("owo.png", File("L:\\Pictures\\1580734785123.png").inputStream())
            }
```

(For now this only adds editing original interactions with files, maybe I will do the webhook edit also in this pull request... or maybe in a different pull request!)

PR #230, opened 2021-03-29
Refactor `files` and `addFile` methods in the REST requests to avoid code duplication

MrPowerGamerBR:
This is blocked by https://github.com/kordlib/kord/pull/228 and https://github.com/kordlib/kord/pull/229

A lot of request builders in Kord have the same `files` list and the same implementation for the `addFile` methods, to avoid code duplication, this PR adds a `MultipleAttachmentsRequest` (I think this is a bad name so if anyone has a better name, please tell me!) which adds default implementations to the `addFile` methods, which should help avoid code duplication.

This will also help if more file related methods are added (example: Loading a file from a `URL`), this way we only need to change the interface instead of adding the new method for every request that has a `files` list. 😊


HopeBaron:
I doubt it's worth adding, but a better name would be `MultipartBuilder` or the like it's not a request.

MrPowerGamerBR:
> I doubt it's worth adding

Keep in mind that there will be other requests that aren't implemented in Kord yet that can also use the `MultipartBuilder` (Example: Webhook File Sending, Webhook Message File Editing, etc), that's why I thought that it would be nice to keep all the file related stuff in a single interface.

And yes, `MultipartBuilder` is a better name :)

HopeBaron:
I'm waiting for the name change to merge this. I've seen the issue related to adding multiple file support.

BartArys:
I'd be careful with the `files` property. In it's not unthinkable to run into a builder that can only accept a maximum of one file, in which case the aim of the interface (to unite all file-addable builders) would fail. And would `addFile` be the correct term for a builder than can only accept one file max?

In essence what you're PR'ing is a `FileAddable` interface, calling it a `MultipartBuilder` or a `MultipleAttachmentsRequest` would be inaccurate, as it could be applied to situations that are neither multiplart form requests (rather unlikely) or related to attachments (consider programmatically changing the icon of a server).

> this way we only need to change the interface instead of adding the new method for every request that has a files list.

If we do introduce this interface, we shouldn't strive to mindlessly break it by adding new methods to it.

---

I don't see people writing much code against such a "file-addable" interface, nor implementing it themselves, the only proper benefit would be extension functions that carry over between builders.

Issue #227, opened 2021-03-29
Add `application_id` to the Interaction object


MrPowerGamerBR:
They are present in the Ping and in the Interaction object

```json
{"application_id":"744361365724069898","id":"826113573226676284","token":"owo","type":1,"user":{"avatar":"8bd2b747f135c65fd2da873c34ba485c","discriminator":"4185","id":"123170274651668480","public_flags":131712,"username":"MrPowerGamerBR"},"version":1}
```

```json
{"application_id":"744361365724069898","channel_id":"358774895850815488","data":{"id":"826111687744421950","name":"imageupload"},"guild_id":"297732013006389252","id":"826113573067554836","member":{"deaf":false,"is_pending":false,"joined_at":"2017-04-01T14:00:43.649000+00:00","mute":false,"nick":"MrPowerGamerBR😘|💁Criador da Lori","pending":false,"permissions":"8589934591","premium_since":"2019-06-04T18:02:56.819000+00:00","roles":["302218401726726152","334734175531696128","351473717194522647","364201981016801281","385579854336360449","399301696892829706","434512654292221952","463652112656629760","493150630173605928","505144985591480333","510788363264196638","513303483659714586","513303543911022593","513310965647933443","513310993326014464","526720753991811072","532659914615554078","534659343656681474","585528935475970064","589159900379086868","592833969133715503","619691791041429574","655132411566358548","655133385899114497","655133474528690177","655133580145459201","655133835037769779","655135974572949514","655136230086017034","655136407404412979","655137001217064960","655137090169733140","668976976387899417","700368699701592185","807236985420906516"],"user":{"avatar":"8bd2b747f135c65fd2da873c34ba485c","discriminator":"4185","id":"123170274651668480","public_flags":131712,"username":"MrPowerGamerBR"}},"token":"uwu","type":2,"version":1}
```

PR #231, opened 2021-03-29
Add Application ID field to the command interaction object

MrPowerGamerBR:
Fixes https://github.com/kordlib/kord/issues/227


HopeBaron (approved):

HopeBaron:
There should be an applicationId in client resources for those who would use the older application before 2012(?) where the application ID differs from the user ID. This won't do much of an effect now. But is still a useful addition. tell me if you are willing to add it in this PR or not so I can merge if no further commits will be made

MrPowerGamerBR:
> There should be an applicationId in client resources for those who would use the older application before 2012(?) where the application ID differs from the user ID. This won't do much of an effect now. But is still a useful addition. tell me if you are willing to add it in this PR or not so I can merge if no further commits will be made
I think it should be added in a different PR just to avoid mixing two similar yet different things

Also if I recall correctly that was in 2016 (if I recall correctly), when Discord added "bot accounts", people could convert user accounts into bot accounts and that's why the application ID differs between them.

PR #232, opened 2021-03-30
Add webhook getters for guilds

HopeBaron:


BartArys (approved):
The only issue I have is that you're calling these getters

PR #234, opened 2021-04-01
Add Ping request and response for interactions

HopeBaron:
This PR aims to add representation of both the ping interaction and the pong response type to the rest module.

Currently, there are no plans to add core support since we believe this is out of Kord's scope.


HopeBaron:
This is no longer a valid PR

PR #239, opened 2021-04-04
Implement voice stage channel

HopeBaron:


HopeBaron:
I'm unsure about:
* VoiceStageChannelChannel is a VoiceBehavior?
* Should we expose the voice state patching methods to GuildBehavior?

BartArys:
> I'm unsure about:
> 
>     * VoiceStageChannelChannel is a VoiceBehavior?

I assume so, the behavior only exposes `voiceStates`, which should be present, might have to test that.

>     * Should we expose the voice state patching methods to GuildBehavior?

I assume you're talking about `Update Self Voice State` and `Update Others Voice State`. Considering the caveats mentioned in the docs, I think it'd be most safe to limit it to the stage channel's behavior only.

core/src/main/kotlin/cache/data/VoiceStateData.kt
```diff
@@ -31,7 +29,8 @@ data class VoiceStateData(
     val selfMute: Boolean,
     val selfStream: OptionalBoolean = OptionalBoolean.Missing,
     val suppress: Boolean,
-) {
+    val requestToSpeakTimestamp: String?
+    ) {
```
BartArys, line 33:
```suggestion
) {
```

core/src/main/kotlin/behavior/channel/StageChannelBehavior.kt
```diff
@@ -0,0 +1,76 @@
...
+        id: Snowflake,
+        guildId: Snowflake,
+        kord: Kord,
+        supplier: EntitySupplier = kord.defaultSupplier
+    ): StageChannelBehavior = object : StageChannelBehavior {
+        override val guildId: Snowflake get() = guildId
+        override val id: Snowflake get() = id
+        override val kord get() = kord
+        override val supplier get() = supplier
+    }
```
BartArys, line 38: This shouldn't be here

core/src/main/kotlin/entity/channel/StageVoiceChannel.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun hashCode(): Int = Objects.hash(id, guildId)
+
+    override fun equals(other: Any?): Boolean = when (other) {
+        is GuildChannelBehavior -> other.id == id && other.guildId == guildId
+        is ChannelBehavior -> other.id == id
+        else -> false
+    }
+
+    override fun toString(): String {
+        return "VoiceChannel(data=$data, kord=$kord, supplier=$supplier)"
```
BartArys, line 48:
```suggestion
        return "StageChannel(data=$data, kord=$kord, supplier=$supplier)"
```

rest/src/main/kotlin/builder/guild/VoiceStateModifyBuilder.kt
```diff
@@ -0,0 +1,37 @@
...
+import dev.kord.rest.json.request.CurrentVoiceStateModifyRequest
+import dev.kord.rest.json.request.VoiceStateModifyRequest
+
+class CurrentVoiceStateModifyBuilder(val channelId: Snowflake) : RequestBuilder<CurrentVoiceStateModifyRequest> {
+
+    private var _requestToSpeakTimestamp: Optional<String> = Optional.Missing()
+
+    private var _suppress: OptionalBoolean = OptionalBoolean.Missing
+
+    var requestToSpeakTimestamp: String? by ::_requestToSpeakTimestamp.delegate()
```
BartArys, line 17: This should probably be a date-related object like LocalDateTime.

Also, does the timestamp matter at all? The docs say:
[You are able to set request_to_speak_timestamp to any present or future time.](https://github.com/discord/discord-api-docs/blob/5dcf679ad25ec449bcb1d77e1b113d6f456c11ea/docs/resources/Guild.md#update-current-user-voice-state--patch-guildsguildiddocs_resources_guildguild-objectvoice-statesme)

Can you request to speak at a specific time? Or does any request in the future behave the same? If it's the latter, we could just abstract this.

core/src/main/kotlin/behavior/channel/StageChannelBehavior.kt
```diff
@@ -0,0 +1,76 @@
...
+    guildId: Snowflake,
+    kord: Kord,
+    supplier: EntitySupplier = kord.defaultSupplier
+): StageChannelBehavior = object : StageChannelBehavior {
+    override val guildId: Snowflake
+        get() = guildId
+    override val kord get() = kord
+    override val id: Snowflake get() = id
+    override val supplier get() = supplier
+}
```
BartArys, line 76:
```suggestion
}

```

core/src/main/kotlin/entity/channel/StageVoiceChannel.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
```
BartArys, line 50:
```suggestion
}

```

rest/src/main/kotlin/builder/guild/VoiceStateModifyBuilder.kt
```diff
@@ -0,0 +1,37 @@
...
+class VoiceStateModifyBuilder(val channelId: Snowflake) : RequestBuilder<VoiceStateModifyRequest> {
+
+    private var _suppress: OptionalBoolean = OptionalBoolean.Missing
+
+    var suppress: Boolean? by ::_suppress.delegate()
+
+    override fun toRequest(): VoiceStateModifyRequest {
+        return VoiceStateModifyRequest(channelId, _suppress)
+    }
+}
```
BartArys, line 37:
```suggestion
}

```

rest/src/main/kotlin/json/request/VoiceStateRequests.kt
```diff
@@ -0,0 +1,24 @@
...
+    val requestToSpeakTimeStamp: Optional<String> = Optional.Missing()
+)
+
+
+@Serializable
+data class VoiceStateModifyRequest(
+    @SerialName("channel_id")
+    val channelId: Snowflake,
+    val suppress: OptionalBoolean = OptionalBoolean.Missing
+)
```
BartArys, line 24:
```suggestion
)

```

BartArys (commented):
Need some docs on this, especially the core stuff and the builder variables

core/src/main/kotlin/behavior/channel/StageChannelBehavior.kt
HopeBaron: Where should that be?

BartArys: outside of the interface ideally, but you already did so: https://github.com/kordlib/kord/pull/239/files/d346d5fd1c918c7d15f95e5d1f8b50df07cf85c7#diff-d6002ad60499d33441ee16a017fd1fcb2ce12050a4c65e56930839376f3c62d7R65-R76

BartArys, line 38: This is still here, while it should be defined outside the interface.
```suggestion
```

core/src/main/kotlin/behavior/channel/StageChannelBehavior.kt
BartArys, line 76: Ideally, this should have a meaningful toString

core/src/main/kotlin/behavior/channel/StageChannelBehavior.kt
HopeBaron: I'm sorry didn't notice that I've copied the declaration even though you have mentioned it earlier

BartArys (approved):

PR #246, opened 2021-04-05
delete/edit acknowledgements

HopeBaron:


core/src/main/kotlin/behavior/GuildInteractionBehavior.kt
```diff
@@ -11,8 +11,4 @@ interface GuildInteractionBehavior : InteractionBehavior {
 
     suspend fun getGuild(): Guild = supplier.getGuild(guildId)
 
-    companion object {
-
-    }
-
```
BartArys, line 17: Why'd we end up deleting this? Companion objects are useful for third-party extension functions: see [Companion object for Boolean](https://youtrack.jetbrains.com/issue/KT-7922).

BartArys (approved):

PR #255, opened 2021-04-18
Add a values field to Permssions

HopeBaron:


common/src/main/kotlin/entity/Permission.kt
```diff
@@ -149,5 +149,5 @@ sealed class Permission(val code: DiscordBitSet) {
     object ManageWebhooks : Permission(0x20000000)
     object ManageEmojis : Permission(0x40000000)
     object UseSlashCommands : Permission(0x80000000)
-    object All : Permission(0x7FFFFDFF)
+    object All : Permission(0xFFFFFDFF)
```
BartArys, line 156: Should consider making this a computed in the future, good catch.

common/src/main/kotlin/entity/Permission.kt
```diff
@@ -150,4 +154,42 @@ sealed class Permission(val code: DiscordBitSet) {
...
+                MoveMembers,
+                UseVAD,
+                PrioritySpeaker,
+                ChangeNickname,
+                ManageNicknames,
+                ManageRoles,
+                ManageWebhooks,
+                ManageEmojis,
+                UseSlashCommands,
+                All,
```
BartArys, line 192: Should all be part of the set? It's technically not a Discord permission and we don't include it in intents.

HopeBaron: Good catch, I will edit it

BartArys (approved):

PR #267, opened 2021-04-21
Update maven central badge

DRSchlaubi:
the nexus one doesn't work LOL


HopeBaron:
Should also modify L53 and maybe remove the snapshot badge

Issue #270, opened 2021-05-01
Assing a coroutine scope to Live Entities


BartArys:
CoroutineScopes provide structured concurrency for entities with a well-defined lifecycle, this perfectly describes Live Entities in Kord and fits in with the `shutDown` mentality. 

Introducing this would allow users to cancel all the running tasks related to a live entity (assuming they were launched in the correct scope), even those that are not directly related to the event flow.

PR #273, opened 2021-05-01
fix: Resolve issue about the onShutDown function for Live entities

Distractic:
# Context

Currently, when using the method `onShutDown` of LiveEntity classes, the method only intercept the shutdown by events, not manually 

### Manual
```kt
val message = // create message
val live = message.live()
live.onShutDown {
   // Never called
}
live.shutdown()
```

### Event
```kt
val message = // create message
val live = message.live()
live.onShutDown {
   // Is called
}
message.delete()
```

## Now

This issue is resolve with the attribut `onShutDown` present in _AbstractLiveKordEntity_

```kt
val message = // create message
val live = message.live()
live.onShutDown {
   // Is called
}

message.delete()
//or
live.shutdown()
```

### DIFF
--- a/core/src/main/kotlin/live/LiveGuild.kt
+++ b/core/src/main/kotlin/live/LiveGuild.kt
@@ -21,7 +21,7 @@ import kotlinx.coroutines.Dispatchers
 
 @KordPreview
 fun Guild.live(dispatcher: CoroutineDispatcher = Dispatchers.Default): LiveGuild =
-    LiveGuild(dispatcher, this)
+    LiveGuild(this, dispatcher)
 
 @KordPreview
 inline fun Guild.live(dispatcher: CoroutineDispatcher = Dispatchers.Default, block: LiveGuild.() -> Unit) =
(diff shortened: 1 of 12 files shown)


HopeBaron:
You shouldn't modify the behavior of the onShutdown function in this manner

See #270, The proposed solution is to provide `LiveEntity`s with their own coroutine scope so they can manage their own lifecycle.
The user should be able to provide a dispatcher, that we will use for the entity to create a supervisor job which manages coroutines launched under it (children).

here is a Mock about how this should look like:

```kt
interface LiveKordEntity : KordEntity, CoroutineScope

@KordPreview
abstract class AbstractLiveKordEntity(dispatcher: CoroutineDispatcher) : LiveKordEntity {
     ...

    override val coroutineContext: CoroutineContext = dispatcher + SupervisorJob()

    @Suppress("EXPERIMENTAL_API_USAGE")
    override val events: Flow<Event>
        get() = kord.events
            .takeWhile { isActive }
            ...
   
}


inline fun <reified T : Event> LiveKordEntity.on(scope: CoroutineScope = this, noinline consumer: suspend (T) -> Unit) =
   ...
   
 ```

Issue #147, opened 2020-12-19
Provide full entities on events


BartArys:
Kord's core entities are designed to optimize caching; instead of a [`Guild`](https://github.com/kordlib/kord/blob/master/core/src/main/kotlin/entity/Guild.kt) keeping a reference to its [`Channels`](https://github.com/kordlib/kord/blob/master/common/src/main/kotlin/entity/DiscordGuild.kt#L128), it'll keep a reference to the [channel IDs](https://github.com/kordlib/kord/blob/master/core/src/main/kotlin/cache/data/GuildData.kt#L44), and [query the `EntitySupplier`](https://github.com/kordlib/kord/blob/master/core/src/main/kotlin/behavior/GuildBehavior.kt#L79-L80) to fetch them. This approach makes it easy for us to keep the cache up to date; every entity is only once in the cache for modifications, insertions and removals.

This does come with some issues however: Discord's gateway events often do provide a 'full' entity (e.g. Guild creates will contain full channels instead of core's channel IDs, as well as full members, roles, etc). For users with cache this means extra queries for data that was originally present. For users running without cache this means that data is gone completely and they have to resort to retrieval via REST, which is a major inefficiency.

An attempt to mitigate this can be seen in [`Message`](https://github.com/kordlib/kord/blob/master/core/src/main/kotlin/entity/Message.kt) objects, which have a reference to a [full user object for their authors](https://github.com/kordlib/kord/blob/master/core/src/main/kotlin/cache/data/MessageData.kt#L15) instead of delegating retrieval to the cache. Since authors are such an important field in the entity, we decided not to separate it from the message itself, even though this does 'corrupt' the cache.

To mitigate this, we should start by providing 'full' entities in events whenever possible. These new entities should be incorporated into the current hierarchy (Behavior -> cache optimized Entity -> Full Entity) as much as possible.  While this does mean a fair amount of extra classes, I believe this to be a worthwhile change for both cache and non-cache use cases, which should translate in some performance gains.

HopeBaron:
However, this would apply to data we know that is not going to change (or are less likely to), or would we transform the cache to have computed properties that would query the cached entity each time we want it? Taking into consideration the frequent hits of the cache in a typical scenario, I don't think that's the way we are planning to solve it.

We can apply this to what we think is a critical entity (e.g: message and author).

PR #283, opened 2021-05-12
Message interaction

HopeBaron:


core/src/main/kotlin/cache/data/MessageInteractionData.kt
```diff
@@ -0,0 +1,24 @@
...
+import dev.kord.core.cache.data.toData
+import kotlinx.serialization.Serializable
+
+@KordPreview
+@Serializable
+data class MessageInteractionData(
+    val id:Snowflake,
+    val type:InteractionType,
+    val name:String,
+    val user:UserData
```
BartArys, line 17: Flatten this out to the user id

core/src/main/kotlin/entity/interaction/MessageInteraction.kt
```diff
@@ -0,0 +1,17 @@
...
+import dev.kord.common.annotation.KordPreview
+import dev.kord.common.entity.InteractionType
+import dev.kord.common.entity.Snowflake
+import dev.kord.core.Kord
+import dev.kord.core.entity.KordEntity
+import dev.kord.core.entity.User
+
+@KordPreview
+class MessageInteraction(val data: MessageInteractionData, override val kord: Kord): KordEntity {
+    override val id: Snowflake get() = data.id
```
BartArys, line 13: This is the interactions's ID, right? Is there any kind of behavior/entity we can get from this?

core/src/main/kotlin/entity/interaction/MessageInteraction.kt
```diff
@@ -0,0 +1,17 @@
...
+    val name: String get() = data.name
+    val user: User get() = User(data.user, kord)
+    val type: InteractionType get() = data.type
+}
```
BartArys, line 17: Consider documenting this

core/src/main/kotlin/entity/Message.kt
```diff
@@ -174,6 +178,9 @@ class Message(
      */
     val mentionedUserBehaviors: Set<UserBehavior> get() = data.mentions.map { UserBehavior(it, kord) }.toSet()
 
+    @KordPreview
+    val interaction: MessageInteraction? get() = data.messageInteraction.mapNullable { MessageInteraction(it, kord) }.value
```
BartArys, line 182: Consider documenting this

BartArys (changes requested):

core/src/main/kotlin/entity/interaction/MessageInteraction.kt
HopeBaron: there isn't but since it had an ID and requires Kord for the user object that it would be fine to consider it a KordEntity

core/src/main/kotlin/cache/data/MessageInteractionData.kt
HopeBaron: I applied the user data based on #147.
But  I guess that's not the scope for this PR will turn this into the user id.

core/src/main/kotlin/entity/interaction/MessageInteraction.kt
```diff
@@ -5,13 +5,36 @@ import dev.kord.common.annotation.KordPreview
...
+    /**
+     * 	the [name][ApplicationCommand.name] of the [ApplicationCommand] that triggered this message.
+     */
     val name: String get() = data.name
-    val user: User get() = User(data.user, kord)
+
+    /**
+     * The [UserBehavior] of the [user][Interaction.user] who invoked the [Interaction]
+     */
+    val user: UserBehavior get() = UserBehavior(data.id, kord)
```
BartArys, line 34: This needs the usual `getUser` and `getUserOrNull`

BartArys:
Additionally, the user of the MessageInteraction should be cached

core/src/main/kotlin/entity/interaction/MessageInteraction.kt
```diff
@@ -37,4 +44,24 @@ class MessageInteraction(
...
+
+    /**
+     * Requests to get the user of this interaction message,
+     * returns null if the [User] isn't present.
+     *
+     * @throws [RequestException] if anything went wrong during the request.
+     */
+    suspend fun getUserOrNull(): User? = supplier.getUserOrNull(user.id)
+
+    override fun withStrategy(strategy: EntitySupplyStrategy<*>): Strategizable {
```
BartArys, line 64:
```suggestion
    override fun withStrategy(strategy: EntitySupplyStrategy<*>): MessageInteraction {
```

BartArys (approved):

PR #286, opened 2021-05-12
 Properly decode null in gateway events

DRSchlaubi:
This resolves an issue with kx.ser causing de serialization of events that end with a `null` to fail

### DIFF
--- a/gateway/src/main/kotlin/Event.kt
+++ b/gateway/src/main/kotlin/Event.kt
@@ -8,10 +8,7 @@ import dev.kord.common.entity.optional.OptionalSnowflake
 import kotlinx.serialization.*
 import kotlinx.serialization.builtins.nullable
 import kotlinx.serialization.builtins.serializer
-import kotlinx.serialization.descriptors.PrimitiveKind
-import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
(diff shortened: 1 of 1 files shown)


gateway/src/main/kotlin/Event.kt
```diff
@@ -31,7 +28,10 @@ private object NullDecoder : DeserializationStrategy<Nothing?> {
 
     @OptIn(ExperimentalSerializationApi::class)
     override fun deserialize(decoder: Decoder): Nothing? {
-        return decoder.decodeNull()
+        // decodeNull() doesn't consume the literal null therefore parsing doesn't end and in e.g. a heartbeat event
+        // the null gets parsed as a key
+        decoder.decodeNotNullMark()
+        return null
```
BartArys, line 34: We should still call `decoder.decodeNull()` for the sake of correctness, even if does literally nothing.
```suggestion
        return decoder.decodeNull()
```

BartArys (commented):
Thanks for taking your time and figuring this one out! It's greatly appreciated.

gateway/src/main/kotlin/Event.kt
DRSchlaubi: Doing nothing sounds fun I like it

BartArys (approved):

PR #287, opened 2021-05-13
Add buttons to Activity

DRSchlaubi:
This one is a little odd because discord/discord-api-docs#2219 added more than just buttons, but buttons are the only one missing that got merged into this: https://discord.com/developers/docs/topics/gateway#activity-object


BartArys (approved):

HopeBaron:
Should we make them inline once we get on 1.5?

BartArys:
> 
> 
> Should we make them inline once we get on 1.5?

If they don't do anything then I'm not sure what the point is. Anyhow, not related to this PR.

PR #294, opened 2021-05-15
Interaction create improvements

HopeBaron:
* documents core interaction create event.
* Fixes a typo in the ephemeralack method.


core/src/main/kotlin/event/interaction/InteractionCreate.kt
```diff
@@ -2,9 +2,28 @@ package dev.kord.core.event.interaction
...
+ * The event should be acknowledged using one of the following methods:
+ * * [acknowledgeEphemeral][Interaction.acknowledgeEphemeral] - acknowledges an interaction ephemerally.
+ * * [acknowledgePublic][Interaction.acknowledgePublic] - acknowledges an interaction in public.
+ * * [respondPublic][Interaction.respondPublic] - same as public acknowledgement, but an immediate result (message) can be supplied.
+ * * [respondEphemeral][Interaction.respondEphemeral] - same as ephemeral acknowledgement, but an immediate result (message) can be supplied.
+ *
+ * Once an interaction has been acknowledged,
+ * you can use the [PublicInteractionResponseBehavior.followUp] or [EphemeralInteractionResponseBehavior.followUp].
+ *
+ * The resulting followup message and its methods may defer based on which method is used.
```
BartArys, line 25: Not sure this has much use without any in-depth explanation, it's probably better suited for the actual methods that do change behavior.

HopeBaron: Will go all possible cases :+1:

BartArys (approved):

PR #297, opened 2021-05-16
Migrate to kotlinx-datetime

### LINKED ISSUE #43: Use kotlinx-datetime
The madlads over at Jetbrains just released [another banger library](https://github.com/Kotlin/kotlinx-datetime) that brings Java's DateTime API to kotlin-multiplatform. Since our project's only real remaining dependency on the Java stdlib is the datetime API we can use this to yeet that stuff out of there and then realize that we just imported megabytes of typealiases.  It'll be grand.

Note that kotlinx-datetime does **not** typealias java.time.  As such, we should leave the external API using java time  intact, simply providing kotlin's version as an alternative, and remove java time as an internal dependency.

DRSchlaubi:
~~As `kotlinx-datetime` version `0.2.0` actually fixed most of the issues we had in #43 I migrated most of the code
The only dependency we still have on `java.time` is the EmbedBuilder needing to convert from an `Instant` to ISO which is not possible afaik (See Kotlin/kotlinx-datetime#116)
All other usages to `java.time` have been replaced~~ NVM

Fixes #43

### DIFF
--- a/rest/src/main/kotlin/json/request/MessageRequests.kt
+++ b/rest/src/main/kotlin/json/request/MessageRequests.kt
@@ -5,6 +5,7 @@ import dev.kord.common.entity.*
 import dev.kord.common.entity.optional.Optional
 import dev.kord.common.entity.optional.OptionalBoolean
 import dev.kord.common.entity.optional.OptionalInt
+import kotlinx.datetime.Instant
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
@@ -38,7 +39,7 @@ data class EmbedRequest(
     val type: Optional<String> = Optional.Missing(),
     val description: Optional<String> = Optional.Missing(),
     val url: Optional<String> = Optional.Missing(),
-    val timestamp: Optional<String> = Optional.Missing(),
+    val timestamp: Optional<Instant> = Optional.Missing(),
     val color: Optional<Color> = Optional.Missing(),
     val footer: Optional<EmbedFooterRequest> = Optional.Missing(),
     val image: Optional<EmbedImageRequest> = Optional.Missing(),
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -4,6 +4,7 @@ object Versions {
     const val ktor = "1.5.3"
     const val kotlinxCoroutines = "1.5.0-RC"
     const val kotlinLogging = "2.0.4"
+    const val dateTime = "0.2.0"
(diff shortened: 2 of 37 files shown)


BartArys (approved):

rest/src/main/kotlin/json/request/MessageRequests.kt
```diff
@@ -38,7 +39,7 @@ data class EmbedRequest(
     val type: Optional<String> = Optional.Missing(),
     val description: Optional<String> = Optional.Missing(),
     val url: Optional<String> = Optional.Missing(),
-    val timestamp: Optional<String> = Optional.Missing(),
+    val timestamp: Optional<Instant> = Optional.Missing(),
```
HopeBaron, line 42: Why did this become an instant?

DRSchlaubi: Because it can be serialized automatically now I talked about this with @BartArys

BartArys: As mentioned, kx.dt's classes are serializable. Considering the amount of API we're breaking by using kx.dt this seemed like an appropriate change and one that improves the intention of the field. 

This should have been discussed in the PR though instead of on Discord, to avoid the confusion we've run into.

HopeBaron: That being the case, why didn't we change other Instant feilds from String to Instants; or did I miss something out?
In other words, why is this the only changed field

DRSchlaubi: Because this is the only Instant field we send which is an ISO string

HopeBaron: Members have joined at field 
Synced at for integrations... etc

BartArys: This is specifically about data we're *sending*. The original design decision surround serializing incoming ISO 8601 fields was that it was an unnecessary waste of time to parse a value that in most use-cases won't be read anyhow. This reasoning does not apply to entities we're sending to Discord, since we're forced to encode the data one way or another if it is present.

DRSchlaubi: > Members have joined at field
> Synced at for integrations... etc

Carefull hope. Almost everything else doesn't use ISO, but unix timestamps

HopeBaron: What Bart mentioned was what I was missing
I was wondering why we switched from stringe when we agreed to use strings since theyvare unlikely to be hit

HopeBaron (approved):

PR #298, opened 2021-05-20
fix: Resolve issue about several calls for update method

Distractic:
# Context

When an event is received and is validated by the filter, for each job children, the method `update` is called.

## Example

For the example, we have a simple live entity class like that : 
```kt
class MyLiveEntity(kord: Kord) : AbstractLiveKordEntity(kord, Dispatchers.Default) {

        override val id: Snowflake = randomId()

        override fun filter(event: Event): Boolean =  event is BanAddEvent

        override fun update(event: Event) {
            when (event) {
                is BanAddEvent -> println("I'm a ban add event")
            }
        }
    }
```

This is the issue, we have several children job, each job manage a type of event, but currently, for each job, the method "update"' is called, however, it should only be called once

```kt
live.on<BanAddEvent> { // 1 user job
   // action
}

live.on<GuildDeleteEvent> { // 2 user job
   // never called
}

// add the job who is created during initialization of live entity

sendEvent(banAddEvent) // A event valid for the filter is sent
// There are 3 jobs, so the update method is called 3 times, so the print will be executed 3 times
```

# Fix

Thanks to the initial job created, the `update` function is only called by it, and all user jobs (from the method `on<..> {..}`) never call the `update` method

### DIFF
--- a/core/src/main/kotlin/live/LiveKordEntity.kt
+++ b/core/src/main/kotlin/live/LiveKordEntity.kt
@@ -31,17 +31,12 @@ abstract class AbstractLiveKordEntity(final override val kord: Kord, dispatcher:
     override val coroutineContext: CoroutineContext = dispatcher + SupervisorJob(kord.coroutineContext.job)
 
     @Suppress("EXPERIMENTAL_API_USAGE")
-    final override val events: Flow<Event>
-        get() = kord.events
-            .takeWhile { isActive }
-            .filter { filter(it) }
+    final override val events: SharedFlow<Event> =
+        kord.events.filter { filter(it) }.onEach { update(it) }.shareIn(this, SharingStarted.Eagerly)
+
 
     protected abstract fun filter(event: Event): Boolean
     protected abstract fun update(event: Event)
-
-    init {
-        events.onEach { update(it) }.launchIn(this)
-    }
 
     override fun shutDown(cause: CancellationException) = cancel(cause)
 }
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/live/LiveKordEntity.kt
```diff
@@ -32,20 +30,17 @@ abstract class AbstractLiveKordEntity(final override val kord: Kord, dispatcher:
...
             .filter { filter(it) }
-            .onEach { mutex.withLock { update(it) } }
 
     protected abstract fun filter(event: Event): Boolean
     protected abstract fun update(event: Event)
 
     init {
-        events.launchIn(this)
+        events.onEach { update(it) }.launchIn(this)
     }
```
BartArys, line 44: This is indeed a problem, but I don't think your solution is the best possible.

There is a possibility that any new event listener fires *before* `update` is finished, this can be a cause of confusion since the the most obvious behaviour is that the event being listened for is already handled by the entity itself.

What we want ideally is the following:

``` 
                                          listener
                                        /
kord.events -> update -> entity.events - listener
                                        \
                                          listener
```

and kx.coroutines does offer a behaviour like that: A [Shared Flow](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-shared-flow/).

If we were to change the event flow to this:

```kotlin
final override val events: SharedFlow<Event> = kord.events.onEach { update(it) }.shareIn(this, SharingStarted.Eagerly)
```

we have:
* a flow that is eagerly subscribed
* a flow that only handles each even internally
* a flow that is cancelled when the scope dies

PR #299, opened 2021-05-23
Update to Kotlin 1.5

BartArys:


DRSchlaubi:
Ktor 1.6 is there so let's update that here too before making 2 new versions with many breaking changes

PR #303, opened 2021-05-26
Support buttons/components

DRSchlaubi:
# Tasks:
- [x] rest entity representations
- [x] core representations
- [x] support to receive
- [x] Builders to create components


common/src/main/kotlin/entity/Interactions.kt
```diff
@@ -170,18 +171,163 @@ data class DiscordInteraction(
...
 @KordPreview
 sealed class InteractionType(val type: Int) {
     object Ping : InteractionType(1)
     object ApplicationCommand : InteractionType(2)
+
+    /**
+     * don't trust the docs:
+     *
+     * this type exists and is needed for components even though it's not documented
+     */
```
BartArys, line 323: Don't put this as a KDoc

BartArys (changes requested):

rest/src/main/kotlin/builder/components/ButtonBuilder.kt
```diff
@@ -0,0 +1,30 @@
...
+    var customId by ::_customId.delegate()
+    private var _url: Optional<String> = Optional.Missing()
+    var url by ::_url.delegate()
+    private var _disabled: OptionalBoolean = OptionalBoolean.Missing
+    var disabled by ::_disabled.delegate()
+
+    fun build(): DiscordComponent = DiscordComponent(
+        ComponentType.Button, _style, _label, _emoji, _customId, _url, _disabled
+    )
+}
```
BartArys, line 30: Considering the differing requirements between link buttons and action(?) buttons, would it not make more sense to have a builder for each individually? It makes little sense to be able to both set `url` and `customId`.

rest/src/main/kotlin/builder/components/CompositeComponentBuilder.kt
```diff
@@ -0,0 +1,34 @@
...
+import dev.kord.common.entity.ComponentType
+import dev.kord.common.entity.DiscordComponent
+import dev.kord.common.entity.optional.optional
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+
+@KordPreview
+class CompositeComponentBuilder {
+    var components: MutableList<DiscordComponent> = mutableListOf()
```
BartArys, line 13: Consider making this a val instead. Certain operator functions don't play nice with vars of mutable lists.

rest/src/main/kotlin/builder/message/MessageCreateBuilder.kt
```diff
@@ -40,6 +39,9 @@ class MessageCreateBuilder : RequestBuilder<MultipartMessageCreateRequest> {
     private var _allowedMentions: Optional<AllowedMentionsBuilder> = Optional.Missing()
     var allowedMentions: AllowedMentionsBuilder? by ::_allowedMentions.delegate()
 
+    private var _components: Optional<List<DiscordComponent>> = Optional.Missing()
+    var components: List<DiscordComponent>? by ::_components.delegate()
```
BartArys, line 43: consider making this a mutable list instead.

rest/src/main/kotlin/builder/components/CompositeComponentBuilder.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    @OptIn(ExperimentalContracts::class)
+    inline fun actionRow(builder: CompositeComponentBuilder.() -> Unit) {
+        contract {
+            callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+        }
+
+        components.add(ActionRow(builder))
+    }
```
BartArys, line 22: It does not make sense for this to be a `CompositeComponentBuilder`, the discord docs state that nesting action rows inside action rows is forbidden. <https://discord.com/developers/docs/interactions/message-components#actionrow>

BartArys (changes requested):

common/src/main/kotlin/entity/DiscordMessage.kt
```diff
@@ -99,6 +99,13 @@ data class DiscordMessage(
     val stickers: Optional<List<DiscordMessageSticker>> = Optional.Missing(),
     @SerialName("referenced_message")
     val referencedMessage: Optional<DiscordMessage?> = Optional.Missing(),
+    val interaction: Optional<DiscordInteraction> = Optional.Missing(),
+
+    /*
+     * don't trust the docs:
+     * This is a list even though the docs say it's a component
+     */
```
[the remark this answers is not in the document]
DRSchlaubi: the other don't trust the docs all use /* as well

rest/src/main/kotlin/builder/components/ActionRowBuilder.kt
```diff
@@ -0,0 +1,19 @@
...
+@OptIn(ExperimentalContracts::class)
+@Suppress("FunctionName")
+@KordPreview
+inline fun ActionRow(builder: CompositeComponentBuilder.() -> Unit): DiscordComponent {
+    contract {
+        callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+    }
+
+    return CompositeComponentBuilder().apply(builder).toComponent(ComponentType.ActionRow)
+}
```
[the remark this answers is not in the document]
DRSchlaubi: originally i wanted to make a own ActionRowBuilder class and then this happened

common/src/main/kotlin/entity/DiscordMessage.kt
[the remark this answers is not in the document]
BartArys: You're right, I completely missed the use of `/*` instead of `/**`, disregard these comments then.

rest/src/main/kotlin/builder/message/MessageCreateBuilder.kt
```diff
@@ -40,6 +40,9 @@ class MessageCreateBuilder : RequestBuilder<MultipartMessageCreateRequest> {
     private var _allowedMentions: Optional<AllowedMentionsBuilder> = Optional.Missing()
     var allowedMentions: AllowedMentionsBuilder? by ::_allowedMentions.delegate()
 
+    private var _components: Optional<MutableList<DiscordComponent>> = Optional.Missing()
+    var components: MutableList<DiscordComponent>? by ::_components.delegate()
```
BartArys, line 44: Make this a val, empty list implies unset

DRSchlaubi: this breaks the components function

BartArys: It doesn't, it's a mutable list, you can just add to it

rest/src/main/kotlin/builder/components/CompositeActionRowBuilder.kt
```diff
@@ -0,0 +1,8 @@
+package dev.kord.rest.builder.components
+
+import dev.kord.common.annotation.KordPreview
+
+@KordPreview
+interface CompositeActionRowBuilder {
+    fun actionRow(builder: CompositeButtonBuilder.() -> Unit)
+}
```
BartArys, line 8: Making this non-inline prevents people from calling suspending functions in the lambda, we generally want to avoid that.

Additionally, there's no way to access, modify or delete any of the already created buttons/rows.

rest/src/main/kotlin/builder/components/CompositeButtonBuilder.kt
```diff
@@ -0,0 +1,8 @@
+package dev.kord.rest.builder.components
+
+import dev.kord.common.annotation.KordPreview
+
+@KordPreview
+interface CompositeButtonBuilder {
+    fun button(builder: ButtonBuilder.() -> Unit)
+}
```
BartArys, line 8: Making this non-inline prevents people from calling suspending functions in the lambda, we generally want to avoid that.

Additionally, there's no way to access, modify or delete any of the already created buttons.

rest/src/main/kotlin/builder/message/MessageCreateBuilder.kt
```diff
@@ -82,6 +85,16 @@ class MessageCreateBuilder : RequestBuilder<MultipartMessageCreateRequest> {
...
 
+    @OptIn(ExperimentalContracts::class)
+    @KordPreview
+    inline fun components(builder: CompositeActionRowBuilder.() -> Unit) {
+        contract {
+            callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+        }
+
+        components = CompositeComponentBuilder().apply(builder).components
+    }
```
BartArys, line 96: Kord's builders should generally be additive, previous values shouldn't be cleared.

BartArys (commented):
Not a fan of how the added builders currently work. Ideally, builders don't *build* anything until the entire structure is built, and allow the user to interact with/modify/delete everything up until that point. Look at how `EmbedBuilder` is implemented to get a better idea of what I'm looking for.

rest/src/main/kotlin/builder/components/CompositeButtonBuilder.kt
DRSchlaubi: The problem here is that I can't inline this as it is an interface, which makes this whole builder thing not really ideal either way

Earlier you could add components at the wrong places, but now you can create confusing code.

```kotlin
createMessage {
  components {
    actionRow {
      actionRow {
      }
   }
}
```

BartArys: You can inline them if you expose a mutable list with all stored components, but the issue there is inheritance. What we have is a builder that can accept components of any type (message) and a builder that can accept components of any type except action rows (action rows). 

"Everything but X" is hard to represent in Kotlin's type system, especially if you then have to reintroduce that type, if you'd go that route it's best to split up the types in separate lists. Have action row builders contain a list of button builders, and the message component builder inherit that but also have a list of action row builders in addition. 

I'm generally not a fan of inheritance in builders like this, it's not unreasonable to assume that Discord would introduce a type that suddenly breaks the rules (like something that can only sit in a row but not in at the root of a message's components). That said, if you can make it work I'll approve it, we break the API whenever Discord adds something anyhow.

rest/src/main/kotlin/builder/interaction/PublicInteractionBuilder.kt
```diff
@@ -58,6 +66,16 @@ class PublicInteractionResponseCreateBuilder :
...
 
+    @OptIn(ExperimentalContracts::class)
+    @KordPreview
+    inline fun components(builder: ActionRowBuilder.() -> Unit) {
+        contract {
+            callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+        }
+
+        components = mutableListOf(ActionRowBuilder().apply(builder).build())
+    }
```
BartArys, line 77: * can messages contain buttons without action rows? If so, this is a misnomer. If not, the `components` field should not be able to accept buttons as it currently does.

* this DSL does not allow you to add more than one component, while messages can contain up to 5 action rows.

rest/src/main/kotlin/builder/interaction/PublicInteractionBuilder.kt
```diff
@@ -36,6 +38,12 @@ class PublicInteractionResponseCreateBuilder :
     private var _tts: OptionalBoolean = OptionalBoolean.Missing
     var tts: Boolean? by ::_tts.delegate()
 
+    @KordPreview
+    private var _components: Optional<MutableList<DiscordComponent>> = Optional.Missing()
+
+    @KordPreview
+    var components: MutableList<DiscordComponent>? by ::_components.delegate()
```
BartArys, line 45: * `var: MutableList` should be replaced with `val: MutableList`

* `Optional<List>` should be avoided because `Value(emptyList())` doesn't make sense for an optional field. Remove the optional and assume that empty means missing.

BartArys (changes requested):
* I'm going to point out again that storing built `DiscordComponent`s in these builders is counter to what Kord does generally. Store their builder forms instead.

* It often slips my mind, but builders should be annotated with `@KordDsl`.

* PublicInteractionResponseCreateBuilder ignores its components when building.

* PublicInteractionResponseModifyBuilder should also support components.

core/src/main/kotlin/builder/components/ButtonBuilderExtensions.kt
```diff
@@ -0,0 +1,17 @@
...
+
+@KordPreview
+fun ButtonBuilder.emoji(emoji: ReactionEmoji) {
+    this.emoji = DiscordPartialEmoji(name = emoji.name, id = null)
+}
+
+@KordPreview
+fun ButtonBuilder.emoji(emoji: GuildEmoji) {
+    this.emoji = DiscordPartialEmoji(id = emoji.id, name = null)
+}
```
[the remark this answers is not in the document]
DRSchlaubi: I can't because they're in seperate modules
ButtonBuilder is rest
Emojis are core

rest/src/main/kotlin/builder/interaction/PublicInteractionBuilder.kt
DRSchlaubi: > can messages contain buttons without action rows?
No

> If so, this is a misnomer.

Why it allows you to add components even though you can only add one

> If not, the components field should not be able to accept buttons as it currently does.

How does it accept buttons? it only accepts an action row

DRSchlaubi:
> I'm going to point out again that storing built DiscordComponents in these builders is counter to what Kord does generally. Store their builder forms instead.

There is no `InteractionBuilder` so this is hard

rest/src/main/kotlin/builder/interaction/PublicInteractionBuilder.kt
BartArys: > Why it allows you to add components even though you can only add one
> How does it accept buttons? it only accepts an action row

Not this DSL, but the builder's `components` accepts built `DiscordComponent`, which is also what buttons build to.

core/src/main/kotlin/builder/components/ButtonBuilderExtensions.kt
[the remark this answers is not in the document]
BartArys: Right, stupid of me, this is absolutely fine then.

BartArys:
> 
> 
> > I'm going to point out again that storing built DiscordComponents in these builders is counter to what Kord does generally. Store their builder forms instead.
> 
> There is no `InteractionBuilder` so this is hard

I'm not sure I understand, how are components related to interaction builders?

BartArys (changes requested):

PR #304, opened 2021-05-27
feat: Add possibility to define Job parent to create live entities

Distractic:
In some case, it's useful to define the `SupervisorJob` different than kord's job.

```
kord -> MyObject -> LiveEntity
```
_MyObject is a CoroutineScope and can be the parent of the live entity_

So for this reason, a specific job can be sent through constructor of live entities


BartArys:
At this point, isn't it more flexible to accept a `CoroutineScope` instead?

BartArys:
This looks good right now, but it's probably better to inherit Kord's entire context to inherit the dispatcher as well. `Dispatchers.Default` might not be Kord's dispatcher.
```kotlin
    coroutineScope: CoroutineScope = CoroutineScope(kord.coroutineContext + SupervisorJob(kord.coroutineContext.job))
```

HopeBaron (approved):

PR #310, opened 2021-06-11
Improve slash command API and add support for components

BartArys:
Cleans up some parts of the slash command API, as well as add parts of the API that have been introduced since then

notable items are :
* the addition of components/buttons ( thanks to @DRSchlaubi )
* a reworking of options now that they have a type field
* the addition of mentionable options


core/src/main/kotlin/behavior/interaction/ComponentInteractionBehavior.kt
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * There is no noticeable difference between this and [acknowledgeEphemeralDeferredMessageUpdate]
+     * when it comes to acknowledging the interaction, both functions can be called
+     * on public and ephemeral messages. The only difference is in the **followUp** calls,
+     * which will become public or ephemeral respectively.
+     */
+    @OptIn(ExperimentalContracts::class)
+    suspend fun acknowledgePublicDeferredMessageUpdate(): PublicInteractionResponseBehavior {
+        val request = InteractionResponseCreateRequest(
+            type = InteractionResponseType.DeferredUpdateMessage
```
HopeBaron, line 36: I see we have changed how we handle requests.
In slash commands I've made the toRequest function determine what's the type response based on input

HopeBaron (changes requested):
I've gone through the PR but it seems to need a rebase since it has lots of commits from other branches (or they were merged to this branch instead of 0.7.x.

core/src/main/kotlin/cache/data/ActivityData.kt
```diff
@@ -22,6 +22,7 @@ data class ActivityData(
     val secrets: Optional<DiscordActivitySecrets> = Optional.Missing(),
     val instance: OptionalBoolean = OptionalBoolean.Missing,
     val flags: Optional<ActivityFlags> = Optional.Missing(),
+    val buttons: Optional<List<String>> = Optional.Missing()
```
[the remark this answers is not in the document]
BartArys: there are:
<https://discord.com/developers/docs/topics/gateway#activity-object-activity-structure> 
<https://discord.com/developers/docs/topics/gateway#activity-object-activity-buttons>

core/src/main/kotlin/behavior/interaction/ComponentInteractionBehavior.kt
BartArys: I'm aware, I didn't really see the use of a builder if we can build the request directly anyhow.

core/src/main/kotlin/cache/data/ActivityData.kt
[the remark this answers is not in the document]
HopeBaron: Seems like I mixed up it with the game-sdk activities.

core/src/main/kotlin/behavior/interaction/ComponentInteractionBehavior.kt
HopeBaron: if that's the case then it'd be better to change the application command to match components

common/src/main/kotlin/entity/Interactions.kt
```diff
@@ -157,32 +156,72 @@ data class ResolvedObjects(
...
+class DiscordInteraction(
     val id: Snowflake,
     @SerialName("application_id")
     val applicationId: Snowflake,
-    val type: InteractionType,
-    val data: DiscordApplicationCommandInteractionData,
+    val data: InteractionCallbackData,
     @SerialName("guild_id")
-    val guildId: OptionalSnowflake = OptionalSnowflake.Missing,
+    val guildId: OptionalSnowflake,
```
HopeBaron, line 165: Optional Missing value seems to be removed here

HopeBaron (changes requested):

buildSrc/build.gradle.kts
```diff
@@ -11,7 +11,7 @@ repositories {
 }
 
 dependencies {
-    implementation(kotlin("gradle-plugin-api", version = "1.4.21"))
+    implementation(kotlin("gradle-plugin-api", version = "1.5.0"))
```
DRSchlaubi, line 14: U sure this works? buildSrc is really slow at upgrading the Kotlin version and if it's to new it get's mad

BartArys: [The version seems to exist](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-gradle-plugin-api), the CI doesn't fail because of it and building it locally with `gradle clean build --no-build-cache` doesn't seem to produce any errors. 

If you have any issues building this yourself because of this, please do share a stacktrace and I'll look into it.

HopeBaron (approved):

buildSrc/build.gradle.kts
DRSchlaubi: that's not the problem the version exist as soon as the new gradle plugin version is out
We need to wait for the gradle release adding support for the Kotlin version but if the build works for you i guess it works now

PR #321, opened 2021-06-17
Fix followups method signiture

HopeBaron:


gdude2002 (approved):

BartArys (approved):
We should probably find a solution that's better integrated with the type system (we *can* know when the response would be ephemeral or public), but this'll do for now.

HopeBaron:
Agreed :+1:

PR #337, opened 2021-06-30
Fix master gateway average ping conversion

BartArys:
The average ping is erroneously converting the average millisecond ping to microseconds, this PR fixes that.


HopeBaron (approved):

HopeBaron:
Although we are using milliseconds everywhere, shouldn't we change that to microseconds as well?

BartArys:
> 
>
> Although we are using milliseconds everywhere, shouldn't we change that to microseconds as well?

I doubt anyone will ever deal with sub ms latency. For rounding I thought it'd be better to store the average as a more precise unit.

PR #338, opened 2021-07-03
Add a cache aware rest entity supplier

HopeBaron:
* Adds a cache-aware rest entity supplier that will store it's fetched elements in the target cache.
* Adds a strategy to use the cache-aware rest entity supplier.
* Adds a strategy to use cache with a cache-aware rest entity supplier fallback.

### DIFF
--- a/core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
+++ b/core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
@@ -7,27 +7,29 @@ import dev.kord.core.Kord
 import dev.kord.core.entity.*
 import dev.kord.core.entity.channel.Channel
 import dev.kord.core.entity.channel.GuildChannel
-import dev.kord.rest.request.RestRequestException
-import dev.kord.rest.service.RestClient
-import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+
 /**
  * [EntitySupplier] that uses a [RestEntitySupplier] to resolve entities.
  *
  * Resolved entities will always be stored in [cache] if it wasn't null or empty for flows.
  */
-class CacheAwareRestEntitySupplier(val restSupplier: RestEntitySupplier,
-                                   val cache: DataCache,
-                                   val kord: Kord) : EntitySupplier {
+class CacheAwareRestEntitySupplier(
+    val restSupplier: RestEntitySupplier,
+    val cache: DataCache,
+    val kord: Kord
+) : EntitySupplier {
 
 
     override val guilds: Flow<Guild>
-        get() = flow {
-            storeAndEmit(restSupplier.guilds) { it.data }
-        }
+        get() = storeAndEmit(restSupplier.guilds) { it.data }
+
     override val regions: Flow<Region>
-        get() = flow {
-            storeAndEmit(restSupplier.regions) { it.data }
-        }
+        get() = storeAndEmit(restSupplier.regions) { it.data }
+
 
     override suspend fun getGuildOrNull(id: Snowflake): Guild? {
         return storeAndReturn(restSupplier.getGuildOrNull(id)) { it.data }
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/supplier/EntitySupplyStrategy.kt
```diff
@@ -36,6 +37,22 @@ interface EntitySupplyStrategy<T : EntitySupplier> {
             override fun toString(): String = "EntitySupplyStrategy.cache"
         }
 
+
+        /**
+         * A supplier providing a strategy which exclusively uses REST calls to fetch entities.
+         * fetched entities are stored in [Kord's cache][kord.cache].
+         * See [CacheAwareRestEntitySupplier] for more details.
+         */
+        val cacheAwareRest = object : EntitySupplyStrategy<EntitySupplier> {
```
BartArys, line 46: I think 'cache aware' is ambiguous, it's only aware in the sense that it caches entities. Something like `cachingRest` would be clearer imo.

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
```diff
@@ -0,0 +1,171 @@
...
+        val data = this.map { transform(it) }
+        kord.cache.put(data)
+    }
+
+    private inline fun <T, R> storeAndEmit(source: Flow<T>, crossinline transform: (T) -> R): Flow<T> {
+        return flow {
+            source.mapAndStore { transform(it) }
+            emitAll(source)
+        }
+    }
```
BartArys, line 164: This iterates the flow twice, and iterates is before emitting any values. That's a major issue when the source is backed by a rest request (and possibly multiple given a paginator).

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
```diff
@@ -0,0 +1,171 @@
...
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+
+/**
+ * [EntitySupplier] that uses a [RestEntitySupplier] to resolve entities.
+ *
+ * Resolved entities will always be stored in [cache] if it wasn't null or empty for flows.
+ */
+class CacheAwareRestEntitySupplier(
+    val restSupplier: RestEntitySupplier,
```
BartArys, line 21: is it possible to lower the requirement to any supplier? It would make this a more flexible solution.

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
```diff
@@ -0,0 +1,171 @@
...
+    val cache: DataCache,
+    val kord: Kord
```
BartArys, line 23: These properties should probably be private, I don't see anyone interaction with them when using a supplier

BartArys (changes requested):

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
HopeBaron: We can emit all, then store. That way there will be no delay on the call site

core/src/main/kotlin/supplier/EntitySupplyStrategy.kt
HopeBaron: Sounds like a better name, into it!

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
HopeBaron: Hmmm, yeah sounds like someone might want to use their own entity supplier and store in a data cache.
Into it as well!

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
HopeBaron: I've made the requested changes, and switched the method calls around. so that we emit before we store.

core/src/main/kotlin/supplier/CacheAwareRestSupplier.kt
```diff
@@ -158,8 +158,9 @@ class CacheAwareRestEntitySupplier(
 
     private inline fun <T, R> storeAndEmit(source: Flow<T>, crossinline transform: (T) -> R): Flow<T> {
         return flow {
-            source.mapAndStore { transform(it) }
             emitAll(source)
+            source.mapAndStore { transform(it) }
+
         }
```
BartArys: This still doesn't fix the issue, you're iterating the flow twice

BartArys (changes requested):

BartArys (approved):

PR #339, opened 2021-07-03
Implement select menus

BartArys:
This PR adds support for [select menus](https://discord.com/developers/docs/interactions/message-components#select-menus). `Component` and `ComponentInteraction` were broken to add support for them.

### DIFF
--- a/rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
+++ b/rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
@@ -27,38 +27,11 @@ class SelectMenuBuilder(
     val options: MutableList<SelectOptionBuilder> = mutableListOf()
 
     /**
-     * The minimum amount of values that can be accepted. Accepts a range of `[0,25]`.
-     * Defaults to `1`.
-     *
-     * If a value is set that is bigger than [maximumValues], then [maximumValues] will
-     * be updated to that new value.
-     *
-     * Setting [minimumValues] to a number higher than the amount of possible options is not allowed.
+     * The range of values that can be accepted. Accepts any range between [0,25].
+     * Defaults to `1..1`.
      */
-    var minimumValues: Int = 1
-        set(value) {
-            field = value
-            if (value > maximumValues) {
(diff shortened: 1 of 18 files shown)


rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
```diff
@@ -0,0 +1,99 @@
...
+     * If a value is set that is bigger than [maximumValues], then [maximumValues] will
+     * be updated to that new value.
+     *
+     * Setting [minimumValues] to a number higher than the amount of possible options is not allowed.
+     */
+    var minimumValues: Int = 1
+        set(value) {
+            field = value
+            if (value > maximumValues) {
+                maximumValues = value
```
HopeBaron, line 42: Why do we check here instead of throwing

BartArys: For one we generally don't throw in builders. I thought it'd be more intuitive that something like this would work:
```kotlin
selectMenu(id) {
    minValues = 2
    
   option("a", "key.a")
   option("b", "key.b")
   option("c", "key.c")
   option("d","key.d")
}
```

If you think that min and max values should always be set instead when raising the minimum above `1` then I'll remove the behaviour.

HopeBaron: honestly, it is kinda hard to know what would be the best approach, I mean, the request would still be valid
using an IntRange is another possible solution, and finally an exception

HopeBaron: Sounds to me like an IntRange would be safe to use. We take the min value for our min_value and likewise for max. it also ensure that one is equal or less than the others. allows combos like 1..3, 1..1, 0..0, 2.downTo 5 etc..

BartArys: Good idea, I'll apply the suggestion

HopeBaron (approved):

PR #356, opened 2021-08-03
Allow multiple embeds in messages

HopeBaron:
old `embed` field is now deprecated and replaced with `embeds`
https://discord.com/developers/docs/resources/channel#create-message-jsonform-params


HopeBaron:
I wonder what to do with this: https://github.com/kordlib/kord/blob/ea21bdaa27dafe7f66c9fa93b5be00d4b87308e1/core/src/main/kotlin/behavior/channel/MessageChannelBehavior.kt#L272-L278

PR #358, opened 2021-08-03
Thread updates

HopeBaron:
* Moves the list active thread endpoint to the guild service.
* Adds Thread audit log related entries.

Changes from: https://github.com/discord/discord-api-docs/pull/3502


core/src/main/kotlin/behavior/channel/threads/ThreadParentChannelBehavior.kt
```diff
@@ -33,7 +34,7 @@ interface ThreadParentChannelBehavior : TopGuildMessageChannelBehavior {
      * [terminal operators](https://kotlinlang.org/docs/reference/coroutines/flow.html#terminal-flow-operators) instead.
 
      */
-    val activeThreads: Flow<ThreadChannel> get() = supplier.getActiveThreads(id)
+    val activeThreads: Flow<ThreadChannel> get() = guild.activeThreads.filter { it.parentId == id }
```
BartArys, line 37: if you're delegating calls to other entities, you should make sure they're using the same supply strategy

BartArys (approved):

Issue #101, opened 2020-10-28
Status of voice


HopeBaron:
This issue will function as a overview of introducing voice:

- [x] Heart beating
- [x] IP Discovery
- [x] Audio formatting (Lava Player)
- [ ] Resuming
- [ ] UDP connection and audio forwarding to channels

HopeBaron:
The module will be on hold until [ktorio/ktor #1049](https://github.com/ktorio/ktor/issues/1049) gets fixed
Resuming and proper judging on the life-cycle handling would require at least a function UDP socket that's able to forward opus packets. However the current (dirty) code will be pushed to the voice branch and improve the current parts on the go.

VapidLinus:
Will this make it possible to move users to another voice channel, or is that already possible?

HopeBaron:
Hello, This is already possible using `MemberBehavior#edit` where you have the ability to change the voice channel id

DRSchlaubi:
THE ISSUE GOT FIXED WOOOOOOOOOHOOOOOOOOO

NyCodeGHG:
POGGERS

HopeBaron:
The next big thing probably, hopefully I can ship this with minor changes... But we need to reimpelement this from scratch

PR #363, opened 2021-08-14
Voice

lost-illusi0n:
This PR adds a voice module to Kord, capable of transmitting audio to Discord along with checking off the following requirements found in #101.

- [X]  Heart beating
- [x]  IP Discovery
- [X]  Audio formatting (Lava Player)
- [X]  Resuming
- [X]  UDP connection and audio forwarding to channels 

This should be *hopefully* fairly stable, and ready to be tested/updated further.

A quick example using this PR
```kotlin
val player: AudioPlayer = TODO() // create a player with LavaPlayer, and start a track
kord.on<ReadyEvent> {
    val voiceChannel = kord.getChannelOf<VoiceChannel>(Snowflake(813711711977865216))!!

    voiceChannel.connect {
        selfDeaf = true

        audioProvider { AudioFrame.fromData(player.provide()?.data) }
    }
}
```

Resolves kordlib/kord#101

### DIFF
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -7,6 +7,7 @@ object Versions {
     const val dateTime = "0.2.1"
     const val atomicFu = "0.16.1"
     const val binaryCompatibilityValidator = "0.5.0"
+    const val xsala20poly1305 = "0.10.1"
 
     //test deps
     const val kotlinTest = kotlin
@@ -49,6 +50,9 @@ object Dependencies {
     const val `ktor-client-websocket` = "io.ktor:ktor-client-websockets:${Versions.ktor}"
     const val `ktor-client-mock` = "io.ktor:ktor-client-mock:${Versions.ktor}"
     const val `ktor-client-mock-jvm` = "io.ktor:ktor-client-mock-jvm:${Versions.ktor}"
+    const val `ktor-network` = "io.ktor:ktor-network:${Versions.ktor}"
+
+    const val `codahale-xsalsa20poly1305` = "com.codahale:xsalsa20poly1305:${Versions.xsala20poly1305}"
 
     const val `cache-api` = "dev.kord.cache:cache-api"
     const val `cache-map` = "dev.kord.cache:cache-map"
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -1,5 +1,5 @@
 object Versions {
-    const val kotlin = "1.5.10"
+    const val kotlin = "1.5.21"
     const val kotlinxSerialization = "1.2.1"
     const val ktor = "1.6.0"
     const val kotlinxCoroutines = "1.5.0"
--- a/voice/src/main/kotlin/FrameInterceptor.kt
+++ b/voice/src/main/kotlin/FrameInterceptor.kt
@@ -1,5 +1,6 @@
 package dev.kord.voice
 
+import dev.kord.common.annotation.KordVoice
 import dev.kord.gateway.Gateway
 import dev.kord.voice.gateway.Speaking
 import dev.kord.voice.gateway.VoiceGateway
@@ -12,12 +13,14 @@ import kotlin.properties.Delegates
  * @param voiceGateway the underlying [VoiceGateway].
  * @param ssrc the current SSRC retrieved from Discord.
  */
+@KordVoice
 data class FrameInterceptorContext(
     val gateway: Gateway,
     val voiceGateway: VoiceGateway,
     val ssrc: Int,
 )
 
+@KordVoice
 internal class FrameInterceptorContextBuilder {
     var gateway: Gateway by Delegates.notNull()
     var voiceGateway: VoiceGateway by Delegates.notNull()
@@ -26,6 +29,7 @@ internal class FrameInterceptorContextBuilder {
     fun build() = FrameInterceptorContext(gateway, voiceGateway, ssrc)
 }
 
+@KordVoice
 internal inline fun FrameInterceptorContext(builder: FrameInterceptorContextBuilder.() -> Unit) =
     FrameInterceptorContextBuilder().apply(builder).build()
 
@@ -34,6 +38,7 @@ internal inline fun FrameInterceptorContext(builder: FrameInterceptorContextBuil
  *
  * @see DefaultFrameInterceptor
  */
+@KordVoice
 fun interface FrameInterceptor {
     suspend fun intercept(audioFrame: AudioFrame?): AudioFrame?
 }
@@ -46,6 +51,7 @@ fun interface FrameInterceptor {
  * @param connection the voice connection.
  * @param speakingState the speaking state that will be used when there is audio data to be sent. By default, it is microphone-only.
  */
+@KordVoice
 open class DefaultFrameInterceptor(
     protected val context: FrameInterceptorContext,
     private val speakingState: SpeakingFlags = SpeakingFlags { +SpeakingFlag.Microphone }
--- a/voice/src/main/kotlin/FrameInterceptor.kt
+++ b/voice/src/main/kotlin/FrameInterceptor.kt
@@ -21,17 +21,15 @@ data class FrameInterceptorContext(
 )
 
 @KordVoice
-internal class FrameInterceptorContextBuilder {
-    var gateway: Gateway by Delegates.notNull()
-    var voiceGateway: VoiceGateway by Delegates.notNull()
+internal class FrameInterceptorContextBuilder(var gateway: Gateway, var voiceGateway: VoiceGateway) {
     var ssrc: Int by Delegates.notNull()
 
     fun build() = FrameInterceptorContext(gateway, voiceGateway, ssrc)
 }
 
 @KordVoice
-internal inline fun FrameInterceptorContext(builder: FrameInterceptorContextBuilder.() -> Unit) =
-    FrameInterceptorContextBuilder().apply(builder).build()
+internal inline fun FrameInterceptorContext(gateway: Gateway, voiceGateway: VoiceGateway, builder: FrameInterceptorContextBuilder.() -> Unit) =
+    FrameInterceptorContextBuilder(gateway, voiceGateway).apply(builder).build()
 
 /**
  * A interceptor for audio frames before they are sent as packets.
--- a/voice/src/main/kotlin/FrameInterceptor.kt
+++ b/voice/src/main/kotlin/FrameInterceptor.kt
@@ -1,0 +1,80 @@
+package dev.kord.voice
+
+import dev.kord.gateway.Gateway
+import dev.kord.voice.gateway.Speaking
+import dev.kord.voice.gateway.VoiceGateway
+import kotlin.properties.Delegates
+
+/**
+ * Variables that are accessible to any FrameInterceptor through the [VoiceConnection.frameInterceptorFactory].
+ *
+ * @param gateway the gateway that handles the guild this voice connection is connected to.
+ * @param voiceGateway the underlying [VoiceGateway].
+ * @param ssrc the current SSRC retrieved from Discord.
+ */
+data class FrameInterceptorContext(
+    val gateway: Gateway,
+    val voiceGateway: VoiceGateway,
+    val ssrc: Int,
+)
+
+internal class FrameInterceptorContextBuilder {
+    var gateway: Gateway by Delegates.notNull()
+    var voiceGateway: VoiceGateway by Delegates.notNull()
+    var ssrc: Int by Delegates.notNull()
+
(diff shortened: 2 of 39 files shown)


voice/src/main/kotlin/FrameInterceptor.kt
```diff
@@ -0,0 +1,86 @@
...
+data class FrameInterceptorContext(
+    val gateway: Gateway,
+    val voiceGateway: VoiceGateway,
+    val ssrc: Int,
+)
+
+@KordVoice
+internal class FrameInterceptorContextBuilder {
+    var gateway: Gateway by Delegates.notNull()
+    var voiceGateway: VoiceGateway by Delegates.notNull()
```
HopeBaron, line 26: provide those in the constructor of the class if they are mandatory while still allowing them to be edited inside the builder using `var`

voice/src/main/kotlin/VoiceConnection.kt
```diff
@@ -0,0 +1,146 @@
...
+     */
+    suspend fun leave() {
+        disconnect()
+
+        gateway.send(
+            UpdateVoiceStatus(
+                guildId = data.guildId,
+                channelId = null,
+                selfMute = false,
+                selfDeaf = false
```
HopeBaron, line 103: are those mandatory?

voice/src/main/kotlin/VoiceConnectionBuilder.kt
```diff
@@ -0,0 +1,129 @@
...
+import mu.KotlinLogging
+
+private val logger = KotlinLogging.logger { }
+
+@KordVoice
+class VoiceConnectionBuilder(
+    val gateway: Gateway,
+    val selfId: Snowflake,
+    val channelId: Snowflake,
+    val guildId: Snowflake
```
HopeBaron, line 32: Make those vars to allow editing inside the builder

voice/src/main/kotlin/gateway/Event.kt
```diff
@@ -0,0 +1,133 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.buildClassSerialDescriptor
+import kotlinx.serialization.encoding.CompositeDecoder
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.json.JsonElement
+import mu.KotlinLogging
+
+private val jsonLogger = KotlinLogging.logger { }
+
+sealed class Event {
```
HopeBaron, line 23: rename this to something other than `Event`, something like `VoiceEvent` since we already have that in the core module and may cause confusion

voice/src/main/kotlin/gateway/Ticker.kt
```diff
@@ -0,0 +1,49 @@
...
+class Ticker(private val dispatcher: CoroutineDispatcher = Dispatchers.Default) : CoroutineScope {
+    override val coroutineContext: CoroutineContext = SupervisorJob() + dispatcher + CoroutineName("Ticker")
+
+    private val mutex = Mutex()
+
+    private var ticker: ReceiveChannel<Unit>? = null
+
+    suspend fun tickAt(intervalMillis: Long, block: suspend () -> Unit) {
+        stop()
+        mutex.withLock {
```
HopeBaron, line 30: why are we using a mutex here?

voice/src/main/kotlin/gateway/Ticker.kt
```diff
@@ -0,0 +1,49 @@
...
+                    } catch (exception: Exception) {
+                        logger.error(exception)
+                    }
+                }
+            }
+        }
+    }
+
+    suspend fun stop() {
+        mutex.withLock {
```
HopeBaron, line 45: same question as above

voice/src/main/kotlin/udp/AudioFramePoller.kt
```diff
@@ -0,0 +1,95 @@
...
+    val interceptorFactory: (FrameInterceptorContext) -> FrameInterceptor
+)
+
+internal class AudioFramePollerConfigurationBuilder {
+    var udp: DiscordUdpConnection? = null
+    var ssrc: Int by Delegates.notNull()
+    var key: List<Byte> by Delegates.notNull()
+    var provider: AudioProvider by Delegates.notNull()
+    var baseFrameInterceptorContext: FrameInterceptorContextBuilder by Delegates.notNull()
+    var interceptorFactory: (FrameInterceptorContext) -> FrameInterceptor by Delegates.notNull()
```
HopeBaron, line 37: Make those part of the constructor

voice/src/main/kotlin/handlers/VoiceUpdateEventHandler.kt
```diff
@@ -0,0 +1,34 @@
...
+private val voiceUpdateLogger = KotlinLogging.logger { }
+
+internal class VoiceUpdateEventHandler(
+    private val connection: VoiceConnection,
+    flow: Flow<GatewayEvent>,
+) : EventHandler<GatewayEvent>(flow, "VoiceUpdateInterceptor") {
+    override fun start() {
+        on<VoiceServerUpdate> { voiceServerUpdate ->
+            // see if this voice server update interests this voice connection
+            if (voiceServerUpdate.voiceServerUpdateData.guildId != connection.data.guildId) return@on
```
HopeBaron, line 21: move this into a variable name called isRelatedToConnection or some meaningful variable names rather than commenting what it does if possible.

voice/src/main/kotlin/VoiceConnection.kt
```diff
@@ -0,0 +1,146 @@
...
+ */
+@KordVoice
+class VoiceConnection(
+    val gateway: Gateway,
+    val voiceGateway: VoiceGateway,
+    val data: VoiceConnectionData,
+    internal var voiceGatewayConfiguration: VoiceGatewayConfiguration,
+    var audioProvider: AudioProvider,
+    var frameInterceptorFactory: (FrameInterceptorContext) -> FrameInterceptor,
+    voiceDispatcher: CoroutineDispatcher
```
HopeBaron, line 56: why are these vars?
and why the config is internal?

voice/src/main/kotlin/VoiceConnection.kt
```diff
@@ -0,0 +1,146 @@
...
+            { audioFramePoller.start(it.withConnection(this)) },
+            voiceDispatcher
+        )
+    }
+
+    /**
+     * Logs into the voice gateway, and begins the process of an audio-ready voice session.
+     */
+    fun connect() {
+        launch {
```
HopeBaron, line 80: I'm not sure if this should be launched, but I believe you should return the reference to the job

HopeBaron (changes requested):
connection wise, this seems fine.
This is my first review, I'll check it to night to see how it behaves
 with thanks for the efforts placed

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -49,6 +50,9 @@ object Dependencies {
     const val `ktor-client-websocket` = "io.ktor:ktor-client-websockets:${Versions.ktor}"
     const val `ktor-client-mock` = "io.ktor:ktor-client-mock:${Versions.ktor}"
     const val `ktor-client-mock-jvm` = "io.ktor:ktor-client-mock-jvm:${Versions.ktor}"
+    const val `ktor-network` = "io.ktor:ktor-network:${Versions.ktor}"
+
+    const val `codahale-xsalsa20poly1305` = "com.codahale:xsalsa20poly1305:${Versions.xsala20poly1305}"
```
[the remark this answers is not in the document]
lost-illusi0n: The xsalsa20poly1305 library? Yes, it is archived, but the spec for this type of encryption doesn't change, so it shouldn't be a problem. Though, changing it shouldn't be hard, as we just have to change the implementation of one method. If you have any better alternatives, I'll change it.

voice/src/main/kotlin/VoiceConnection.kt
lost-illusi0n: According to Discord docs, they are not nullable, so yes.

voice/src/main/kotlin/gateway/Ticker.kt
lost-illusi0n: I copied this from the gateway Ticker and didn't think changing one but not the other was the right move.

voice/src/main/kotlin/gateway/Ticker.kt
lost-illusi0n: I copied this from the gateway Ticker and didn't think changing one but not the other was the right move.

voice/src/main/kotlin/udp/AudioFramePoller.kt
lost-illusi0n: Hard to do... This builder helps facilitate an easy way to build out this configuration over the time. As we won't ever have all this information all at once. They are split up into 3 different places, two different event listeners build them in UdpLifeCycleHandler, and finally in VoiceConnection with the method AudioFramePollerConfigurationBuilder#withConnection.

voice/src/main/kotlin/VoiceConnection.kt
lost-illusi0n: Guess, I rushed this constructor, regarding the vars. Besides, config is changed in the VoiceUpdateHandler, and didn't think the end user should be changing this for *any* reason. Should I change the config to be public?

voice/src/main/kotlin/VoiceConnection.kt
lost-illusi0n: Kord Core eventually does the same thing in it's logIn method, but ye I can return a reference to the job.

lost-illusi0n:
Added a helper method for connecting to stage channels in 3bcbc56ecfab95d8838f5ed54292b845d2d0e0dd
Though, the one thing that is bugging me is that if the bot is a listener, the bot will still send audio packets. Now, to Discord this isn't a problem, however I'm not sure if this is what a user would expect. Should this case be handled be Kord or just leave it to the user?

voice/src/main/kotlin/VoiceConnection.kt
HopeBaron: mhm, it does make sense for the user to create their own voice connection if they have the data without the channel itself

HopeBaron:
Would it be better to use a channel?

voice/src/main/kotlin/udp/AudioFramePoller.kt
HopeBaron: so the builder is built over time?

lost-illusi0n: exactly

lost-illusi0n:
Could you clarify what you mean by that?

voice/src/main/kotlin/VoiceConnection.kt
lost-illusi0n: Well, if internal or not, the user can still create their `VoiceConnection` and provide their own initial voice gateway configuration, if they desire to do so. Though once it set, I don't see why a user would be changing it during the lifetime of a `VoiceConnection`, as that is done so properly by the `VoiceUpdateEventHandler`. Though I'll change it if you think it is best.

HopeBaron:
Using a channel to poll frames from the user they provide the frames through it.

voice/src/main/kotlin/VoiceConnection.kt
HopeBaron: example?

lost-illusi0n: Example of...? The initial gateway configuration is built in VoiceConnectionBuilder, Gateway#updateVoiceState. Any user could do this logic themselves and pass it into the VoiceConnection. Maybe we should make a public, read-only variable for the config, and an internal var that is changed by the VoiceUpdateEventHandler?? This seems reasonable to me...

lost-illusi0n:
I'll experiment with it.

lost-illusi0n:
Hey, 6ab76869fd8e7da96d6ebbb75d800889ce283159, is my first try at using a Channel for polling for audio frames.
I ran into several Kotlin bugs, you'll see comments briefly explaining them.
Now, for an average user, the api for AudioProvider will feel mostly the same, however it now provides a function to poll frames into a Channel. This is then used in the AudioFramePoller. This also allows a user more fine-grained control over the actually polling of frames, as they can override it if one would want to... What do you think?

HopeBaron:
try updating the kotlin version, it's fine to upgrade if it solves your problem. regarding inline classes, it's fine to make it non-inline

lost-illusi0n:
Did what you said, was able to keep the inline class as well.

voice/src/main/kotlin/VoiceConnection.kt
HopeBaron: No, there is no need to create a default config variable, the user will edit the parts they care about and  also this val object if it's immutable the user will need to create a new copy any how

PR #366, opened 2021-08-17
Contexts

HopeBaron:
This PR rewrites the interaction/application command api
* Provide concrete events instead of manual casting
* Provide a type safe option values for commands
* Provide resolvable message id.
* Implements User/Message functions.
* Rewrite the tree to better express the difference between Guild/Global Contexts

Note: In the new iteration any string type is resolved if it was a snowflake, which you would have access to in the `resolved` property of a `ChatInputInteraction`.


core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+import dev.kord.core.supplier.EntitySupplyStrategy
+
+/**
+ * Represents an interaction of type [ApplicationCommand][dev.kord.common.entity.InteractionType.ApplicationCommand]
+ */
+sealed interface ApplicationCommandInteraction : Interaction, ApplicationCommandInteractionBehavior
+
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through chat input.
```
BartArys, line 20:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through chat input.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+ */
+sealed interface  ChatInputCommandInteraction : ApplicationCommandInteraction {
+    val command: InteractionCommand get() =  InteractionCommand(data.data, kord)
+
+
+}
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through chat input specific to a guild.
```
BartArys, line 29:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through chat input specific to a guild.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+class GuildChatInputCommandInteraction(
+    override val data: InteractionData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : ChatInputCommandInteraction, GuildApplicationCommandInteraction
+
+
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through chat input.
```
BartArys, line 40:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through chat input.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+ */
+class GlobalChatInputCommandInteraction(
+    override val data: InteractionData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : ChatInputCommandInteraction, GlobalApplicationCommandInteraction
+
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through user commands.
```
BartArys, line 50:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through user commands.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+ */
+sealed interface  UserCommandInteraction : ApplicationCommandInteraction {
+    private val resolvedUsersData  get() = data.data.resolvedObjectsData.value?.users?.value
+    val users get() = resolvedUsersData.orEmpty().mapValues { User(it.value, kord) }
+}
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through user commands specific to a guild.
```
BartArys, line 58:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through user commands specific to a guild.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : UserCommandInteraction, GuildApplicationCommandInteraction {
+    private val resolvedMembersData  get() = data.data?.resolvedObjectsData.value?.members?.value
+    val members get() = resolvedMembersData.orEmpty().mapValues { memberData ->
+        Member(memberData.value, users[memberData.key]!!.data, kord) }
+}
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through user commands.
```
BartArys, line 71:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through user commands.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+ */
+class GlobalUserCommandInteraction(
+    override val data: InteractionData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : UserCommandInteraction, GlobalApplicationCommandInteraction
+
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through messages.
```
BartArys, line 81:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through messages.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+ */
+sealed interface  MessageCommandInteraction : ApplicationCommandInteraction {
+    private val resolvedMessagesData get() = data.data.resolvedObjectsData.value?.messages?.value
+    val messages get() = resolvedMessagesData.orEmpty().mapValues { Message(it.value, kord) }
+
+}
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through messages specific to a guild.
```
BartArys, line 90:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through messages specific to a guild.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+ */
+class GuildMessageCommandInteraction(
+    override val data: InteractionData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : MessageCommandInteraction, GuildApplicationCommandInteraction
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through messages.
```
BartArys, line 99:
```suggestion
 * An [ApplicationCommandInteraction] that's invoked through messages.
```

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+    override val data: InteractionData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : MessageCommandInteraction, ApplicationCommandInteraction {
+    override val user: UserBehavior
+        get() = UserBehavior(data.user.value!!.id, kord)
+
+    override fun withStrategy(strategy: EntitySupplyStrategy<*>): Interaction {
+        TODO("Not yet implemented")
+    }
```
BartArys, line 118: ?

BartArys (commented):
This looks good overall. It's a shame we can't seal the ApplicationCommandBehavior interfaces because of the package restrictions.

core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -0,0 +1,119 @@
...
+    override val data: InteractionData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier
+) : ChatInputCommandInteraction, GlobalApplicationCommandInteraction
+
+
+/**
+ * A [ApplicationCommandInteraction] that's invoked through user commands.
+ */
+sealed interface  UserCommandInteraction : ApplicationCommandInteraction {
```
thiakil, line 52: The only way to access the command id/name seems to be via `interaction.data.data.name.value` (same goes for Message commands)

Chat input commands have `interaction.command.rootName`

HopeBaron: that's not the thing you are looking for.
you can solve this by casting you have three types
RootCommand, SubCommand, GroupSubCommand.

just do ```kt
val command = interaction.command
if(command is SubCommand) ...
```

Comment (thiakil) [reply core/src/main/kotlin/entity/interaction/ContextInteraction.kt:52]: `UserCommandInteraction` _does not have_ a `command` field - my issue is not with slash commands, but with the other 2 types

Comment (HopeBaron) [reply core/src/main/kotlin/entity/interaction/ContextInteraction.kt:52]: You were right, I missed that one, now it's included in both interactions and it's derivaites

Comment (thiakil) [reply core/src/main/kotlin/entity/interaction/ContextInteraction.kt:52]: can we get it for the id field too? (`Interaction.data.data.id`)
I match mine with their handlers by snowflake

**Review by BartArys — COMMENTED** [review]
Interactions compare equality based on the backing data, but other entities compare equality on id instead.

**Review by BartArys — APPROVED** [review]

File: core/src/main/kotlin/entity/interaction/ContextInteraction.kt [hunk 12]
```diff
@@ -51,10 +65,7 @@ class GlobalChatInputCommandInteraction(
  */
 sealed interface  UserCommandInteraction : ApplicationCommandInteraction {
 
-    val name: String get() = data.data.name.value!!
-
-    private val resolvedUsersData  get() = data.data.resolvedObjectsData.value?.users?.value
-    val users get() = resolvedUsersData.orEmpty().mapValues { User(it.value, kord) }
+    val users get() = resolvedObjects!!.users
```
thiakil, line 88: This seems to have changed the nullability of `users` - before this change it was non-null, now it's nullable because that's how it is on `ResolvedObjects`

I'm not sure it can contain more than one value for a `UserCommandInteraction` anyway?

PR #367, opened 2021-08-17
Clean gradle files

DRSchlaubi:
This PR aims to improve the structure of the Gradle files by using Version Catalogs and extracting combined logic into the buildSrc folder

Current Issues
- The version bundles need to get added for each module manually
- How to properly addependencies (kotlin() function)?﻿


BartArys:
Thanks for keeping our gradle files up to date, could you elaborate on the issues you're having right now?

DRSchlaubi:
> Thanks for keeping our gradle files up to date, could you elaborate on the issues you're having right now?

Sure.

For one we have Gradle plugin Versions in the normal build gradle.kts files and in the buildsrc build.gradle.kts
These are required for the plugin scrips to work and can't access the usual versions object in buildSRac as buildSrc buildscript doesn't have access to buildSrc


Then I use the kotlin() function sometimes and don't really know how to specify that

Last but not least the common and test lib bundles need to be applied manually. The reason being that the libs extension gets registered after buildSrc scripts and after subprojects

HopeBaron:
Can you modify the voice module as well please?

DRSchlaubi:
> Can you modify the voice module as well please?

Uuuuhm in theory yes, but it will take a few days probably

HopeBaron:
Take all the time you need, just inform me when done so we can merge this

DRSchlaubi:
> Take all the time you need, just inform me when done so we can merge this

done

HopeBaron (approved):

PR #370, opened 2021-08-24
Fixes and improvements for Snowflake

Lukellmann:
## Fixes for `Snowflake`
- constructor that takes an `Instant` took the epoch milliseconds from the instant, shifted it to the left and *then* substracted the Discord Epoch but it should first subtract and then shift (see https://discord.com/developers/docs/reference#snowflake-ids-in-pagination-generating-a-snowflake-id-from-a-timestamp-example) -- fixed in https://github.com/kordlib/kord/commit/76a35b6024da0d033004e8f9eb46112b357c8265
- `SnowflakeMark.elapsedNow()` returned a `Duration` with the wrong sign (`Snowflake.min.timeMark.hasPassedNow()` returned `false` and `Snowflake.max.timeMark.hasPassedNow()` returned `true`) -- fixed in https://github.com/kordlib/kord/commit/b0bafb82b65817d7efee26a0e4f388b8724cab73
- `Snowflake.timeMark` passed unadjusted milliseconds since Discord Epoch to `SnowflakeMark`'s constructor which e.g. caused `Snowflake(Clock.System.now()).timeMark.elapsedNow()` to return a very long `Duration` where it should return a very short one -- fixed in https://github.com/kordlib/kord/commit/86d1412ccfcfba76bf95ab9cb467de475a835380 (and deduplicated logic from `timeStamp` and `timeMark` in https://github.com/kordlib/kord/commit/c47caa04f99e2bccc3a8054ec908d7651fe377a7)

## Improvements for `Snowflake`
- ~~https://github.com/kordlib/kord/commit/ad0d746239185aae6d9f55b4250cb2732233b05c: using `ushr` instead of `shr` to get correct behavior for `Snowflake`s with a negative `value` - Would it be a good idea to just use `ULong` for `value`? (It would bring a more accurate `Snowflake.max` with `ULong.MAX_VALUE`, avoid confusion with negative `Snowflake` values etc. But I'm not sure if this would break things elsewhere and if/how it would work with serialization.)~~ - replaced by https://github.com/kordlib/kord/commit/c889ff95b87709bfefb126ec42f2a02ba26b3d7f
- https://github.com/kordlib/kord/commit/9c6ba3dbaadbf89407afafa4bfdb33185f62c1fd: Additional documentation for `compareTo()` and `equals()` behavior for `Snowflake`.
- https://github.com/kordlib/kord/commit/3f9ad822c08bbe14fa5744d4fdd68b8f0ccc8ef5: Timestamp related tests, `Snowflake` constructed with `Instant` too far in the past/future has min/max timestamp instead of undefined overflow behavior, `Snowflake.endOfTime` constant and more documentation.
- https://github.com/kordlib/kord/commit/c889ff95b87709bfefb126ec42f2a02ba26b3d7f: Type of `value` is now `ULong` (was `Long` before) - this included modifying the serializer for `Snowflake`s.
- https://github.com/kordlib/kord/commit/361b87a49a31fcdf660f0498514a66329aa28e36: Factory function / pseudo constructor as a replacement for old `Long` constructor (with additional note on negative values).

### DIFF
--- a/common/src/main/kotlin/entity/Interactions.kt
+++ b/common/src/main/kotlin/entity/Interactions.kt
@@ -1,7 +1,6 @@
 package dev.kord.common.entity
 
 import dev.kord.common.annotation.KordExperimental
-import dev.kord.common.annotation.KordPreview
 import dev.kord.common.entity.optional.Optional
 import dev.kord.common.entity.optional.OptionalBoolean
 import dev.kord.common.entity.optional.OptionalSnowflake
@@ -625,7 +624,7 @@ fun CommandArgument<*>.boolean(): Boolean {
 
 
 fun CommandArgument<*>.snowflake(): Snowflake {
-    val id = string().toLongOrNull() ?: error("$value wasn't a Snowflake")
+    val id = string().toULongOrNull() ?: error("$value wasn't a Snowflake")
     return Snowflake(id)
 }
 
--- a/common/src/main/kotlin/entity/Snowflake.kt
+++ b/common/src/main/kotlin/entity/Snowflake.kt
@@ -117,3 +117,11 @@ private class SnowflakeMark(private val timeStamp: Instant) : TimeMark() {
 
     override fun elapsedNow(): Duration = Clock.System.now() - timeStamp
 }
+
+/**
+ * Creates a [Snowflake] from a given Long [value].
+ *
+ * Note: a negative [value] will be interpreted as an unsigned integer with the same binary representation, e.g.
+ * passing `-1L` for [value] will return a [Snowflake] with a [value][Snowflake.value] of [ULong.MAX_VALUE].
+ */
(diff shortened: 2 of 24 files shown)


BartArys:
Thanks for bringing to light this series of horrible mistakes and taking your time to fix them.

> Would it be a good idea to just use ULong for value? (It would bring a more accurate Snowflake.max with ULong.MAX_VALUE, avoid confusion with negative Snowflake values etc. 

Yes. Since 1.5.0 unsigned integers are stable and inline classes are serializable, so we do have that option now. Seeing as Snowflakes can never be negative, this seems like a good idea overall.

Also, if it's not too much to ask, could you write up a few tests for your fixes? It'd be a good idea to get these fixes solidified as regression tests.

Lukellmann:
Sure, I will try to write some tests later today.

PR #374, opened 2021-08-26
Application commands improvements

HopeBaron:


core/src/main/kotlin/entity/interaction/Interaction.kt
```diff
@@ -119,6 +119,16 @@ sealed interface InteractionCommand : KordObject {
...
 
+    val users: Map<String, User> get() = filterOptions()
+
+    val members: Map<String, Member> get() = filterOptions()
+
+    val channels: Map<String, ResolvedChannel> get() = filterOptions()
+
+    val roles: Map<String, Role> get() = filterOptions()
+
+    val mentionables: Map<String, Entity> get() = filterOptions()
```
BartArys, line 130: Don't these keys represent Snowflakes?

HopeBaron: those could be shorthanded from resolved. I think last time we talked about this we said that not all resolved values are in options

HopeBaron: it would work fine, each of those should be a part of options

BartArys (approved):

PR #379, opened 2021-08-29
add named files and attachments

### LINKED ISSUE #372: Missing `attachments` for message/webhook edit
The `attachments` field can be used to keep already existing attachments in a message, I guess that because Kord never uses that field, all attachments in the message are kept.

https://discord.com/developers/docs/resources/channel#edit-message
https://discord.com/developers/docs/resources/webhook#edit-webhook-message

HopeBaron:
fixes #372

### DIFF
--- a/rest/src/main/kotlin/builder/message/modify/EphemeralInteractionResponseModifyBuilder.kt
+++ b/rest/src/main/kotlin/builder/message/modify/EphemeralInteractionResponseModifyBuilder.kt
@@ -1,6 +1,5 @@
 package dev.kord.rest.builder.message.modify
 
-import dev.kord.common.annotation.KordPreview
 import dev.kord.common.entity.MessageFlag
 import dev.kord.common.entity.MessageFlags
 import dev.kord.common.entity.optional.Optional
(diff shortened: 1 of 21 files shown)


rest/src/main/kotlin/builder/message/modify/EphemeralInteractionResponseModifyBuilder.kt
```diff
@@ -36,10 +35,11 @@ class EphemeralInteractionResponseModifyBuilder
         return MultipartInteractionResponseModifyRequest(
             InteractionResponseModifyRequest(
                 content = state.content,
+                embeds = state.embeds.mapList { it.toRequest() },
                 allowedMentions = state.allowedMentions.map { it.build() },
+                flags = Optional(MessageFlags(MessageFlag.Ephemeral)),
                 components = state.components.mapList { it.build() },
-                embeds = state.embeds.mapList { it.toRequest() },
-                flags = Optional(MessageFlags(MessageFlag.Ephemeral))
+                attachments = state.attachments
```
BartArys, line 42: attachments aren't declared anywhere in the builder itself

BartArys (commented):
There's no easy way for the user to convert an `Attachment` to a `DiscordAttachment` in core, I don't see this as easy to use at all.

HopeBaron:
I haven't had a clear way to do it yet either wither we will apply the same utility functions that already exist for embeds

BartArys (approved):

PR #382, opened 2021-09-01
Limit Snowflakes to valid range

Lukellmann:
This PR implements the Snowflake limitations described in https://github.com/discord/discord-api-docs/pull/3745.

The valid range for a `Snowflake`'s raw value is represented by the new `validULongRange` property of `Snowflake`'s companion object. All constructors will [coerce their input in that range](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/coerce-in.html) so there is no way to create an invalid `Snowflake`.
If Discord ever decides to widen the valid range, it would just be a matter of adjusting the `validULongRange` accordingly to support that. (I intended this range to be the single point of truth).

https://github.com/kordlib/kord/commit/b96f3bbf081ceae1227349efc1a5e92cfe47c102 removes a test that was depending on an invalid `Snowflake` to pass. Since there is no longer a way to create such invalid `Snowflake`s, it was impossible to make this test pass.

This PR should be merged after https://github.com/kordlib/kord/pull/381 as it reverts (https://github.com/kordlib/kord/commit/46c746263fa3aeb9eafcab8ef37779a697ea3ece) the changes made there and replaces them with a more permanent solution.
I would also want to keep this PR open until https://github.com/discord/discord-api-docs/pull/3745 gets accepted.

### DIFF
--- a/common/src/main/kotlin/entity/Snowflake.kt
+++ b/common/src/main/kotlin/entity/Snowflake.kt
@@ -1,5 +1,6 @@
 package dev.kord.common.entity
 
+import dev.kord.common.entity.Snowflake.Companion.validULongRange
 import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
 import kotlinx.serialization.ExperimentalSerializationApi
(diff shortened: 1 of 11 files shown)


common/src/main/kotlin/entity/Snowflake.kt
```diff
@@ -72,8 +91,17 @@ class Snowflake(val value: ULong) : Comparable<Snowflake> {
...
-        private const val maxMillisecondsSinceDiscordEpoch = 0b111111111111111111111111111111111111111111L
+        private const val nonTimestampBitCount = 22
+        private const val discordEpochLong = 1420070400000L
+
+        /**
+         * The range that contains all valid raw Snowflake [value]s.
+         *
+         * Note that this range might change in the future.
+         */
+        val validULongRange: ULongRange = ULong.MIN_VALUE..Long.MAX_VALUE.toULong() // 0..9223372036854775807
```
Lukellmann, line 102: Should I use literals here?
like: `val validULongRange: ULongRange = 0uL..9223372036854775807uL`

PR #380, opened 2021-08-30
Change some types from String to Snowflake

Lukellmann:
- https://github.com/kordlib/kord/commit/d109712a47579a20ea2a8233f1bb1a77c592f28b: `DiscordGuildBan.guildId`
  see:
  https://discord.com/developers/docs/topics/gateway#guild-ban-add-guild-ban-add-event-fields
  https://discord.com/developers/docs/topics/gateway#guild-ban-remove-guild-ban-remove-event-fields
- https://github.com/kordlib/kord/commit/9784b5a095404a1ddd7d9618e384e39548513c7e: `DMCreatRequest.userId`
  see: https://discord.com/developers/docs/resources/user#create-dm
-  https://github.com/kordlib/kord/commit/6dfd86167b0c3e2fa30dee88c86b44d6f9b03f96: `ChannelFollowRequest.webhookChannelId`
  see: https://discord.com/developers/docs/resources/channel#follow-news-channel

PR #384, opened 2021-09-02
Restrict Snowflake range, type changes, consistent naming

Lukellmann:
This PR combines https://github.com/kordlib/kord/pull/380, https://github.com/kordlib/kord/pull/382 and some additional changes. I would want to keep this PR open until https://github.com/discord/discord-api-docs/pull/3745 gets accepted.

## Change some types from `String` to `Snowflake` (https://github.com/kordlib/kord/pull/380)

- https://github.com/kordlib/kord/commit/d109712a47579a20ea2a8233f1bb1a77c592f28b: `DiscordGuildBan.guildId`
  see:
  https://discord.com/developers/docs/topics/gateway#guild-ban-add-guild-ban-add-event-fields
  https://discord.com/developers/docs/topics/gateway#guild-ban-remove-guild-ban-remove-event-fields
- https://github.com/kordlib/kord/commit/9784b5a095404a1ddd7d9618e384e39548513c7e: `DMCreatRequest.userId`
  see: https://discord.com/developers/docs/resources/user#create-dm
-  https://github.com/kordlib/kord/commit/6dfd86167b0c3e2fa30dee88c86b44d6f9b03f96: `ChannelFollowRequest.webhookChannelId`
  see: https://discord.com/developers/docs/resources/channel#follow-news-channel

## Restrict `Snowflake`s to valid range (https://github.com/kordlib/kord/pull/382)

Implement the Snowflake limitations described in https://github.com/discord/discord-api-docs/pull/3745. This will e.g. fix an error in pagination (invalid Snowflakes were used before).

The valid range for a `Snowflake`'s raw value is represented by the new `validValues` property of `Snowflake`'s companion object. All constructors will [coerce their input in that range](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/coerce-in.html) so there is no way to create an invalid `Snowflake`.
If Discord ever decides to widen the valid range, it would just be a matter of adjusting `validValues` accordingly to support that. (I intended this range to be the single point of truth).

https://github.com/kordlib/kord/commit/b96f3bbf081ceae1227349efc1a5e92cfe47c102 removes a test that was depending on an invalid `Snowflake` to pass. Since there is no longer a way to create such invalid `Snowflake`s, it is impossible to make this test pass.

## Consistent naming

- https://github.com/kordlib/kord/commit/957fcc47a71abd18531a94be0cc21a48d3661bbb renames `Snowflake.discordEpochStart` to `Snowflake.discordEpoch` as it is the term used in Discord's documentation
- https://github.com/kordlib/kord/commit/fd4f387a6b8d7daddcff7cd765abf93def7f9d61 brings consistent naming to all `XxxTimestamp` classes and properties

### DIFF
--- a/common/src/main/kotlin/entity/Snowflake.kt
+++ b/common/src/main/kotlin/entity/Snowflake.kt
@@ -15,7 +15,7 @@ import kotlin.time.TimeMark
 
 /**
  * A unique identifier for entities [used by discord](https://discord.com/developers/docs/reference#snowflakes).
- * Snowflakes are IDs with a [timeStamp], which makes them [comparable][Comparable] based on their timestamp.
+ * Snowflakes are IDs with a [timestamp], which makes them [comparable][Comparable] based on their timestamp.
  *
  * Note: this class has a natural ordering that is inconsistent with [equals],
  * since [compareTo] only compares the first 42 bits of the ULong [value] (comparing the timestamp),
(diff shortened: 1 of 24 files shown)


common/src/main/kotlin/entity/Snowflake.kt
```diff
@@ -72,13 +98,33 @@ class Snowflake(val value: ULong) : Comparable<Snowflake> {
...
-        private const val maxMillisecondsSinceDiscordEpoch = 0b111111111111111111111111111111111111111111L
+        private const val nonTimestampBitCount = 22
+        private const val discordEpochLong = 1420070400000L
+
+        /**
+         * A range that contains all valid raw Snowflake [value]s.
+         *
+         * Note that this range might change in the future.
+         */
+        val validValues: ULongRange = ULong.MIN_VALUE..Long.MAX_VALUE.toULong() // 0..9223372036854775807
```
Lukellmann, line 109: Should I use literals here?
like: `val validValues: ULongRange = 0uL..9223372036854775807uL`

HopeBaron: it'd be better

BartArys: I disagree, magic numbers should be avoided as much as possible

HopeBaron: sounds like a fine argument to me :+1:

Lukellmann: And I have a comment for it anyway, so it should be fine

PR #385, opened 2021-09-03
Add discord messaging timestamps

warriorzz:
https://discord.com/developers/docs/reference#message-formatting

### DIFF
--- a/common/src/main/kotlin/DiscordTimestamp.kt
+++ b/common/src/main/kotlin/DiscordTimestamp.kt
@@ -1,45 +1,9 @@
 package dev.kord.common
 
 import kotlinx.datetime.Instant
-import kotlinx.datetime.LocalDateTime
-import kotlinx.datetime.TimeZone
-import kotlinx.datetime.toInstant
-import kotlinx.serialization.Serializable
 
-/**
- * The [timestamp format](https://discord.com/developers/docs/reference#message-formatting-formats) of the message formats
- */
-@Serializable
-data class DiscordTimestamp(
-
-    /**
-     * The Unix timestamp to be displayed
-     */
-    val time: Long,
-
-    /**
-     * The style of the timestamp
-     */
-    val style: DiscordTimestampStyle = DiscordTimestampStyle.ShortDateTime
-) : Comparable<DiscordTimestamp> {
-
-    val value: String
-        get() = "<t:$time:${style.style}>"
-
-    override fun equals(other: Any?): Boolean = other is DiscordTimestamp && other.time == time
-    operator fun plus(value: Long): DiscordTimestamp = DiscordTimestamp(time + value, style)
-    operator fun minus(value: Long): DiscordTimestamp = DiscordTimestamp(time - value, style)
-
-    override fun compareTo(other: DiscordTimestamp): Int = time.compareTo(other.time)
-
-    override fun toString() = value
-}
-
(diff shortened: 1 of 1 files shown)


common/src/main/kotlin/DiscordTimestamp.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    override fun toString() = value
+}
+
+fun LocalDateTime.toDiscordTimestamp(style: DiscordTimestampStyle? = null) =
+    DiscordTimestamp(toInstant(TimeZone.UTC).toEpochMilliseconds(), style ?: DiscordTimestampStyle.ShortDateTime)
+
+fun Instant.toDiscordTimestamp(style: DiscordTimestampStyle? = null) =
+    DiscordTimestamp(toEpochMilliseconds(), style ?: DiscordTimestampStyle.ShortDateTime)
```
Lukellmann, line 46: I would make `LocalDateTime.toDiscordTimestamp` call `Instant.toDiscordTimestamp` but this is just a nit-pick.
And use `DiscordTimestampStyle.ShorrDateTime` as the default for `style`, no need for this to be nullable.

Lukellmann (changes requested):

common/src/main/kotlin/DiscordTimestamp.kt
```diff
@@ -0,0 +1,87 @@
...
+import kotlinx.datetime.Instant
+import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toInstant
+import kotlinx.serialization.Serializable
+
+/**
+ * The [timestamp format](https://discord.com/developers/docs/reference#message-formatting-formats) of the message formats
+ */
+@Serializable
```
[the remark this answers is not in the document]
warriorzz: No real need, but I thought if anyone wants to insert this into their database, they don't need to create an extra data class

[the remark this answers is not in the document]
Lukellmann: Sure, did not think about that 😄

BartArys:
What do we gain from introducing a DiscordTimestamp? Its sole reason for existence seems to be to convert it to a String. Why not just make it a String directly instead?

The entire lib uses kx's `Instant` for time already, it seems more prudent to introduce a `toMessageFormat` (or whatever you want to call it) extension that accepts your `DiscordTimestampStyle` to those types directly instead, and save the hassle of using an adapter type.

warriorzz:
I didn't thought about that, but sounds good to me

common/src/main/kotlin/DiscordTimestamp.kt
```diff
@@ -3,7 +3,7 @@ package dev.kord.common
 import kotlinx.datetime.Instant
 
 fun Instant.toMessageFormat(style: DiscordTimestampStyle = DiscordTimestampStyle.ShortDateTime) =
-    "<t:${toEpochMilliseconds()}:${style.style}>"
+    "<t:${toEpochMilliseconds() / 1000}:${style.style}>"
```
NyCodeGHG, line 6: better use epochSeconds instead?

NyCodeGHG:
```suggestion
    "<t:$epochSeconds:${style.style}>"
```
now i can suggest thanks github

PR #386, opened 2021-09-06
Add support for receiving voice and opening up the voice api

lost-illusi0n:
All incoming packets are abstracted away into the `Streams` class where it provides a few different flows in different states of processing.

Attempted to open up the api.


voice/src/main/kotlin/Streams.kt
```diff
@@ -0,0 +1,69 @@
...
+@KordVoice
+class Streams(
+    connection: VoiceConnection,
+    dispatcher: CoroutineDispatcher
+) : CoroutineScope {
+    override val coroutineContext: CoroutineContext =
+        SupervisorJob() + dispatcher + CoroutineName("Voice Connection Incoming Streams")
+
+    // this will be set before it is used as the key is received before the udp connection is even established
+    internal lateinit var key: ByteArray
```
Lukellmann, line 21: This is not thread safe, there is no happens-before relationship between setting and using `key`.
-> `@Volatile` or `atomic`?

voice/src/main/kotlin/Streams.kt
```diff
@@ -0,0 +1,69 @@
...
+     */
+    val incomingUserStreams: SharedFlow<Pair<Snowflake, AudioFrame>> = _incomingUserAudioFrames
+
+    private val _ssrcToUser: MutableMap<UInt, Snowflake> = mutableMapOf()
+
+    /**
+     * A mapping of ssrc to user id.
+     * This cache is built over time through the [dev.kord.voice.gateway.VoiceGateway].
+     */
+    val ssrcToUser: Map<UInt, Snowflake> get() = _ssrcToUser
```
Lukellmann, line 52: Not thread safe -> `ConcurrentHashMap`?

voice/src/main/kotlin/Streams.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    init {
+        connection.voiceGateway.on<Speaking> {
+            if (ssrcToUser[ssrc] == null) {
+                _ssrcToUser[ssrc] = userId
```
Lukellmann, line 57: `computeIfAbsent` (especially with `ConcurrentHashMap`)

voice/src/main/kotlin/gateway/VoiceGateway.kt
```diff
@@ -83,6 +83,34 @@ interface VoiceGateway : CoroutineScope {
...
+ * Logger used to report throwables caught in [VoiceGateway.on].
+ */
+@PublishedApi
+internal val voiceGatewayOnLogger = KotlinLogging.logger("Gateway.on")
+
+/**
+ * Convenience method that will invoke the [consumer] on every event [T] created by [VoiceGateway.events].
+ *
+ * The events are buffered in an [unlimited][Channel.UNLIMITED] [buffer][Flow.buffer] and
+ * [launched][CoroutineScope.launch] in the supplied [scope], which is [Gateway] by default.
```
Lukellmann, line 97: should be `[VoiceGateway]` instead of `[Gateway]` in the documentation comment

voice/src/main/kotlin/handlers/UdpLifeCycleHandler.kt
```diff
@@ -4,45 +4,36 @@ package dev.kord.voice.handlers
...
 internal class UdpLifeCycleHandler(
     flow: Flow<VoiceEvent>,
-    private val send: suspend (Command) -> Unit,
-    private val poll: (AudioFramePollerConfigurationBuilder) -> Unit,
-    private val udpDispatcher: CoroutineDispatcher = Dispatchers.Default
+    private val connection: VoiceConnection
 ) : EventHandler<VoiceEvent>(flow, "UdpInterceptor") {
-    private var framePollerConfigurationBuilder = AudioFramePollerConfigurationBuilder()
-    private var udp: DiscordUdpConnection? = null
+    private var ssrc by Delegates.notNull<UInt>()
```
Lukellmann, line 27: This is not thread safe, there is no happens-before relationship between setting and using `ssrc`.
-> `@Volatile` or `atomic`?

voice/src/main/kotlin/udp/AudioFrameSender.kt
```diff
@@ -0,0 +1,85 @@
...
+) : AudioFrameSender {
+    override val coroutineContext: CoroutineContext =
+        SupervisorJob() + data.dispatcher + CoroutineName("Default Audio Frame Poller")
+
+    private fun createFrameInterceptor(configuration: AudioFrameSenderConfiguration): FrameInterceptor =
+        with(configuration) {
+            val builder = configuration.baseFrameInterceptorContext
+            builder.ssrc = configuration.ssrc
+            return interceptorFactory(builder.build()) // we should assume that everything else is set before-hand in the base builder
+        }
```
Lukellmann, line 52: `with` not used for `baseFrameInterceptorContext` and `ssrc`

voice/src/main/kotlin/udp/AudioFrameSender.kt
```diff
@@ -0,0 +1,85 @@
...
+                            encryptedData = frame.data
+                        ).encrypt(key)
+
+                        data.udp.send(encryptedPacket.asByteReadPacket())
+                        sequence++
+                    }
+                }
+        } catch (e: Exception) {
+            /* we're done polling, nothing to worry about */
+        }
```
Lukellmann, line 81: Where does the `Exception` come from?

voice/src/main/kotlin/udp/AudioPacket.kt
```diff
@@ -1,43 +1,78 @@
...
-        private const val RTP_VERSION = 0x78.toByte()
-        private const val NONCE_LENGTH = 24
-        private const val RTP_HEADER_LENGTH = 12
-    }
-
-    private var audioData = frame.data
-
     @OptIn(ExperimentalIoApi::class)
-    private val header = BytePacketBuilder().also {
+    val header = BytePacketBuilder().also {
```
Lukellmann, line 14: `apply` instead of `also`?

voice/src/main/kotlin/udp/AudioPacket.kt
```diff
@@ -1,43 +1,78 @@
...
+    ) : AudioPacket(sequence, timestamp, ssrc, encryptedData) {
+        fun decrypt(key: ByteArray): DecryptedPacket {
+            val nonce = ByteArray(NONCE_LENGTH)
+            header.copyInto(nonce, 0, 0, RTP_HEADER_LENGTH)
+            val decrypted = XSalsa20Poly1305Codec.decrypt(data, key, nonce) ?: error("couldn't decrypt audio data")
+            return DecryptedPacket(sequence, timestamp, ssrc, decrypted)
+        }
     }
 
     fun asByteReadPacket() = BytePacketBuilder().also {
```
Lukellmann, line 36: `apply` instead of `also`?

voice/src/main/kotlin/udp/AudioPacket.kt
```diff
@@ -1,43 +1,78 @@
...
         it.writeFully(header)
-        it.writeFully(audioData)
+        it.writeFully(data)
     }.build()
+
+    class DecryptedPacket(
+        sequence: UShort,
+        timestamp: UInt,
+        ssrc: UInt,
+        encryptedData: ByteArray
```
Lukellmann, line 45: should be `decryptedData` instead of `encryptedData`

voice/src/main/kotlin/udp/DefaultVoiceUdpConnection.kt
```diff
@@ -0,0 +1,66 @@
...
+)
+
+internal class DefaultVoiceUdpConnection(
+    val data: DefaultVoiceUdpConnectionData
+) : VoiceUdpConnection, CoroutineScope {
+    override val coroutineContext: CoroutineContext =
+        SupervisorJob() + data.dispatcher + CoroutineName("Discord Voice UDP Connection")
+
+    private lateinit var socket: ConnectedDatagramSocket
+    private lateinit var configuration: VoiceUdpConnectionConfiguration
```
Lukellmann, line 25: This is not thread safe, there is no happens-before relationship between setting and using `socket` and `configuration` (between `start`, `discoverIp` and `send`).
-> `@Volatile` or `atomic`?

Lukellmann (changes requested):
In addition to my comments, I was wondering whether you can disable receiving voice if you don't need it.

voice/src/main/kotlin/Streams.kt
lost-illusi0n: not something we can do easily if we want to go multiplatform. next point should be enough

voice/src/main/kotlin/udp/AudioFrameSender.kt
lost-illusi0n: anything that could cause the flow to terminate.
kord follows this same procedure in DefaultGatway and DefaultVoiceGateway

voice/src/main/kotlin/udp/AudioPacket.kt
lost-illusi0n: nice catch

voice/src/main/kotlin/Streams.kt
Lukellmann: Multiplatform could use some sort of expect and actual with `ConcurrentHashMap` on JVM

voice/src/main/kotlin/udp/DefaultVoiceUdpConnection.kt
lost-illusi0n: The lateinit socket is something done in Gateway and VoiceGateway, I think it would be best to rework the concurrency structure of each gateway/connection in its own pr, which *I think* is being done.

i don't mind making configuration atomic, however.

voice/src/main/kotlin/udp/AudioFrameSender.kt
Lukellmann: Alright, just me beeing curious

lost-illusi0n:
do you have an idea how disabling/enabling receiving voice would look like?

voice/src/main/kotlin/Streams.kt
lost-illusi0n: Still the same problem, we'd have to find different implementations for each platform. The actual mutablemap is private, and no other unsafe operations are being done.

Lukellmann:
> do you have an idea in mind how disabling/enabling receiving voice would look like? 

Maybe a `Boolean` setting in `VoiceConnectionBuilder`, disabled by default and some Noop `Streams` implementation?

voice/src/main/kotlin/Streams.kt
Lukellmann: Sure, the same would be true for other caches. Problem is that this map could be edited and accessed from any thread, so there is no guarantee for changes to become visible.

BartArys: The multiplatform solution is to make the variable atomic and interactions with the map sequential. You can achieve the latter by consuming the flow directly without launching.

voice/src/main/kotlin/Streams.kt
```diff
@@ -0,0 +1,76 @@
...
+     */
+    val incomingAudioFrames get() = incomingAudioPackets.map { it.ssrc to AudioFrame.fromData(it.data)!! }
+
+    private val _incomingUserAudioFrames: MutableSharedFlow<Pair<Snowflake, AudioFrame>> = MutableSharedFlow()
+
+    /**
+     * A flow of incoming [AudioFrame]s mapped to its corresponding user id. Streams for each user are built over time,
+     * or whenever the [dev.kord.voice.gateway.VoiceGateway] receives a [Speaking] event.
+     */
+    val incomingUserStreams: SharedFlow<Pair<Snowflake, AudioFrame>> = _incomingUserAudioFrames
```
Lukellmann, line 49: `incomingUserStreams` could also use `_incomingUserAudioFrames.asSharedFlow()` to forbid casting

voice/src/main/kotlin/gateway/VoiceGateway.kt
```diff
@@ -83,6 +83,34 @@ interface VoiceGateway : CoroutineScope {
...
+ */
+@KordVoice
+inline fun <reified T : VoiceEvent> VoiceGateway.on(
+    scope: CoroutineScope = this,
+    crossinline consumer: suspend T.() -> Unit
+): Job {
+    return this.events.buffer(Channel.UNLIMITED).filterIsInstance<T>().onEach {
+        launch { it.runCatching { it.consumer() }.onFailure(voiceGatewayOnLogger::error) }
+    }.launchIn(scope)
+}
```
Lukellmann, line 112: This should not launch inside `onEach` to make sure that changes to the `MutableMap` in `Streams` are sequential (see @BartArys comment). Other solution would be to not use `VoiceGateway.on()` in `Streams`.

voice/src/main/kotlin/udp/DefaultVoiceUdpConnection.kt
```diff
@@ -0,0 +1,71 @@
...
+    val data: DefaultVoiceUdpConnectionData
+) : VoiceUdpConnection, CoroutineScope {
+    override val coroutineContext: CoroutineContext =
+        SupervisorJob() + data.dispatcher + CoroutineName("Discord Voice UDP Connection")
+
+    private lateinit var socket: ConnectedDatagramSocket
+    private val configuration: AtomicRef<VoiceUdpConnectionConfiguration?> = atomic(null)
+
+    private val _incoming: MutableSharedFlow<ByteReadPacket> = MutableSharedFlow()
+    override val incoming: SharedFlow<ByteReadPacket> = _incoming
```
Lukellmann, line 33: `incoming` could also use `_incoming.asSharedFlow()` to forbid casting

Lukellmann (changes requested):

voice/src/main/kotlin/gateway/VoiceGateway.kt
lost-illusi0n: This function should stay the same, as it would be confusing that the VoiceGateway.on functions differently from Gateway.on
Though I now understand what you meant by not using launch, I did not realize it was used here :p

Lukellmann:
```kotlin
connection.voiceGateway.events.buffer(Channel.UNLIMITED).filterIsInstance<Speaking>().onEach { speaking ->
    // ...
}.launchIn(this)
```
in `init` for `Streams` should do the trick then

voice/src/main/kotlin/Streams.kt
lost-illusi0n: This pattern is used a few other places throughout Kord, is there any other benefit to do this?

Lukellmann: Nothing I know about, this is not a must (but I thought I could mention it).

voice/src/main/kotlin/handlers/StreamsHandler.kt
```diff
@@ -0,0 +1,20 @@
+package dev.kord.voice.handlers
+
+import dev.kord.common.annotation.KordVoice
+import dev.kord.voice.Streams
```
Lukellmann, line 4: wrong package, won't compile

voice/src/main/kotlin/streams/Streams.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * A representation of receiving voice through Discord and different stages of processing.
+ */
+@KordVoice
+abstract class Streams {
+    /**
+     * An encryption key used for decryption of Discord packets.
+     */
+    internal abstract var key: ByteArray?
```
Lukellmann, line 18: abstract class with internal abstract member cannot be inherited outside of voice module

voice/src/main/kotlin/handlers/StreamsHandler.kt
lost-illusi0n: intellij cannot refactor packages apparently, or properly at least :-(

voice/src/main/kotlin/streams/Streams.kt
lost-illusi0n: was a bit too easy i guess :p

Lukellmann (approved):

lost-illusi0n:
it seems like discord uses much more of the rtp spec than implemented before, and it differs from voice server to voice server. time to implement that

lost-illusi0n:
there may or may not be a memory leak that may or may not have to be fixed.

HopeBaron:
for what reason?

lost-illusi0n:
not completely sure yet but something is definitely there. here is a graph from jprofiler
image
this is not apparent when running a test bot without a voice connection.

lost-illusi0n:
Seems like there are actually two memory leaks, one major, and one much, much more minor.
I believe I fixed the major one, now only the minor one left. It seems to be in the default audio frame sender.

PR #387, opened 2021-09-11
Interaction improvements

HopeBaron:


core/src/main/kotlin/Kord.kt
```diff
@@ -371,6 +370,10 @@ class Kord(
         return defaultSupplier.getGuildApplicationCommands(resources.applicationId, guildId)
     }
 
+    suspend fun getOriginalInteractionResponse(token: String): Message? {
+        return EntitySupplyStrategy.rest.supply(this).getOriginalInteractionOrNull(resources.applicationId, token)
+    }
```
BartArys, line 375: What's the motivation of making this a 'top-level' Kord function? Seems like it belongs to an interaction behavior instead.

core/src/main/kotlin/entity/application/ApplicationCommand.kt
```diff
@@ -29,6 +29,10 @@ sealed interface ApplicationCommand : ApplicationCommandBehavior {
     val name: String
         get() = data.name
 
+    val version: Snowflake get() = data.version
+
+    val defaultPermission: Boolean? get() = data.defaultPermission.discordBoolean
```
BartArys, line 34: These properties should get documented

BartArys (approved):

PR #388, opened 2021-09-13
Add bill of materials

NyCodeGHG:
- Extract publishing into its own gradle script plugin
- Add bill of materials

This allows aligning kord dependency versions via [Maven BOMS](https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import)

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -1,0 +1,37 @@
+plugins {
+    `java-platform`
+    `kord-publishing`
+}
+
+javaPlatform {
+    allowDependencies()
+}
+
+val me = project
+rootProject.subprojects {
+    if (name != me.name) {
+        me.evaluationDependsOn(path)
+    }
+}
+
+dependencies {
+    constraints {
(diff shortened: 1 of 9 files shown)


bom/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+rootProject.subprojects {
+    if (name != me.name) {
+        me.evaluationDependsOn(path)
+    }
+}
+
+dependencies {
+    constraints {
+        rootProject.subprojects.forEach { subproject ->
+            if (subproject.plugins.hasPlugin("maven-publish") && subproject.name != name) {
```
HopeBaron, line 20: What is this block of code for

NyCodeGHG: It adds every kord module which applies the maven-publish plugin and is not the bom module as a contraint dependency.
it's also possible to add every module by hand

HopeBaron (approved):

BartArys (approved):

PR #389, opened 2021-09-16
Consistency fixes

ByteAlex:
This PR scopes to fix a few consistency fixes I encountered when I worked with Kord.


HopeBaron (approved):

core/src/main/kotlin/entity/channel/VoiceChannel.kt
```diff
@@ -35,6 +39,20 @@ class VoiceChannel(
...
+    val rtcRegion: String? get() = data.rtcRegion.value
+
+    /**
+     * Requests to get the [voice region][Region] of this guild.
+     *
+     * @throws [RequestException] if anything went wrong during the request.
+     * @throws [EntityNotFoundException] if the [Region] wasn't present.
+     * @throws [NoSuchElementException] if the [rtcRegion] is not in the available.
+     */
+    suspend fun getRegion(): Region = guild.regions.first { it.id == rtcRegion }
```
HopeBaron, line 54: There must be a orNull version as a safe approach for this one.

ByteAlex: Fixed

HopeBaron: shouldn't you update the docs as well.
This doesn't get the reagion of the guild, but the channel.
iirc Discord recently moved this property to be channel specific instead of guild

ByteAlex: That is correct. I adjusted the docs. I think word still pulls the region from the guild entity though. I'm not sure.

BartArys (approved):

PR #392, opened 2021-09-16
guildId abstraction on "Event"

ByteAlex:
Nothing complex, just a guildId field on the "Event" interface. Convenience.

### DIFF
--- a/core/src/main/kotlin/event/Event.kt
+++ b/core/src/main/kotlin/event/Event.kt
@@ -1,5 +1,6 @@
 package dev.kord.core.event
 
+import dev.kord.common.entity.Snowflake
 import dev.kord.core.Kord
 import dev.kord.gateway.Gateway
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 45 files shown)


BartArys:
It doesn't make sense for every event to have a guild id, adding a field that is 'sometimes present' depending on the implementation goes against the Interface segregation principle.

Additionally, there is nothing stopping someone else from suggesting the same style of changes for every property conceivable, you would end up with a horribly bloated Event interface where the vast majority of fields would end up returning null. This isn't a sustainable approach to the problem.

ByteAlex:
This one was solely a convenience PR, if it doesnt fit with you, you're good to close it.
The more I'm thinking about this, I may not need the guildId at all...? Probably been a bit too much in workaholic mode :^)

ByteAlex:
Cross that, I remember why I needed the guild id - Some of your cache actions (i.e. channels) are not key'd by guild if its a guild channel, but for me they're key'd by guild, so I need to resolve it's guild from the context.

PR #394, opened 2021-09-19
Add previous value to GuildUpdateEvent

### LINKED ISSUE #393: Add old guild object to GuildUpdateEvent
It'd be nice if `GuildUpdateEvent` had an `old` property, so that - assuming it was already in the cache - I'd be able to compare the old and new guild objects to figure out what changed.

BartArys:
Fixes #393 

Adds an `old` field with the previous (hopefully cached) guild to the `GuildUpdateEvent`.

### DIFF
--- a/core/src/main/kotlin/gateway/handler/GuildEventHandler.kt
+++ b/core/src/main/kotlin/gateway/handler/GuildEventHandler.kt
@@ -87,10 +87,11 @@ internal class GuildEventHandler(
 
     private suspend fun handle(event: GuildUpdate, shard: Int) {
         val data = GuildData.from(event.guild)
+        val old = cache.query<GuildData> { idEq(GuildData::id, event.guild.id) }.singleOrNull()
         cache.put(data)
         event.guild.cache()
 
-        coreFlow.emit(GuildUpdateEvent(Guild(data, kord), shard))
+        coreFlow.emit(GuildUpdateEvent(Guild(data, kord), old?.let { Guild(it, kord) }, shard))
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/gateway/handler/GuildEventHandler.kt
```diff
@@ -87,10 +87,11 @@ internal class GuildEventHandler(
 
     private suspend fun handle(event: GuildUpdate, shard: Int) {
         val data = GuildData.from(event.guild)
+        val old = cache.query<GuildData> { idEq(GuildData::id, event.guild.id) }.singleOrNull()
```
ByteAlex, line 90: Instead of a cache query eventually use the EntityProvider?
If not, thinking about my use case with distributed cache / own gateway impl, how would I populate this field?

HopeBaron: The goal is to supply what we know before the event happened. Im not sure I understand

ByteAlex: Thinking on my distributed setup, where Kord does not have any cache - How can I populate this field from an outside scope?

BartArys: Cache and EntitySupplier do not provide the same functionality. You're free to implement your own `DataEntryCache` and hook them up in the `KordBuilder`, though we are moving towards a new API that will make it easier to provide custom implementations and optimize queries.

PR #396, opened 2021-09-23
Behaviour improvement

ByteAlex:
Kord entities inherit from their Behaviours, so to keep generic in some places you use a behaviour.

When calling as{Entity} on a Behaviour, I'd expect it to return itself when it's already an entity, but it's doing a cache-lookup instead.

This PR fixes that and returns the Entity in case the Behaviour is an Entity already.


HopeBaron (changes requested):
instead of including this in the interface just override it in the core entity to reflect `this` just like we do in sub channel types
https://github.com/kordlib/kord/blob/0.8.x/core/src/main/kotlin/entity/channel/Category.kt#L29

ByteAlex:
I force pushed a fix for that, this should resolve that!

BartArys (approved):

PR #397, opened 2021-09-24
Ephemeral files

HopeBaron:


BartArys (approved):

common/src/main/kotlin/ImageFormat.kt
```diff
@@ -0,0 +1,16 @@
...
+    WebP("webp"),
+    PNG("png"),
+    Lottie("json");
+
+    companion object {
+        fun isSupported(fileName: String): Boolean {
+            return ImageFormat.values().any { fileName.endsWith(it.extension, true) }
+        }
+    }
+}
```
BartArys, line 16: We already have something like `ImageFormat`, `Image.Format`. Probably better to use that.

PR #398, opened 2021-09-27
Add missing `invitable` field

MrPowerGamerBR:
The `invitable` field controls this option

image


HopeBaron:
shouldn't this be added to the create data as well

MrPowerGamerBR:
> shouldn't this be added to the create data as well

Yes, I forgot about that, let me add it there too

MrPowerGamerBR:
Done 😊

HopeBaron:
I guess you missed the thread create builder as well

MrPowerGamerBR:
> I guess you missed the thread create builder as well

I think that a ThreadCreateBuilder doesn't exist (yet) :(

image

HopeBaron:
it's called the StartThreadRequest but since properties are being added to it maybe make a builder if you will.

MrPowerGamerBR:
I added the `StartThreadBuilder`, however I'm not sure if the changes that I made to the core module are correct

core/src/main/kotlin/behavior/channel/TextChannelBehavior.kt
```diff
@@ -69,18 +72,24 @@ interface TextChannelBehavior : PrivateThreadParentChannelBehavior {
...
+        reason: String? = null,
     ): TextChannelThread {
-        return unsafeStartThread(name, archiveDuration, ChannelType.PublicGuildThread, reason) as TextChannelThread
+        return unsafeStartThread(
+            name,
+            archiveDuration,
+            ChannelType.PublicGuildThread
+        ) {
+            this.reason = reason
+        } as TextChannelThread
```
HopeBaron, line 83: remove the reason from here, add the builder as a part of the function
so: 
```kt 
suspend fun startPublicThread(
        name: String,
        archiveDuration: ArchiveDuration = ArchiveDuration.Day,
        builder: StartThread.() -> Unit
    ): TextChannelThread
```
and then pass it to unsafe

rest/src/main/kotlin/builder/channel/thread/StartThreadBuilder.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import dev.kord.common.entity.ArchiveDuration
+import dev.kord.common.entity.ChannelType
+import dev.kord.common.entity.optional.OptionalBoolean
+import dev.kord.common.entity.optional.delegate.delegate
+import dev.kord.common.entity.optional.optional
+import dev.kord.rest.builder.AuditRequestBuilder
+import dev.kord.rest.json.request.StartThreadRequest
+
+class StartThreadBuilder(var name: String, var autoArchiveDuration: ArchiveDuration, var type: ChannelType) : AuditRequestBuilder<StartThreadRequest> {
```
HopeBaron, line 11: make type a `val` here, you don't want to make a switch in the types which makes the builder no longer useful.

HopeBaron (changes requested):

rest/src/main/kotlin/builder/channel/thread/StartThreadBuilder.kt
MrPowerGamerBR: Only `val` for the `ChannelType`, or for `name` and `autoArchiveDuration` too?

HopeBaron: `type` only

PR #401, opened 2021-09-30
Member Avatars & Avatar->Icon refactor

ByteAlex:


core/src/main/kotlin/entity/Icon.kt
```diff
@@ -0,0 +1,75 @@
...
+        getUrl(format, size)?.let { Image.fromUrl(kord.resources.httpClient, it) }
+
+    override fun toString(): String {
+        return "Icon(type=${javaClass.name},animated=$animated,rawAssetUri=$rawAssetUri,kord=$kord)"
+    }
+
+    class EmojiIcon(kord: Kord, data: EmojiData) : Icon(
+        kord,
+        data.animated.discordBoolean,
+        "$CDN_BASE_URL/emojis/${data.id.asString}"
```
HopeBaron, line 52: make the rawAssertUri an abstract property and not a part of the constructor so you'd be able to define it inside the entity itself. Maybe create a `Route` like implementation for these.

core/src/main/kotlin/entity/Icon.kt
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    class DefaultUserAvatar(kord: Kord, data: UserData) :
+        Icon(kord, false, "$CDN_BASE_URL/embed/avatars/${data.discriminator.toInt() % 5}")
+
+    class UserAvatar(kord: Kord, data: UserData) :
+        Icon(
+            kord,
+            data.avatar!!.startsWith("a_"),
+            "$CDN_BASE_URL/avatars/${data.id.asString}/${data.avatar}"
```
HopeBaron, line 62: Same as above

core/src/main/kotlin/entity/Icon.kt
```diff
@@ -0,0 +1,75 @@
...
+        )
+
+    class MemberAvatar(kord: Kord, data: MemberData) :
```
HopeBaron, line 65: Same as above.

core/src/main/kotlin/entity/User.kt
```diff
@@ -24,7 +23,14 @@ open class User(
     override val id: Snowflake
         get() = data.id
 
-    val avatar: Avatar get() = Avatar(data, kord)
+    /**
+     * The users avatar as [Icon] object
+     */
+    val avatar: Icon
+        get() = when {
+            data.avatar != null -> Icon.UserAvatar(kord, data)
```
HopeBaron, line 31: split those into a defaultAvatar and an avatar one that is nullable instead of squashing both into a this or that fallback logic. and the user then would be able to pick if they want to fall back to the default avatar.

With this. the `DefaultAvatar` sealed instance would become redundant.

HopeBaron (changes requested):
The overall design is pretty good. Im against making retrieving something related to the user from the avatar, when we can get the default avatar from the user itself with a property or a function as a `defaultAvatar`. `Icon`s shall represent an image and shall not depend on external data of another entity.

core/src/main/kotlin/entity/User.kt
ByteAlex: I would still like to maintain the DefaultAvatar instance in the Icon class, as this is kind of the "Route" enum. Making the avatar nullable and having the user fall back to the "defaultAvatar" is fine for me though

core/src/main/kotlin/entity/Icon.kt
ByteAlex: The inner classes of the Icon class, are kind of a "Route" enum. I really like this approach and I think you should reconsider this.
The rawAssetUri is a private val on purpose, since it does not contain a format yet and is used in the `val url` field which appends the default format.

Would you want to reconsider this, or is this decision final?
Please let me know, so I can adjust if needed.

HopeBaron: None of the feedback I gave is final.
I'm not speaking about making them inner classes. I don't see a problem with that. tho I don't like that you are explicitly passing the string there.
Maybe if you could make it so someone else can do something like `CDN.Emojis` would be of great help for everyone :D

rest/src/main/kotlin/route/DiscordCDN.kt
```diff
@@ -0,0 +1,35 @@
...
+}
+
+class CDNUrl(private val rawAssetUri: String) {
+
+    fun toUrl(builder: UrlBuilder.() -> Unit = {}): String {
+        val config = UrlBuilder().apply(builder)
+        var cdnUrl = "$rawAssetUri.${config.format.extension}"
+        config.size?.let { cdnUrl += "?size=${it.maxRes}" }
+        return cdnUrl
+    }
```
BartArys, line 28: making this lambda non-inline is problematic, turning every call into its own class, doubly so since this lib heavily depends on suspend calls which can't happen in there.

You're better off splitting the internals into a separate public function, creating a `toUrl(UrlBuilder)` and inlining this one.

rest/src/main/kotlin/route/DiscordCDN.kt
```diff
@@ -0,0 +1,35 @@
...
+
+}
+
+class CDNUrl(private val rawAssetUri: String) {
+
+    fun toUrl(builder: UrlBuilder.() -> Unit = {}): String {
+        val config = UrlBuilder().apply(builder)
+        var cdnUrl = "$rawAssetUri.${config.format.extension}"
+        config.size?.let { cdnUrl += "?size=${it.maxRes}" }
+        return cdnUrl
```
BartArys, line 27: String concat is relatively expensive on the JVM, consider using a `StringBuilder` (possibly the `buildString` DSL) instead.

rest/src/main/kotlin/route/DiscordCDN.kt
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    override fun toString(): String {
+        return "CdnUrl(rawAssetUri=$rawAssetUri)"
+    }
+
+    data class UrlBuilder(var format: Image.Format = Image.Format.WEBP, var size: Image.Size? = null)
+}
```
BartArys, line 35: This class should get its own file

BartArys (approved):

PR #402, opened 2021-09-30
Implement Role Icons

ByteAlex:
Depends on #401 - Uses DiscordCDN


BartArys:
Did you test whether Kord's HttpClient works when retrieving these images? I remember there being an issue with Discord's CDN failing any request that had unknown header values, and Kord by default puts bot auth in every request.

ByteAlex:
This was working in the old code, so I assumed it would.
https://github.com/kordlib/kord/pull/401/files#diff-2fc55b4c2e7355d05121c524cc9f02992854c6edaba29d4979f19ad41d44c9a0L167

BartArys:
> 
> 
> This was working in the old code, so I assumed it would. 

Fair point hah. I believe we just pushed it assuming it was a bug on Discord's end that would eventually get fixed. It's not really relevant to the PR then I suppose.

ByteAlex:
I resolved the merge conflict, this would be ready for review now too

HopeBaron (approved):

PR #406, opened 2021-10-01
Enable explicit api mode for core module

NyCodeGHG:
Enables explicit api mode for the core module as discussed on the discord server.
CI will fail because @ByteAlex broke the test :)


ByteAlex:
And what exactly did I break? 🤔

NyCodeGHG:
i don't actually now, i rebased onto the latest commit and now that test is failing

HopeBaron:
I fixed it, you used an `asChannel` instead of `fetchChannel` in one test.

Issue #395, opened 2021-09-22
Intents can't be specified lazily


gdude2002:
I'm running into an issue with [Kord Extensions](https://github.com/Kord-Extensions/kord-extensions), which recently gained a new feature to automatically enable the intents required for commands and event handlers to work. This has resulted in a load order issue. 

In order for intents to be registered properly, all extensions must have been set up before the `Kord` instance is constructed via the `KordBuilder`. This is because the `KordBuilder` has a single `intents` property rather than an intents _builder_ that's evaluated as part of login. However, this means that it's now unsafe to try to access the bot's `Kord` instance during extension setup, as it doesn't yet exist.

So, this issue is a formal request to change the `intents` property in `KordBuilder` to a DSL function that registers an intents builder which is called just before login in order to resolve the required intents.

I've worked around this in KordEx for now, but it does make some things considerably less intuitive, so it'd be great to get this looked at.

BartArys:
If the proposed builder is only called when logging in then we should just move the intents to the log in function instead. We've had requests from other users to modify intents between gateway identifies, which would also be solved by this change. 

The notable downside is that intents won't be usable by other parts of the API (I can only think of cache being potentially affected), and if we are to move it to the log in call it'll become the user's responsibility to remember intents between manual log in attempts (not sure how many times people actually manually reconnect, but I don't think it'd be much of an issue).

gdude2002:
I feel like "builder supplied early and called at login" negates the last downside there, but you're probably right that it isn't much of an issue - bots at that scale are going to be frameworking that anyway, I think.

PR #409, opened 2021-10-03
Lazy Intents

HopeBaron:
Addresses #395 moves intents and allows modifying a few properties.

### DIFF
--- a/gateway/src/main/kotlin/builder/LoginBuilder.kt
+++ b/gateway/src/main/kotlin/builder/LoginBuilder.kt
@@ -1,0 +1,21 @@
+package dev.kord.gateway.builder
+
+import dev.kord.common.entity.PresenceStatus
+import dev.kord.gateway.Intents
+import kotlin.contracts.ExperimentalContracts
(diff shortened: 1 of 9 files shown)


gateway/src/main/kotlin/builder/LoginBuilder.kt
```diff
@@ -0,0 +1,21 @@
...
+import dev.kord.common.entity.PresenceStatus
+import dev.kord.gateway.Intents
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+
+data class Shards(val totalShards: Int, val indices: Iterable<Int> = 0 until totalShards)
+
+class LoginBuilder {
+    var presence: PresenceBuilder.() -> Unit = { status = PresenceStatus.Online }
```
BartArys, line 12: What's the reason for this being a lambda? Why not keep a presence builder locally instead?

BartArys (approved):

PR #420, opened 2021-10-13
Unify permission overwrite API (Fix #302)

### LINKED ISSUE #302: add addRoleOverwrite() to Channels Builder too
the addRoleOverwrite() method provides a much convenient way to modify permissions, If it could also be in Channel Builders, that's gonna be great!

DRSchlaubi:
This unifys the api of all builders allowing creation oor modification of permission overrites﻿

Fixes #302


rest/src/main/kotlin/builder/channel/PermissionOverritesBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+import dev.kord.common.entity.OverwriteType
+import dev.kord.common.entity.Snowflake
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+
+/**
+ * Interface containing shared behavior on all builders modifying permission overwrites.
+ */
+interface PermissionOverritesBuilder {
```
BartArys, line 13:
```suggestion
interface PermissionOverwritesBuilder {
```

rest/src/main/kotlin/builder/channel/PermissionOverritesBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import dev.kord.common.entity.Overwrite
+import dev.kord.common.entity.OverwriteType
+import dev.kord.common.entity.Snowflake
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+
+/**
+ * Interface containing shared behavior on all builders modifying permission overwrites.
```
BartArys, line 11:
```suggestion
 * Builder that can modify [permission overwrites](Overwrite).
```

rest/src/main/kotlin/builder/channel/VoiceChannelCreateBuilder.kt
```diff
@@ -27,15 +28,15 @@ class VoiceChannelCreateBuilder(var name: String) : AuditRequestBuilder<GuildCha
     private var _position: OptionalInt = OptionalInt.Missing
     var position: Int? by ::_position.delegate()
 
-    val permissionOverwrites: MutableList<Overwrite> = mutableListOf()
+    override var permissionOverwrites: MutableSet<Overwrite>? = mutableSetOf()
```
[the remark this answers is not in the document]
BartArys: > Because it needs to be nullable in the modify builders

If something doesn't apply to a class, then it shouldn't have to implement it.

Replace the property with an `addOverwrite(Overwrite)` function and delegate to that instead. None of the DSLs require the removal or modification of overwrites anyhow.

rest/src/main/kotlin/builder/channel/PermissionOverritesBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+interface PermissionOverritesBuilder {
+    /**
+     *  The permission overwrites for this category.
+     */
+    var permissionOverwrites: MutableSet<Overwrite>?
+}
+
+/**
+ * adds a [Overwrite] for the [memberId].
```
BartArys, line 21:
```suggestion
 * Adds an [Overwrite] for the [memberId].
```

rest/src/main/kotlin/builder/channel/PermissionOverritesBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+    contract {
+        callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
+    }
+    val overwrite = permissionOverwrites ?: mutableSetOf()
+    overwrite.add(PermissionOverwriteBuilder(OverwriteType.Member, memberId).apply(builder).toOverwrite())
+    permissionOverwrites = overwrite
+}
+
+/**
+ * adds a [Overwrite] for the [roleId].
```
BartArys, line 34:
```suggestion
 * Adds an [Overwrite] for the [roleId].
```

rest/src/main/kotlin/builder/channel/PermissionOverwritesBuilder.kt
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Builder that can modify [permission overwrites][Overwrite].
+ */
+interface PermissionOverwritesBuilder {
+    /**
+     * Adds [overwrite] to this builder.
+     */
+    fun addOverwrite(overwrite: Overwrite)
+}
```
BartArys, line 18: Not that we should expect people to implement this, or accept it as a parameter, but this should probably be a sealed interface to prevent future issues where we somehow need to figure out the type of builder ourselves.

```suggestion
sealed interface PermissionOverwritesBuilder {
    /**
     * Adds [overwrite] to this builder.
     */
    fun addOverwrite(overwrite: Overwrite)
}
```

DRSchlaubi: Shouldn't Modify and Create be sealed too then?

BartArys: I can almost guarantee that's going to cause issue with our rest<->core modules if we do

PR #423, opened 2021-10-15
Voice Concurrency

psmarko:
This PR fixes the coroutine-related downfalls in the current voice implementation. There should no longer be any lingering coroutines after calling `shutdown` on a `VoiceConnection`. This PR also tries to adhere to a more defined coroutine structure, along with polishing some reconnection/disconnection logic.

Based off of #386


lost-illusi0n:
ah right, it is based off of #386 but when it was merged, the history was all squashed into c1e1bf5f4c78dd6e5f5ae0ddb7e6bdd2b5573e3f

Issue #436, opened 2021-11-03
Add old property to Update events


Galarzaa90:
`MemberUpdateEvent`, `GuildUpdateEvent`, `UserUpdateEvent` all have an `old` property that allows you to compare it to the current entity in order to know what changed.

However, other events like `RoleUpdateEvent` and `ChannelUpdateEvent` don't have it.

HopeBaron:
Should find them on latest snapshot

PR #437, opened 2021-11-04
Added old channel on ChannelUpdateEvent

onlyfx:
Refers to the https://github.com/kordlib/kord/issues/436, I don't know if there should be an "old" property in RoleUpdateEvent, the only way not to be null (from what I've seen) is by using cachingRest


HopeBaron:
surpassed by #438 will keep open until merged

onlyfx:
alright

PR #440, opened 2021-11-09
Fix bom

NyCodeGHG:
When I submitted the bom in #388 i used the wrong function and the bom is broken and includes all dependencies instead of just setting the version.

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -3,10 +3,6 @@ plugins {
     `kord-publishing`
 }
 
-javaPlatform {
-    allowDependencies()
-}
-
 val me = project
 rootProject.subprojects {
     if (name != me.name) {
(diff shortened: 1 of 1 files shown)


HopeBaron:
What's the difference between the two function calls, shouldn't these be the same?

NyCodeGHG:
That's what I thought when I submitted the first PR, but there is no api(String, String, String) in the constraints scope, that's the regular api function from the dependencies scope.
In the pom the dependencies are regular dependencies: https://repo1.maven.org/maven2/dev/kord/kord-bom/0.8.0-M7/kord-bom-0.8.0-M7.pom
But it should look like this: https://s01.oss.sonatype.org/content/repositories/snapshots/dev/nycode/github/bom/1.0.0-SNAPSHOT/bom-1.0.0-20211109.142957-14.pom

HopeBaron:
what about the removal of `allowDependencies`

NyCodeGHG:
Thats Gradle's way to prevent you from doing the mistake I did. But I did the mistake anyway. It's only needed when you want to include external dependencies in the bom when i understood correctly.

PR #441, opened 2021-11-10
Make Snowflake -> String conversion more idiomatic

BenWoodworth:
This PR deprecates `Snowflake.asString`, and gives its functionality to `Snowflake.toString()` instead.
(And migrates all usages of `.asString` in the repo to `.toString()`)

With this change, `"$snowflake"` can be done instead of `"${snowflake.asString}"`.

The main changes are in Snowflake.kt, with all the other changes being the migrations.

<hr>

Thinking this is more idiomatic is mostly based on intuition, but I have some rationale:
- Can't think of a use case needing the string `"Snowflake(value=...)"` outside of println debugging
  - `Snowflake.toString()` is never used in this repo
  - `"...${snowflake.asString}..." is used 26 times`, when `"...$snowflake..."` would be more convenient
- `asString` being a property feels out of place, since semantically it's a conversion, not an attribute of Discord snowflakes
  - Although I can think of kotlinx libraries that use properties for convenience conversions (e.g. [JSON](https://github.com/Kotlin/kotlinx.serialization/blob/f6a34325e557c51ed472f9f4c98497ccc035e39e/formats/json/commonMain/src/kotlinx/serialization/json/JsonElement.kt#L148-L229))
- I expect symmetry when a value type has a String constructor: `snowflake == Snowflake(snowflake.toString())`
- I expect symmetry between `.to*()` and `.as*()` functions: `snowflake.asString == snowflake.toString()`

### DIFF
--- a/common/src/main/kotlin/entity/DiscordMessage.kt
+++ b/common/src/main/kotlin/entity/DiscordMessage.kt
@@ -332,7 +332,7 @@ class MessageFlags internal constructor(val code: Int) {
         return builder.flags()
     }
 
-    override fun toString(): String = "MessageFlags(flags=${flags.toString()})"
+    override fun toString(): String = "MessageFlags(flags=$flags)"
 
     internal object Serializer : KSerializer<MessageFlags> {
 
--- a/common/src/main/kotlin/entity/Presence.kt
+++ b/common/src/main/kotlin/entity/Presence.kt
@@ -52,7 +52,7 @@ data class DiscordPresenceUser(
         override fun serialize(encoder: Encoder, value: DiscordPresenceUser) {
             val jsonEncoder = encoder as? JsonEncoder ?: error("Can be serialized only by JSON")
             val details = value.details.toMutableMap()
-            details["id"] = JsonPrimitive(value.id.asString)
+            details["id"] = JsonPrimitive(value.id.toString())
 
             jsonEncoder.encodeJsonElement(JsonObject(details))
         }
--- a/common/src/main/kotlin/entity/Snowflake.kt
+++ b/common/src/main/kotlin/entity/Snowflake.kt
@@ -68,6 +68,7 @@ class Snowflake : Comparable<Snowflake> {
     /**
      * A [String] representation of this Snowflake's [value].
      */
+    @Deprecated("Use toString() instead", ReplaceWith("toString()"))
     val asString get() = value.toString()
 
     /**
@@ -91,7 +92,10 @@ class Snowflake : Comparable<Snowflake> {
     override fun compareTo(other: Snowflake): Int =
         millisecondsSinceDiscordEpoch.compareTo(other.millisecondsSinceDiscordEpoch)
 
-    override fun toString(): String = "Snowflake(value=$value)"
+    /**
+     * A [String] representation of this Snowflake's [value].
+     */
+    override fun toString(): String = value.toString()
 
     override fun hashCode(): Int = value.hashCode()
 
--- a/common/src/test/kotlin/json/ChannelTest.kt
+++ b/common/src/test/kotlin/json/ChannelTest.kt
@@ -19,14 +19,14 @@ class ChannelTest {
         val channel = Json.decodeFromString(DiscordChannel.serializer(), file("dmchannel"))
 
         with(channel) {
-            lastMessageId.value!!.asString shouldBe "3343820033257021450"
+            lastMessageId.value!!.toString() shouldBe "3343820033257021450"
             type.value shouldBe 1
-            id.asString shouldBe "319674150115610528"
+            id.toString() shouldBe "319674150115610528"
             recipients.value!!.size shouldBe 1
             with(recipients.value!!.first()) {
                 username shouldBe "test"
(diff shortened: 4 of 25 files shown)


HopeBaron:
I've to mention that before we proceed that `asString` doesn't have the same functionality as `toString` as your PR gives the impression you thing they do.
`asString` gives the id value in the form of a String, such as: "908077612016992316"
where as `toString` shows `"Snowflake(908077612016992316)"`
that being cleared up, what makes this PR valid.
you can't compare json retrieved values with `toString` since Snowflake(908077612016992316) != 908077612016992316 (assuming strings)

BenWoodworth:
The PR changes the functionality `toString()` to be the same as `asString`: [diff](https://github.com/kordlib/kord/pull/441/files#diff-14c16318d3f43f29642b42acbb7fbab4585c49b89eb2101803e8f4472827276fL94-R98)

So that means you can compare json retrieved values with `toString`, since
`Snowflake(908077612016992316).toString() == "908077612016992316"`

Is that what you mean?

HopeBaron:
But this is not a string representation of a snowflake

BenWoodworth:
I think I get what you're saying. Is there a reason the citation's needed in the String representation?

It's normal in idiomatic Kotlin for `toString()` to be indistinguishable from other types (e.g. `1`, `1uL`, `"1"`)
This also is the representation used by Discord's REST APIs (since snowflakes *are* integers, just with special semantics)

HopeBaron:
But they are numbers still, just limited, Snowflakes are not normal numbers in the system, they hold things like timestamps, date of creation etc.

BenWoodworth:
Yep! That's the semantics I was talking about

HopeBaron:
I think I was also unclear in the previous comment comment.
You said
> It's normal in idiomatic Kotlin for toString() to be indistinguishable from other types (e.g. 1, 1uL, "1")

but these are numbers, when you read the value you know it's a number, however, when you read a number that belongs to s a snowflake, you have no clue it's a snowflake.

Snowflakes are not just numerical numbers, they hold, again, data like when the entity was created.
So it makes no sense to make a representation for snowflakes that give the impression that they are just numerical values.

Thus this is why asString exist. to allow get the string value of the snowflake, where as toString is for the representation

BenWoodworth:
Yeah I completely agree, but I'm not saying the string representation should give the impression that it's just a number. That wasn't really my point

This is the general rule for `toString()` that you'll find in idiomatic Kotlin:
- If a type has a useful string representation, return that for `toString()`
- Otherwise, return something generic that textually represents it
  - `"SomeClass(a=7)"` is obviously a lot better than `"SomeClass@4cb2c100"`
<br />

This is how you'll find it done everywhere in the official Kotlin libraries. Like `Regex("1234").toString()` is `"1234"`. That's a lot more useful to developers than `"Regex(pattern=1234)"`, even though there's nothing to indicate that `"1234"` is a regex.

I think you could agree that for `Snowflake`, `"1111"` is a lot more useful to programmers than `"Snowflake(value=1111)"` when writing code. The fact that `Snowflake.asString` is used 86 times in this repo while `Snowflake.toString()` isn't even used once. (besides in other `toString()` methods)

There is a good reason why `toString()` is done like this. It's much nicer to write...
```kotlin
CdnUrl("$BASE_URL/guilds/$guildId/users/$userId/avatars/$hash") // this
CdnUrl("$BASE_URL/guilds/${guildId.asString}/users/${userId.asString}/avatars/$hash") // than this
```
<br />

I'm also not sure I understand the concern with not knowing the number is a snowflake. E.g. `"GuildBehavior(id=1111)"` seems okay the same way `{id: "1111", ...}` from the API, clicking **Copy ID** in discord and getting `"1111"` in my clipboard, and `Regex("1234").toString() == "1234"` are okay. If that makes sense...

BenWoodworth:
And sorry for the wall of text. I hope I'm not being annoying because I don't mean to be, I just want to flesh out my thoughts. I've been using Kotlin as my language of choice since it was in beta and have a firm grasp on its conventions. I'm loving Kord so far and want to nudge it in the right direction if can, helping iron out any awkward or non-idiomatic APIs. Don't stress out about this PR if you really don't want to include the change \:)

HopeBaron:
Hey Ben, sorry for the late reply.
I appreciate your help on this. Also thanks for bringing this up.

I looked that up but I didn't find any source that mention such a convension. could you link me to the source?

BenWoodworth:
Hah no worries, and good question! I don't think it's documented anywhere as a hard-and-fast rule, that's just how I see it done in the official libraries. Like there aren't right/wrong ways to implement `toString()`, just more/less useful ways

HopeBaron:
Hey Ben, sorry for not responding, Your points are valid and I don't see a reason to keep the toString as is, users can always define their own logging string on the entities. Tho I wonder how would this affect logging

BenWoodworth:
Hmm, you'd know better than I do since I haven't paid attention to Kord's logging at all so far

HopeBaron:
Will merge this, we just have to wait for Scheduled events and auto complete to be merged.

PR #450, opened 2021-11-22
No longer send Authorization header on endpoints without authorization

DRSchlaubi:
As pointed out by @ByteAlex sending an `Authorization`header on endpoints not requiering it can cause rate-limit issues, this PR changes the current behavior of always sending the header to only sending it when needed﻿

### DIFF
--- a/rest/src/main/kotlin/request/RequestHandler.kt
+++ b/rest/src/main/kotlin/request/RequestHandler.kt
@@ -6,6 +6,11 @@ package dev.kord.rest.request
 interface RequestHandler {
 
     /**
+     * The Discord bot authorization token used on requests.
+     */
+    val token: String
+
+    /**
      * Executes the [request], abiding by the active rate limits and returning the response [R].
      * Throws an [RestRequestException] when a non-rate limit error response is returned.
      */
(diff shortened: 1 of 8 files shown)


rest/src/main/kotlin/request/RequestHandler.kt
```diff
@@ -5,11 +5,16 @@ package dev.kord.rest.request
  */
 interface RequestHandler {
 
+    /**
+     * The Discord bot authorization token used on requests.
+     */
+    val token: String
```
HopeBaron, line 11: I don't really think all request handlers will have concerns over the token, maybe someone would like to handle token related issues else where, is it a requirement for the request handler?

DRSchlaubi: Here yes, or we would need to change lots of code, because neither the request builder nor the call method have access to the token, so without changing 500 method calls this was the easiest way

HopeBaron: No, don't get me wrong, remove the token from the interface and add it to the ktor request handler class. This way, whoever wants to implement their own request handler are not required to use a token-based system

DRSchlaubi: RestService doesn't have a KtorRequestHandler http://rice.by.devs-from.asia/u/CWZfCb.png

PR #477, opened 2022-01-06
Add limit to AuditLogGetRequestBuilder

### LINKED ISSUE #476: Support for auditory log limit parameter
Hey there.

I'm opening this issue as a request to add the limit parameter to the get audit log request. Rather than using a default value. 

Also I attach the documentation link that shows the property I'm talking about.

https://discord.com/developers/docs/resources/audit-log#get-guild-audit-log

Hope someone can add it!
Thanks in advance.

DRSchlaubi:
Fix #476﻿

### DIFF
--- a/rest/src/main/kotlin/builder/auditlog/AuditLogGetRequestBuilder.kt
+++ b/rest/src/main/kotlin/builder/auditlog/AuditLogGetRequestBuilder.kt
@@ -22,5 +22,10 @@ class AuditLogGetRequestBuilder : RequestBuilder<AuditLogGetRequest> {
      */
     var before: Snowflake? = null
 
-    override fun toRequest(): AuditLogGetRequest = AuditLogGetRequest(userId, action, before, 100)
+    /**
+     * How many entries are returned (default 50, minimum 1, maximum 100).
+     */
+    var limit: Int = 100
+
+    override fun toRequest(): AuditLogGetRequest = AuditLogGetRequest(userId, action, before, limit)
 }


rest/src/main/kotlin/builder/auditlog/AuditLogGetRequestBuilder.kt
```diff
@@ -22,5 +22,10 @@ class AuditLogGetRequestBuilder : RequestBuilder<AuditLogGetRequest> {
      */
     var before: Snowflake? = null
 
-    override fun toRequest(): AuditLogGetRequest = AuditLogGetRequest(userId, action, before, 100)
+    /**
+     * How many entries are returned (default 50, minimum 1, maximum 100).
+     */
+    var limit: Int = 100
```
Lukellmann, line 28: the initial value of the property and the default value in the documentation are different, which could be confusing

DRSchlaubi: it defaulted to 100 in Kord before, judging whether this was the right choice isn't part of this PR

Lukellmann: I would suggest updating the documentation to something like:
>How many entries are returned (default 50, minimum 1, maximum 100). ([discord docs](https://discord.com/developers/docs/resources/audit-log#get-guild-audit-log))
100 by default.

DRSchlaubi: I see, well now this is merged, we will need another PR for that

PR #478, opened 2022-01-08
Model fixes

HopeBaron:
Adds missing Invite properties / endpoints
Adds missing PartialGuild endpoints
Adds missing scheduled events behavior and properties


common/src/main/kotlin/entity/DiscordGuild.kt
```diff
@@ -81,82 +81,54 @@ data class DiscordGuild(
...
-    val nsfwLevel: NsfwLevel
+    @SerialName("premium_tier") val premiumTier: PremiumTier,
+    @SerialName("premium_subscription_count") val premiumSubscriptionCount: OptionalInt = OptionalInt.Missing,
+    @SerialName("preferred_locale") val preferredLocale: String,
+    @SerialName("public_updates_channel_id") val publicUpdatesChannelId: Snowflake?,
+    @SerialName("max_video_channel_users") val maxVideoChannelUsers: OptionalInt = OptionalInt.Missing,
+    @SerialName("approximate_member_count") val approximateMemberCount: OptionalInt = OptionalInt.Missing,
+    @SerialName("approximate_presence_count") val approximatePresenceCount: OptionalInt = OptionalInt.Missing,
+    @SerialName("welcome_screen") val welcomeScreen: Optional<DiscordWelcomeScreen> = Optional.Missing(),
+    @SerialName("nsfw_level") val nsfwLevel: NsfwLevel
```
DRSchlaubi, line 131: whilst youre updating guilds, you could also add `guild_scheduled_events` and `premium_progress_bar_enabled`

common/src/main/kotlin/entity/DiscordGuild.kt
```diff
@@ -176,9 +148,15 @@ class DiscordPartialGuild(
...
-    val welcomeScreen: Optional<DiscordWelcomeScreen> = Optional.Missing()
-)
+    @SerialName("welcome_screen") val welcomeScreen: Optional<DiscordWelcomeScreen> = Optional.Missing(),
+    @SerialName("vanity_url_code") val vanityUrlCode: Optional<String?> = Optional.Missing(),
+    val description: Optional<String?> = Optional.Missing(),
+    val banner: Optional<String?> = Optional.Missing(),
+    val splash: Optional<String?> = Optional.Missing(),
+    @SerialName("nsfw_level") val nsfwLevel: Optional<NsfwLevel> = Optional.Missing(),
+    @SerialName("verification_level")
+    val verificationLevel: Optional<VerificationLevel> = Optional.Missing()
```
DRSchlaubi, line 158: same for partial guilds?

DRSchlaubi (changes requested):

common/src/main/kotlin/entity/DiscordGuild.kt
Lukellmann: `stage_instances` and `stickers` are also missing

Lukellmann (changes requested):

common/src/main/kotlin/entity/DiscordGuild.kt
HopeBaron: But I added them
 https://github.com/kordlib/kord/pull/478/files/c599823687518aec6522c8203f5ee75c408e191b#diff-81b031b8fa9334a46827df60852dd1720abfbd54770d8c1b1ab62bbe1ce9ba0bR151-R158

HopeBaron:
I've added the said properties, however, there are some conflicts between behaviors and the data of the guild, like scheduledEvents. read #147

common/src/main/kotlin/entity/DiscordGuild.kt
```diff
@@ -81,82 +81,61 @@ data class DiscordGuild(
...
+    @SerialName("approximate_presence_count") val approximatePresenceCount: OptionalInt = OptionalInt.Missing,
+    @SerialName("welcome_screen") val welcomeScreen: Optional<DiscordWelcomeScreen> = Optional.Missing(),
+    @SerialName("nsfw_level") val nsfwLevel: NsfwLevel,
+    @SerialName("stage_instances")
+    val stageInstances: Optional<List<DiscordStageInstance>> = Optional.Missing(),
+    val stickers: Optional<List<DiscordMessageSticker>> = Optional.Missing(),
+    @SerialName("guild_scheduled_events")
+    val guildScheduledEvents: Optional<List<DiscordGuildScheduledEvent>> = Optional.Missing(),
+    @SerialName("premium_progress_bar_enabled")
+    val premiumProgressBarEnabled: OptionalBoolean = OptionalBoolean.Missing
```
Lukellmann, line 138: Is this really optional?

core/src/main/kotlin/cache/data/GuildData.kt
```diff
@@ -58,7 +58,11 @@ public data class GuildData(
     val approximatePresenceCount: OptionalInt = OptionalInt.Missing,
     val welcomeScreen: Optional<WelcomeScreenData> = Optional.Missing(),
     val nsfwLevel: NsfwLevel,
-    val threads: Optional<List<ChannelData>> = Optional.Missing()
+    val threads: Optional<List<ChannelData>> = Optional.Missing(),
+    val stageInstances: Optional<List<StageInstanceData>> = Optional.Missing(),
+    val stickers: Optional<List<MessageStickerData>> = Optional.Missing(),
+    val guildScheduledEvents: Optional<List<GuildScheduledEventData>> = Optional.Missing(),
+    val premiumProgressBarEnabled: OptionalBoolean = OptionalBoolean.Missing
 ) {
```
Lukellmann, line 66: Same here: is it really optional?

core/src/main/kotlin/supplier/CacheEntitySupplier.kt
```diff
@@ -440,6 +442,38 @@ public class CacheEntitySupplier(private val kord: Kord) : EntitySupplier {
...
+    public override fun getGuildScheduledEventUsersAfter(
+        guildId: Snowflake,
+        eventId: Snowflake,
+        limit: Int,
+        withMember: Boolean?,
+        after: Snowflake
+    ): Flow<User> {
+        val flow = cache.query<UserData> { idGt(UserData::id, after) }.asFlow()
+            .map { User(it, kord) }
+        return if (limit != Int.MAX_VALUE) flow.take(limit) else flow
```
Lukellmann, line 474: Implementations are different.

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
```diff
@@ -26,17 +26,14 @@ import kotlin.contracts.contract
...
      * @throws RequestException if anything goes wrong during the request
      */
-    public val users: Flow<User> get() = flow {
-        kord.rest.guild.getScheduledEventUsers(guildId, id).users.forEach {
-            val userData = UserData.from(it)
-            emit(User(userData, kord, supplier))
-        }
-    }
+    @Deprecated("Use getUsersBefore or getUsersAfter")
+    public val users: Flow<User> get() = supplier.getGuildScheduledEventUsersBefore(guildId, id)
```
Lukellmann, line 36: Why deprecate it? There is `supplier.getGuildScheduledEventUsers()`

core/src/main/kotlin/supplier/EntitySupplier.kt
```diff
@@ -473,6 +473,31 @@ public interface EntitySupplier {
...
+        withMember: Boolean? = null,
+        after: Snowflake = Snowflake.min
+    ): Flow<User>
+
+    public fun getGuildScheduledEventUsers(
+        guildId: Snowflake,
+        eventId: Snowflake,
+        limit: Int = Int.MAX_VALUE,
+        withMember: Boolean? = null,
+    ): Flow<User> = getGuildScheduledEventUsersAfter(guildId, eventId, limit, withMember)
```
Lukellmann, line 498: Is there a `GuildScheduledEventUser` type? That's what the endpoint should return, see https://discord.com/developers/docs/resources/guild-scheduled-event#get-guild-scheduled-event-users

core/src/main/kotlin/supplier/RestEntitySupplier.kt
```diff
@@ -272,13 +272,25 @@ public class RestEntitySupplier(public val kord: Kord) : EntitySupplier {
...
     }
 
-    public suspend fun getInviteOrNull(code: String, withCounts: Boolean): Invite? = catchNotFound {
-        val response = invite.getInvite(code, withCounts)
+    public suspend fun getInviteOrNull(
+        code: String,
+        withCounts: Boolean,
+        withExpiration: Boolean = true,
+        scheduledEventId: Snowflake? = null
+    ): Invite? = catchNotFound {
```
Lukellmann, line 280: Default for `withCounts`?

core/src/main/kotlin/supplier/RestEntitySupplier.kt
```diff
@@ -475,6 +487,40 @@ public class RestEntitySupplier(public val kord: Kord) : EntitySupplier {
...
+        guildId: Snowflake,
+        eventId: Snowflake,
+        limit: Int,
+        withMember: Boolean?,
+        after: Snowflake
+    ): Flow<User> {
+
+        val batch = min(100, limit)
+        return paginateBackwards(after, batch, { it.id }) { position ->
+            kord.rest.guild.getScheduledEventUsers(guildId, eventId, batch, withMember, position).users
```
Lukellmann, line 518: Should use `paginateForwards`.

core/src/test/kotlin/UtilKtTest.kt
```diff
@@ -2,15 +2,15 @@ package dev.kord.core
...
+import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Assertions
 import org.junit.jupiter.api.Test
 
 internal class UtilKtTest {
 
     @Test
-    fun `paginate forwards selects the right id`() = runBlockingTest {
-
+    @ExperimentalStdlibApi
```
Lukellmann, line 12: `@ExperimentalStdlibApi` not needed, `buildList` is stable now

core/src/test/kotlin/UtilKtTest.kt
```diff
@@ -27,7 +27,8 @@ internal class UtilKtTest {
     }
 
     @Test
-    fun `paginate backwards selects the right id`() = runBlockingTest {
+    @ExperimentalStdlibApi
```
Lukellmann, line 30: same here

Lukellmann (changes requested):

core/src/main/kotlin/cache/data/GuildData.kt
HopeBaron: Yes, tests were failing

common/src/main/kotlin/entity/DiscordGuild.kt
HopeBaron: Yes, tests were failing

Lukellmann: Because of Discord or the JSON test files?

HopeBaron: I'll give it another look; I defintely didn't update the json file

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
HopeBaron: because the values are hardcoded are not interchangeable

core/src/main/kotlin/supplier/EntitySupplier.kt
HopeBaron: I didn't add it yet

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
Lukellmann: Then you would also have to remove `EntitySupplier.getGuildScheduledEventUsers()`, right?

Lukellmann:
If you plan to merge 0.8.x into this PR, you could also make `limit` nullable in suppliers and use the new `limitedPagination()`, `checkLimit()` and `limit()` functions.

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
HopeBaron: I did that already (used the after function) however, it's generally discouraged to use. I refactored it a bit and removed the deprecation

core/src/main/kotlin/supplier/EntitySupplier.kt
HopeBaron: It's kinda hard to decide wither this should be it's own thing or not; closed I got was the UserOptionallyMember entity 
as the given event Id is already at hand, for lower level representations they already have access to the full payload

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
Lukellmann: 👍
Although I would probably still use the `EntitySupplier` method instead of the `Behavior` method just in case it is overriden in any Supplier implementation

core/src/main/kotlin/supplier/EntitySupplier.kt
Lukellmann: I would also make `before`/`after` the first parameter in their functions to be consistent with `getMessages...`, same in `Behavior`.

DRSchlaubi (approved):

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
HopeBaron: but it does use the supplier; the behavior function just calls the supplier underneath; so just to reduce code duplication

Lukellmann: Let me clarify:
I would use
```kotlin
supplier.getGuildScheduledEventUsers(guildId, id, withMember = true)
```
and not one of these
```kotlin
supplier.getGuildScheduledEventUsersAfter(guildId, id, withMember = true)
supplier.getGuildScheduledEventUsersBefore(guildId, id, withMember = true)
this.getGuildScheduledEventUsersAfter()
this.getGuildScheduledEventUsersBefore()
````
Otherwise `EntitySupplier.getGuildScheduledEventUsers()` wouldn't be necessary since it isn't used elsewhere.


#

Unrelated but also noticed that it's `withMember` in some places and `withMembers` in other places.

Comment (HopeBaron) [reply core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt]: Depends on the endpoint's naming schema.

Supplier isn't private in anyway; so if we don't use it elsewhere doesn't mean we restrict it's use

The functions in the instance are just convincance methods it's unlikely you are going to do 

```kt
this.this.getGuildScheduledEventUsersAfter()
```
and specify the id and guild each time

core/src/main/kotlin/supplier/EntitySupplier.kt
HopeBaron: Im still thinking of how to model this; the endpoint expresses a "garenteed user" but an "optional member"
However, we cache the member and provide the user

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
Lukellmann: I think we are talking past each other, I'll try to be clear now:

I would suggest having a call chain like this (arrows representing which methods/getters call others):
```
GuildScheduledEventBehavior                EntitySupplier

users -----------------------------------> getGuildScheduledEventUsers() ----------+
                                                                                   |
getGuildScheduledEventUsersAfter()  ---->  getGuildScheduledEventUsersAfter()  <---+

getGuildScheduledEventUsersBefore()  --->  getGuildScheduledEventUsersBefore()
```
instead of this:
```
GuildScheduledEventBehavior                EntitySupplier

users ------+                              getGuildScheduledEventUsers() ----------+
            |                                                                      |
            v                                                                      |
getGuildScheduledEventUsersAfter()  ---->  getGuildScheduledEventUsersAfter()  <---+

getGuildScheduledEventUsersBefore()  --->  getGuildScheduledEventUsersBefore()
```
This way `GuildScheduledEventBehavior` would just delegate to the Suppliers and let them handle the fallback in case of all users (the fallback 'logic' is not duplicated), feels like a better separation of concerns to me.

Lukellmann: > Depends on the endpoint's naming schema.

If you mean `withMember` vs `withMembers`: it's literally `withMember` in `EntitySupplier.getGuildScheduledEventUsers[Before|After]()` and `withMembers` in `GuildScheduledEventBehavior.getGuildScheduledEventUsers[Before|After]()`

core/src/main/kotlin/supplier/EntitySupplier.kt
Lukellmann: Another idea: a second set of methods would also be an option, so you can't provide a `withMember` option but instead call another method.
But we might also just go all in for a `GuildScheduledEventUser` entity, maybe Discord will use it more in the future / add new fields to it.

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
HopeBaron: > I would suggest having a call chain like this (arrows representing which methods/getters call others):

You are looking @ outdated  file; users actually the after function of the guild itself so user calls the function defined in the same file.
I Fixed the typo in regards to the parameter name

core/src/main/kotlin/supplier/EntitySupplier.kt
HopeBaron: So one function returns Users; the other returns Members
Perfecto

HopeBaron: Done

core/src/main/kotlin/behavior/GuildScheduledEventBehavior.kt
```diff
@@ -27,18 +23,32 @@ import kotlin.contracts.contract
...
      */
-    public val users: Flow<User> get() = flow {
-        kord.rest.guild.getScheduledEventUsers(guildId, id).users.forEach {
-            val userData = UserData.from(it)
-            emit(User(userData, kord, supplier))
-        }
-    }
 
+    public val users: Flow<User>
+        get() = getGuildScheduledEventUsersAfter()
```
Lukellmann, line 34:
```suggestion
    public val users: Flow<User>
        get() = supplier.getGuildScheduledEventUsers(guildId, id)
```
this is what I mean, should have used the suggestion feature before but had to find it first

core/src/main/kotlin/supplier/EntitySupplier.kt
Lukellmann: Also a `getGuildScheduledEventMembers()` or only for users?

PR #480, opened 2022-01-12
Add support for Interaction localization

DRSchlaubi:
https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object

### DIFF
--- a/common/src/main/kotlin/Locale.kt
+++ b/common/src/main/kotlin/Locale.kt
@@ -12,7 +12,7 @@ import java.util.Locale as JLocale
 /**
  * Representation of a locale.
  *
- * @property language an ISO 369 language code representing the language
+ * @property language an ISO 639 language code representing the language
  * @property country an ISO 3166 country code representing the country
  */
 @Serializable(with = Locale.Serializer::class)
--- a/common/src/main/kotlin/Locale.kt
+++ b/common/src/main/kotlin/Locale.kt
@@ -1,0 +1,60 @@
+package dev.kord.common
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import java.util.Locale as JLocale
+
+/**
+ * Representation of a locale.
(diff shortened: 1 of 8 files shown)


HopeBaron (approved):

Lukellmann (approved):

Lukellmann:
https://github.com/discord/discord-api-docs/commit/c39627c76f049c9f3c26fe40edde75017b76ba7e: would it make sense to
a) change `Locale` into a sealed class / interface with `object` subtypes for the listed locales and a `data class Unknown` subtype for unknown ones
b) use the `Locale` type in other places too (see Links in that commit) - probably something for another PR

DRSchlaubi:
> [discord/discord-api-docs@c39627c](https://github.com/discord/discord-api-docs/commit/c39627c76f049c9f3c26fe40edde75017b76ba7e): would it make sense to a) change `Locale` into a sealed class / interface with `object` subtypes for the listed locales and a `data class Unknown` subtype for unknown ones b) use the `Locale` type in other places too (see Links in that commit) - probably something for another PR

Ofc they add this after I made this PR sigh

But yes I aggree we should add some logic for that, although I am uncertain whether we should use subtypes or static variables storing an instance for every language like `java.util.Locale`

Anyway as this would also affect different entities (as you mentioned previously) I would move that into another PR

Lukellmann:
Yeah, it's probably better to just have some constants as static variables, they don't really have their own type after all.

common/src/main/kotlin/Locale.kt
```diff
@@ -0,0 +1,245 @@
...
+         * Decodes the language from a `languageCode_countryCode` or `languageCode` format.
+         */
+        fun fromString(string: String): Locale {
+            val match = languageTagFormat.matchEntire(string) ?: error("$string is not a valid Locale")
+            val (language) = match.destructured
+            val country = match.groupValues[2]
+
+            return ALL.firstOrNull { (l, c) ->
+                language == l && country == (c ?: "")
+            } ?: error("This locale is not supported by Discord")
```
Lukellmann, line 226: Should this really throw an exception? There is still a public constructor.

DRSchlaubi: So you mean I should just default to Locale(country, language)?

Lukellmann: Yes, this way Discord randomly adding new locales won't cause exceptions.

HopeBaron: This is a common pattern in our `types` values, making the system error tolerant when Discord adds something to this.

PR #481, opened 2022-01-12
Clarify limit of AuditLogGetRequestBuilder

Lukellmann:
See https://github.com/kordlib/kord/pull/477#pullrequestreview-845729652.

I would actually like to make `limit` nullable and set it to `null` by default, since this parameter is optional like the others, but I'm not sure if this change could be justified (it would change the default behavior). Would be nice to get some feedback about this.


HopeBaron:
In core, we use this to paginate through results, null would be kinda annoying to work with. Int.MAX_VALUE fetches full values.

Lukellmann:
> In core, we use this to paginate through results, null would be kinda annoying to work with. Int.MAX_VALUE fetches full values.

Not really, it is only use

Lukellmann:
> In core, we use this to paginate through results, null would be kinda annoying to work with. Int.MAX_VALUE fetches full values.

Fair point, then it should be set to 50 probably.

While looking into this I also found some other problems with pagination in `RestEntitySupplier`, so I will widen the scope of this PR for fixing those issues too.

PR #485, opened 2022-01-16
Add support for InteractionCreateEvent in LiveMessage

will421:
First PR on Github and started Kotlin not long ago

Thought it was weird that InteractionCreateEvent was not supported in LiveMessage.

Condition in update() is doing nothing but wanted to keep consistency.
Still need to cast the interaction to ComponentInteraction in the event consumer..

### DIFF
--- a/core/src/main/kotlin/live/LiveMessage.kt
+++ b/core/src/main/kotlin/live/LiveMessage.kt
@@ -11,6 +11,7 @@ import dev.kord.core.entity.ReactionEmoji
 import dev.kord.core.event.Event
 import dev.kord.core.event.channel.ChannelDeleteEvent
 import dev.kord.core.event.guild.GuildDeleteEvent
+import dev.kord.core.event.interaction.InteractionCreateEvent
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/live/LiveMessage.kt
```diff
@@ -162,6 +163,8 @@ public class LiveMessage(
         is ChannelDeleteEvent -> event.channel.id == message.channelId
 
         is GuildDeleteEvent -> event.guildId == guildId
+
+        is InteractionCreateEvent -> event.interaction.data.message.value?.id == message.id
```
Lukellmann, line 167: The event might also be related to the message if a message command was invoked on it (`MessageCommandInteractionCreateEvent`).

HopeBaron: But that doesn't affect it's data

Lukellmann: Hm  true, I'm not so familiar with Live Entities yet to be honest.

PR #489, opened 2022-01-18
Fix / add entdpoints without authorization, remove shards field from GatwayResponse

Lukellmann:
- require authorization for `WebhookPatch`
- don't require authorization for `ExecuteWebhookPost`, `ExecuteSlackWebhookPost`, `ExecuteGithubWebhookPost`, `EditWebhookMessage`, `FollowupMessageCreate`, `FollowupMessageModify` and `FollowupMessageDelete` (see https://github.com/discord/discord-api-docs/pull/4349 regarding followups)
- remove `shards` field from `GatwayResponse`, see https://discord.com/developers/docs/topics/gateway#get-gateway or https://discord.com/api/v9/gateway
- make private shorthand getters in `EntitySupplier`s `inline`
- `null` as default for all `reason: String?` parameters

### DIFF
--- a/core/src/main/kotlin/behavior/channel/TopGuildChannelBehavior.kt
+++ b/core/src/main/kotlin/behavior/channel/TopGuildChannelBehavior.kt
@@ -89,7 +89,7 @@ public interface TopGuildChannelBehavior : GuildChannelBehavior {
      * @param reason the reason showing up in the audit log
      * @throws [RestRequestException] if something went wrong during the request.
      */
-    public suspend fun addOverwrite(overwrite: PermissionOverwrite, reason: String?) {
+    public suspend fun addOverwrite(overwrite: PermissionOverwrite, reason: String? = null) {
         kord.rest.channel.editChannelPermissions(
             channelId = id,
             overwriteId = overwrite.target,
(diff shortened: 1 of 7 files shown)


HopeBaron:
https://discord.com/developers/docs/topics/gateway#get-gateway-bot
is the endpoint we use.

Lukellmann:
> https://discord.com/developers/docs/topics/gateway#get-gateway-bot
> is the endpoint we use. so Shards are actually there

That endpoint returns a `BotGatewayResponse`, take a look at where `GatewayResponse` is used.

DRSchlaubi:
> https://discord.com/developers/docs/topics/gateway#get-gateway-bot is the endpoint we use. so Shards are actually there

We don't actually use that class, the builder has "auto-sharding" but it doesn't decode to that class

Lukellmann:
Kind of, it's only used in one Route that someone could call manually.

HopeBaron:
The question is; if shards were of type int, how come it didn't throw on it when the field didn't exist :thinking:

Lukellmann:
There are two endpoints for getting the gatway info, one without auth and one with auth: https://discord.com/developers/docs/topics/gateway#get-gateway.   
The shard info only exists on the one with auth, what I changed here is the response of the endpoint without auth. It didn't throw before because the enpoint without auth isn't used anywhere in Kord.   
You can verify that it currently throws by calling
```kotlin
val gatwayResponse = kord.rest.unsafe(Route.GatewayGet) {}
```

PR #491, opened 2022-01-20
Deprecate Stage Discovery, stickers in GuildPreview, createTimestamp for Threads

Lukellmann:
- deprecate Stage Discovery, see https://github.com/discord/discord-api-docs/issues/3152, https://github.com/discord/discord-api-docs/pull/4191 and https://github.com/discord/discord-api-docs/pull/4296
- add new `stickers` field to `DiscordGuildPreview`, `GuildPreviewData` and `GuildPreview`, see https://github.com/discord/discord-api-docs/commit/5b023e04b3bcc055b5be83ad0562862431620e4c
- add new `createTimestamp` field/property to `DiscordThreadMetadata`, `ThreadMetadataData` and `ThreadData`, see https://github.com/discord/discord-api-docs/commit/53ee0109b874bde5ec2316379bf266bb332630a0
- add missing `isInvitable` property to `TextChannelThread` (`NewsChannelThread`s are always public so it would be useless there), see https://discord.com/developers/docs/resources/channel#thread-metadata-object
- `data class` for `DiscordThreadMetadata`

### DIFF
--- a/common/src/main/kotlin/entity/DiscordChannel.kt
+++ b/common/src/main/kotlin/entity/DiscordChannel.kt
@@ -4,6 +4,7 @@ import dev.kord.common.entity.optional.Optional
 import dev.kord.common.entity.optional.OptionalBoolean
 import dev.kord.common.entity.optional.OptionalInt
 import dev.kord.common.entity.optional.OptionalSnowflake
+import kotlinx.datetime.Instant
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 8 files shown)


Lukellmann:
I made `createTimestamp` of type `Optional<Instant>`, however most other timestamps documented to be ISO8601 are of type `String` in kord. Is there a reason for this as `createTimestamp` seems to (de-)serialize just fine.

DRSchlaubi:
> I made `createTimestamp` of type `Optional<Instant>`, however most other timestamps documented to be ISO8601 are of type `String` in kord. Is there a reason for this as `createTimestamp` seems to (de-)serialize just fine.

The answer is in the Serializer of `Instant`: https://github.com/Kotlin/kotlinx-datetime/blob/5a920eaee7d640f4f5f7c956eb6706f7b333de81/core/common/src/Instant.kt#L33

Before kx.dt hat support for kx.ser we didn't really have a way of deserializing timestamps and did that in core using java.time,

#297  did update everything to kx.dt, however it didn't update all entites in common, so that's why there are still some left as strings

HopeBaron:
Was this before merging the Sticker Revamp branch; since Im sure I renamed MessageSticker to Sticker

Lukellmann:
> Was this before merging the Sticker Revamp branch; since Im sure I renamed MessageSticker to Sticker

No, this is after the sticker revamp, only the data class in common is still named `DiscordMessageSticker`, in core it's `StickerData` and `Sticker`.

Lukellmann:
It seems like there is an issue with the type of `create_timestamp` in gatway events: https://github.com/discord/discord-api-docs/issues/4418

Lukellmann:
> It seems like there is an issue with the type of `create_timestamp` in gateway events

I'm not able to reproduce this though.

PR #492, opened 2022-01-20
Add disabled property to SelectMenuComponent

### LINKED ISSUE #487: SelectMenuComponent is missing a disabled field
The SelectMenuComponent is missing the disabled field it is present in ComponentData and the discord docs say that the disabled field is applicable to Select Menus.

Edit: https://discord.com/developers/docs/interactions/message-components#select-menu-object-select-menu-structure

DRSchlaubi:
Fix #487

### DIFF
--- a/core/src/main/kotlin/entity/component/SelectMenuComponent.kt
+++ b/core/src/main/kotlin/entity/component/SelectMenuComponent.kt
@@ -40,6 +40,11 @@ public class SelectMenuComponent(override val data: ComponentData) : Component {
      */
     public val maxValues: Int get() = data.maxValues.orElse(1)
 
+    /**
+     * Disable the select, default `false`
+     */
+    public val disabled: Boolean get() = data.disabled.discordBoolean
+
     override fun equals(other: Any?): Boolean {
         if (other !is SelectMenuComponent) return false
 


Lukellmann (commented):
Maybe name it `isDisabled`?

DRSchlaubi:
Slash commands called it disabled too

Lukellmann:
I'm fine with that too, just noticed that some names in core entities are a little different, like `isNsfw` or `isArchived` in `ThreadChannel`.  
The documentation of `disabled` also makes it seem like the property is mutable.

HopeBaron:
do we have it in create/modify requests?

Lukellmann:
> Slash commands called it disabled too

And in `Button` it's called `disabled` too, so don't mind my naming suggestion, would still change the doc comment though.

Lukellmann:
> do we have it in create/modify requests?

Seems so, they use `SelectMenuBuilder` https://github.com/kordlib/kord/blob/0.8.x/rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt

HopeBaron:
Ya both had different PRs under the same tasks  you got me comparing both :P

PR #493, opened 2022-01-20
Fix issues with autocomplete serialization

DRSchlaubi:
Users can enter strings into numeric types in auto complete


core/src/main/kotlin/entity/interaction/ContextInteraction.kt
```diff
@@ -225,6 +226,8 @@ public class UnknownApplicationCommandInteraction(
  *
  * **Follow-ups and normals responses don't work on this type**
  *
+ * **No matter what argument type is used all arguments will be [CommandArgument.AutoCompleteArgument]s
```
Lukellmann, line 229: Only focused arguments, right?

DRSchlaubi: yes

PR #497, opened 2022-01-23
Preserve binary compatibility for ButtonBuilder

Lukellmann:
https://github.com/kordlib/kord/pull/494 changed the type of `ButtonBuilder.disabled` from `kotlin.Boolean` to `kotlin.Boolean?`.
On Jvm this means that it will no longer be of the primitve type `boolean` but its wrapper `java.lang.Boolean` instead, which is a binary incompatible change, here's how it got fixed:

This PR adds `ButtonBuilder.primitiveDisabledForBinaryCompatibility`, a property of type `kotlin.Boolean`, and renames its getter and setter to give them the same signatures as the previous `ButtonBuilder.disabled`. The implementation just delegates to `ButtonBuilder.disabled`.
It is given a deprecation level of `HIDDEN` to only allow it to be accessed from already compiled classes but not newly written/compiled source code.

### DIFF
--- a/rest/src/main/kotlin/builder/component/ButtonBuilder.kt
+++ b/rest/src/main/kotlin/builder/component/ButtonBuilder.kt
@@ -13,6 +13,21 @@ import dev.kord.common.entity.optional.delegate.delegate
 @KordDsl
 sealed class ButtonBuilder : ActionRowComponentBuilder() {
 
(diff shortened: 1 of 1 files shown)


HopeBaron:
there is no need for this; you can just break the compatibility rather than creating hacks

Lukellmann:
Alright, will close then

PR #506, opened 2022-01-28
Added missing properties for attachments.

bvanseg:


core/src/main/kotlin/entity/Attachment.kt
```diff
@@ -23,6 +24,16 @@ public data class Attachment(val data: AttachmentData, override val kord: Kord)
...
 
+    /**
+     * The description for the file.
+     */
+    val description: Optional<String> get() = data.description
+
+    /**
+     * The attachment's [media type](https://en.wikipedia.org/wiki/Media_type).
+     */
+    val contentType: Optional<String> get() = data.contentType
```
Lukellmann, line 35: Use nullable types instead of `Optional` in core entities.
```suggestion
    /**
     * The description for the file.
     */
    val description: String? get() = data.description.value

    /**
     * The attachment's [media type](https://en.wikipedia.org/wiki/Media_type).
     */
    val contentType: String? get() = data.contentType.value
```

Lukellmann (approved):

HopeBaron:
There seem to be no set of pre-defined media types so. I guess we are fine with this thanks @bvanseg

PR #512, opened 2022-01-30
Revert change to `choices` type in `BaseChoiceBuilder` from #505

Lukellmann:
There is currently an issue where you can't set a nullable star projected property to `null`.
I opened [an issue in the Kotlin YouTrack](https://youtrack.jetbrains.com/issue/KT-51045) for this.
To temporarily fix this I reverted the type change made in #505.


HopeBaron:
Already pushed a fix a few minutes ago

Lukellmann:
> Already pushed a fix a few minutes ago

On 0.8.x? Don't see it...

HopeBaron:
Seems that I escaped it by mistake; will merge this then

Issue #513, opened 2022-01-31
Scheduled Events serializer is broken


NikkyAI:
using kord-extension 1.5.2-SNAPSHOT and kord 0.8.x-SNAPSHOT

when trying to list scheduled events:

```
2022-01-31 00:54:44.880 ERROR com.kotlindiscord.kord.extensions.commands.application.slash.SlashCommand : Error during execution of list slash command (dev.kord.core.event.interaction.GuildChatInputCommandInteractionCreateEvent@6054ac99)
kotlinx.serialization.MissingFieldException: Fields [channelId, user_count] are required for type with serial name 'dev.kord.common.entity.DiscordGuildScheduledEvent', but they were missing
    at kotlinx.serialization.internal.PluginExceptionsKt.throwMissingFieldException(PluginExceptions.kt:20)
    at dev.kord.common.entity.DiscordGuildScheduledEvent.<init>(DiscordGuildScheduledEvent.kt:34)
    at dev.kord.common.entity.DiscordGuildScheduledEvent$$serializer.deserialize(DiscordGuildScheduledEvent.kt:34)
    at dev.kord.common.entity.DiscordGuildScheduledEvent$$serializer.deserialize(DiscordGuildScheduledEvent.kt:34)
    at kotlinx.serialization.json.internal.PolymorphicKt.decodeSerializableValuePolymorphic(Polymorphic.kt:59)
    at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:36)
    at kotlinx.serialization.encoding.AbstractDecoder.decodeSerializableValue(AbstractDecoder.kt:43)
    at kotlinx.serialization.encoding.AbstractDecoder.decodeSerializableElement(AbstractDecoder.kt:70)
    at kotlinx.serialization.encoding.CompositeDecoder$DefaultImpls.decodeSerializableElement$default(Decoding.kt:535)
```

code:
```kt
val messages = guild.scheduledEvents.toList().map { event ->
    """
    description: ${event.description}
    start: ${event.scheduledStartTime}
    end: ${event.scheduledEndTime}
    status: ${event.status}
    type: ${event.entityType}
    type: ${event.entityMetadata}
    data: ${event.data}
""".trimIndent()
}
```

when trying to create a scheduled event:

[...]

PR #514, opened 2022-01-31
Fixes for #513 (Guild Scheduled Events)

Lukellmann:
### Fixes, see https://discord.com/developers/docs/resources/guild-scheduled-event
- add missing `@SerialName("snake_case_name")` annotations
- make `creatorId` nullable instead of optional
- make `entityMetadata` nullable
- change type of `userCount` from `Int` to `OptionalInt`

### create and modify endpoints
- support audit log reason
- modify: `channelId` is nullable and optional now, was only optional before

### Remove no longer documented things
- `ScheduledEntityType.None`, see https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-guild-scheduled-event-entity-types
- `GuildScheduledEventEntityMetadata.speakerIds`, see https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-guild-scheduled-event-entity-metadata

### New type for `privacyLevel`: `GuildScheduledEventPrivacyLevel`
This was `StageInstancePrivacyLevel` before but they are documented as two independent types in the docs, see https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-guild-scheduled-event-privacy-level and https://discord.com/developers/docs/resources/stage-instance#stage-instance-object-privacy-level.

### Other
- move `GuildScheduledEventCreateRequest` from `GuildRequests.kt` to `ScheduledEventRequests.kt`

### DIFF
--- a/common/src/main/kotlin/entity/DiscordGuildScheduledEvent.kt
+++ b/common/src/main/kotlin/entity/DiscordGuildScheduledEvent.kt
@@ -113,11 +113,9 @@ public sealed class GuildScheduledEventStatus(public val value: Int) {
 /**
  * Entity metadata for [DiscordGuildScheduledEvent].
  *
- * @property speakerIds the speakers of the stage channel
  * @property location location of the event
  */
 @Serializable
 public data class GuildScheduledEventEntityMetadata(
-    val speakerIds: Optional<List<Snowflake>> = Optional.Missing(),
     val location: Optional<String> = Optional.Missing()
 )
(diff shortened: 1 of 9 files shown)


Lukellmann:
Would be nice to have some opinions on removing no longer documented stuff:
Should we actually remove or just keep it (maybe with a deprecation)?

Also: should we introduce the new `GuildScheduledEventPrivacyLevel` or just keep `StageInstancePrivacyLevel`?

PR #520, opened 2022-02-04
Generic interaction create events

Lukellmann:
## Changes

* make `InteractionCreateEvent` generic over its `interaction` property
* change superinterface of `GuildComponentInteractionCreateEvent` and `GlobalComponentInteractionCreateEvent`

#### before
```
Event <- InteractionCreateEvent <- [Guild|Global]ComponentInteractionCreateEvent
```
#### after
```
Event <- InteractionCreateEvent <- ComponentInteractionCreateEvent <- [Guild|Global]ComponentInteractionCreateEvent
```

#### Without generic interaction
```kotlin
suspend inline fun <E : InteractionCreateEvent> E.foo(ack: E.() -> InteractionResponseBehavior) {
    val response = this.ack()
    response.edit { /* ... */ }
}

suspend fun ButtonInteractionCreateEvent.bar() {
    // you have to obtain `interaction` here to access all its methods
    foo { interaction.acknowledgePublicDeferredMessageUpdate() }
}

suspend fun ChatInputCommandInteractionCreateEvent.baz() {
    foo { interaction.acknowledgePublic() }
}
```

#### With generic interaction
```kotlin
suspend inline fun <I : Interaction, E : InteractionCreateEvent<I>> E.foo(ack: I.() -> InteractionResponseBehavior) {
    val response = interaction.ack()
    response.edit { /* ... */ }
}

suspend fun ButtonInteractionCreateEvent<*>.bar() {
    // access methods of interaction directly, its type is know through generics
    foo { acknowledgePublicDeferredMessageUpdate() }
}

suspend fun ChatInputCommandInteractionCreateEvent<*>.baz() {
    foo { acknowledgePublic() }
}
```

### DIFF
--- a/core/src/main/kotlin/event/interaction/ApplicationCreate.kt
+++ b/core/src/main/kotlin/event/interaction/ApplicationCreate.kt
@@ -1,12 +1,7 @@
 package dev.kord.core.event.interaction
 
 import dev.kord.core.Kord
-import dev.kord.core.behavior.interaction.AutoCompleteInteractionBehavior
-import dev.kord.core.behavior.interaction.EphemeralInteractionResponseBehavior
-import dev.kord.core.behavior.interaction.PublicInteractionResponseBehavior
-import dev.kord.core.behavior.interaction.followUp
-import dev.kord.core.behavior.interaction.respondEphemeral
-import dev.kord.core.behavior.interaction.respondPublic
+import dev.kord.core.behavior.interaction.*
 import dev.kord.core.entity.application.ApplicationCommand
 import dev.kord.core.entity.interaction.*
 import dev.kord.core.event.kordCoroutineScope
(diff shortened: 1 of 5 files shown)


HopeBaron:
I don't get it; you are calling the function in their class
For example
```kt
suspend fun ChatInputCommandInteractionCreateEvent<*>.baz() {
    foo { acknowledgePublic() }
}
```

You already assume it's type is ChatInputCommandInteraction at the very least

Lukellmann:
> I don't get it; you are calling the function in their class
> For example
> ```kt
> suspend fun ChatInputCommandInteractionCreateEvent<*>.baz() {
>     foo { acknowledgePublic() }
> }
> ```
> 
> You already assume it's type is ChatInputCommandInteraction at the very least

Yes, but `acknowledgePublic()` is still  a method on the `interaction` of the event, not the event itself. If it isn't generic, there would be no way to have the correct interaction type directly in scope of the lambda in `foo()`.

Lukellmann:
waiting for #531

Lukellmann:
After thinking about this for a while I'm not sure if this change is actually a good idea. The major downside comes with Kotlin's syntax for generic `is` checks. If you want to listen for an event that isn't a concrete class (e.g. `ButtonInteractionCreateEvent`), you would have to add a star projection:
```kotlin
kord.on<ButtonInteractionCreateEvent<*>> {
    // ...
}
```
This does not only look somewhat strange but will also break the compilation for a lot of (if not all) projects using kord.

Lukellmann:
I'm closing this for now, we can reopen it if needed.

PR #530, opened 2022-02-09
Invite and Application revamp

Lukellmann:
* update `ApplicationInfo` fields in common, cache and entity and rename to `Application`
* add `ApplicationFlag` and `ApplicationFlags`
* update `Invite` fields in common, cache and entity
* include `InviteMetadata` fields in `DiscordInviteWithMetadata`, it is not a separate type
* deprecate `TargetUserType`
* add `InviteTargetType` (https://discord.com/developers/docs/resources/invite#invite-object-invite-target-types)
* remove `UserFlag.None`, it's no flag and contained in every `UserFlags` (`0 & x == 0`)
* add `UserFlag.BotHttpInteractions`
* add `CategorizableChannelBehavior`: `invites` (previously on `TopGuildChannelBehavior`) and `createInvite()` (previously on `CategorizableChannel`), this is the new supertype of all `TopGuildChannelBehavior`s except `CategoryBehavior`, categories can not be invited to

PR #533, opened 2022-02-12
Allow getting ephemeral followups

Lukellmann:
### Changes
* change return type of `getFollowupMessage[OrNull]()` in `InteractionResponseBehavior` and `EntitySupplier`s from `PublicFollowupMessage` to `FollowupMessage`:
You can not only get public followups but also ephemeral ones, this isn't documented yet though (it was for a short time but got reverted). Instead of hard coding the visibility of the followup message for these methods decide on the actual type by the flags of the message. This is safe for any changes Discord decides to make.
* rename message type `ApplicationCommand` to `ChatInputCommand` (changed in https://github.com/discord/discord-api-docs/commit/3973676f437a21714d15338ccab9585c2492847f)
* change `FollowupMessageCreateBuilder.ephemeral` from `var` to `val` to prevent something like this:
```kotlin
val followup: PublicFollowupMessage = interactionResponse.followUpPublic {
    content = "foo"
    ephemeral = true // this will result in an ephemeral followup but the function still returns PublicFollowupMessage
}
followup.delete() // this will fail, you can't delete ephemeral followups
```

### Additions
* add `flags` property to `Message`
* add `MessageFlag.FailedToMentionSomeRolesInThread` (introduced in https://github.com/discord/discord-api-docs/commit/2d13f670b988da3be4e7aea98c74496701be3762)
* add `MessageType.ContextMenuCommand` (introduced in https://github.com/discord/discord-api-docs/commit/3973676f437a21714d15338ccab9585c2492847f)


Lukellmann:
Let's see what happens to this: https://github.com/discord/discord-api-docs/pull/4496

Lukellmann:
> Let's see what happens to this: [discord/discord-api-docs#4496](https://github.com/discord/discord-api-docs/pull/4496)

Ok, seems like getting ephemeral followup messages is not official (yet?) and subject to change.

I would still keep the type change from `PublicFollowupMessage` to `FollowupMessage` for these functions. If we receive a message with the ephemeral flag set it is ephemeral and not public.

Lukellmann:
waiting for #531

Lukellmann:
This is now ready, I merged the interaction refactor into this.

HopeBaron:
does getting them means we can delete?

Lukellmann:
> does getting them means we can delete?

Nope, that did not work the last time I checked. So we still need the type distinction between public/ephemeral. The user can just check with `is PublicFollowupMessage` and then delete.

core/src/main/kotlin/supplier/RestEntitySupplier.kt
```diff
@@ -440,11 +440,11 @@ public class RestEntitySupplier(public val kord: Kord) : EntitySupplier {
         applicationId: Snowflake,
         interactionToken: String,
         messageId: Snowflake,
-    ): PublicFollowupMessage? = catchNotFound {
```
HopeBaron, line 443: Why is it that the types here changed

Lukellmann: When the endpoint returns a message with the ephemeral flag it shouldn't be a `PublicFollowupMessage`.

HopeBaron: so we are going to use flags to determine ephemeral and public?

Lukellmann: That's exactly what the `FollowupMessage()` function does. The `edit` extension function for `FollowupMessageBehavior` already did this before.

Lukellmann: Or is there a downside to this that I'm missing?

HopeBaron: well the FollowupMessage is not a concrete types so the people may need to cast; maybe provide convenience methods as well.

Lukellmann: What would they look like? `asPublicFollowupMessage()`? And would they return `null` or throw on wrong type?

HopeBaron: no, make two distinct getters without this `asX` pattern here; because you can then call this function on a ephemeral function and make it public? is this of a value to us?

Lukellmann: Can you give me an example?

HopeBaron: Yeah that's what I was trying to imply; have there been any talks about supporting it? (deleting an ephemeral)

Lukellmann: This comment did somewhat imply it: https://github.com/discord/discord-api-docs/pull/3866#issuecomment-930408139

HopeBaron: Good news; then this issue is resolved.

HopeBaron: is it worth merging knowing that this will change?
I think you can just remove the behavioral distinction rather than re adjusting later

HopeBaron: Not entirely; the type distinction between the Ephemeral and Public is still required; but the behavior will be the same

HopeBaron: you know Discord have the ephmeral flag; we don't usually compare by types and flags for entites 
e.g DmChannel and TextChannel; they have their own types
Ephemeral and Public are parts of the concepts of interactions; also it's nice to express the distinction as an `is` rather than ==; it reads better and being a firm type makes it good.

On the other hand; your PR seems to have changed some of the parts that are going to change in V10 (which are not that hard to migrate to  https://github.com/discord/discord-api-docs/discussions/4510

not sure abotu the attachment patch thing yet tho

Lukellmann: Ok makes sense to have separate types for this.

So the thing you suggest now is to migrate all methods on followup messages to `FollowupMessageBehavior` (and including a warning that it might not work for ephemerals)?


> On the other hand; your PR seems to have changed some of the parts that are going to change in V10 (which are not that hard to migrate to https://github.com/discord/discord-api-docs/discussions/4510

Not sure which changes exactly you are referring to here.

HopeBaron: > [ALL VERSIONS] application.summary now returns an empty string. This field will be removed in v11

although we need a separate PR for this

Lukellmann: Oh, this could be in scope for https://github.com/kordlib/kord/pull/530

Lukellmann:
Discord removed the ability to edit ephemeral followups two days ago (see https://github.com/discord/discord-api-docs/issues/4414). Should we remove this function for ephemeral followups then?

Lukellmann:
> Discord removed the ability to edit ephemeral followups two days ago (see [discord/discord-api-docs#4414](https://github.com/discord/discord-api-docs/issues/4414)). Should we remove this function for ephemeral followups then?

Seems like they reverted that, it's working again.

HopeBaron:
Any official word on it?

Lukellmann:
> Any official word on it?

I've taken this from @DRSchlaubi (https://discord.com/channels/556525343595298817/631147109311053844/949635018844692530), nothing official.

Lukellmann:
I also found this message: https://discord.com/channels/613425648685547541/788586647142793246/947415392530337812

HopeBaron:
Since there is no change on this yet ig we can reintroduce them; this is kinda annoying

Lukellmann:
I didn't even remove the ability to edit ephemeral followups in the first place since it wasn't clear what would happen next, so we are all set in this regard.

PR #544, opened 2022-02-23
Update permission type

Lukellmann:
* rename `ManageEmojis` to `ManageEmojisAndStickers` (renamed in commit https://github.com/discord/discord-api-docs/pull/3128/commits/db5a0ccf82bf1c93c38b97b638fccf8e9df17474 of PR https://github.com/discord/discord-api-docs/pull/3128)
* rename `UseSlashCommands` to `UseApplicationCommands` (renamed in commit https://github.com/discord/discord-api-docs/pull/3614/commits/a6ca5e97dc70fcfcc789084673244f980fb0aaec of PR https://github.com/discord/discord-api-docs/pull/3614)
* add `UseExternalStickers` (added in commit https://github.com/discord/discord-api-docs/pull/3128/commits/03626e1e75e1aecc8580339bff4bd9c9023bf23b of PR https://github.com/discord/discord-api-docs/pull/3128)
* add `UseEmbeddedActivities` (renamed in https://github.com/discord/discord-api-docs/pull/4555)
* fix `Stream` missing in `Permission.values`
* use shifts instead of hex constants
* add docs

I recommend a commit-wise review.


Lukellmann:
Why is `Permission` no `enum class` like other flags? Should we change this?

HopeBaron:
Because of the Unknown class; in case Discord provided a permission that the user know but we don't

Lukellmann:
> Because of the Unknown class; in case Discord provided a permission that the user know but we don't

There is no Unknown class here, it would also be more difficult to use then an Unknown class for some sort of type.

HopeBaron:
It should have an unknown class; in case we encounter a permission we don't know

HopeBaron:
I think we had a compiler bug with enums so we decided to use this approach

Lukellmann:
Should I add a `Unknown` class then? This would mean we also have to change `Permissions.values`, it currently works by filtering the known values.

HopeBaron:
the unknown class won't be in the set anyways; it's unknown but it may provide the user with insight given that we hold the full permission set so they can interact with them.

Lukellmann:
> the unknown class won't be in the set anyways;

I don't mean `Permission.values`, I mean `Permissions.values` (the values of the combined flags).

Lukellmann:
I added a `Unknown` class.
Also no longer a draft, https://github.com/discord/discord-api-docs/pull/4555/files got merged.

PR #550, opened 2022-03-03
Switch to v10 for rest

HopeBaron:
https://github.com/discord/discord-api-docs/discussions/4510
as per this discussion; most of the changes have already been done I just added the intent and switched the version.


gateway/src/main/kotlin/Intent.kt
```diff
@@ -151,6 +151,9 @@ public sealed class Intent(public val code: DiscordBitSet) {
      */
     public object DirectMessageTyping : Intent(1 shl 14)
 
+    @PrivilegedIntent
+    public object MessageContent : Intent(1 shl 15)
+
```
Lukellmann, line 156: Also update the `values` property in the companion object, it seems to miss `GuildScheduledEvents` too (same order as the declaration would also be nice).

`Intents.nonPrivileged` should now also remove `MessageContent`, I suggest adding `Intents.privileged` and building `Intents.nonPrivileged` with `+Intents.all; -Intents.privileged`

Lukellmann (changes requested):

gateway/src/main/kotlin/Intent.kt
Lukellmann: I've taken the liberty to do this myself.

Lukellmann (approved):

PR #551, opened 2022-03-03
allow changing baseUrl and intercepting builder before handling

HopeBaron:
* Allow specifying baseUrls for routes other than Discord

### DIFF
--- a/core/src/test/kotlin/regression/CacheMissRegression.kt
+++ b/core/src/test/kotlin/regression/CacheMissRegression.kt
@@ -81,7 +81,7 @@ object FakeGateway : Gateway {
 }
 
 class CrashingHandler(val client: HttpClient, override val token: String) : RequestHandler {
-    override suspend fun <B : Any, R> handle(request: Request<B, R>): R {
+    override suspend fun <B : Any, R> handle(request: Request<B, R>, context: Any?): R {
         if (request.route != Route.CurrentUserGet) throw IllegalStateException("shouldn't do a request")
         val response = client.request<HttpStatement> {
             method = request.route.method
--- a/rest/src/main/kotlin/request/KtorRequestHandler.kt
+++ b/rest/src/main/kotlin/request/KtorRequestHandler.kt
@@ -2,6 +2,7 @@ package dev.kord.rest.request
 
 import dev.kord.rest.json.response.DiscordErrorResponse
 import dev.kord.rest.ratelimit.*
+import dev.kord.rest.route.Route
 import dev.kord.rest.route.optional
 import io.ktor.client.*
 import io.ktor.client.engine.cio.*
@@ -39,7 +40,9 @@ public class KtorRequestHandler(
     override val token: String
 ) : RequestHandler {
     private val logger = KotlinLogging.logger("[R]:[KTOR]:[${requestRateLimiter.javaClass.simpleName}]")
-
+    override suspend fun <T> intercept(builder: RequestBuilder<T>): RequestBuilder<T> {
+        return builder.defaultInterception(builder.route, token)
+    }
     override tailrec suspend fun <B : Any, R> handle(request: Request<B, R>): R {
         val response = requestRateLimiter.consume(request) {
             val httpRequest = client.createRequest(request)
@@ -76,7 +79,7 @@ public class KtorRequestHandler(
         headers.appendAll(request.headers)
 
         url {
-            url.takeFrom(dev.kord.rest.route.Route.baseUrl)
+            url.takeFrom(request.baseUrl)
             encodedPath += request.path
             parameters.appendAll(request.parameters)
         }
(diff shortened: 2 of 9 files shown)


rest/src/main/kotlin/request/RequestHandler.kt
```diff
@@ -16,4 +20,13 @@ public interface RequestHandler {
...
+
+    public suspend fun <T> intercept(builder: RequestBuilder<T>): RequestBuilder<T>
+}
+
+public fun  <T> RequestBuilder<T>.defaultInterception(route: Route<T>, token: String): RequestBuilder<T> {
+        if (route.requiresAuthorizationHeader) {
+            unencodedHeader(HttpHeaders.Authorization, "Bot $token")
+        }
+    return this
 }
```
Lukellmann, line 32:
```suggestion
    public suspend fun <T> RequestBuilder<T>.intercept(): RequestBuilder<T> {
        defaultInterception(token)
    }
}

public fun  <T> RequestBuilder<T>.defaultInterception(token: String): RequestBuilder<T> = apply {
        if (route.requiresAuthorizationHeader) {
            unencodedHeader(HttpHeaders.Authorization, "Bot $token")
        }
}
```
Why not use a member extension function?
We can provide a default, `RequestBuilder` also has access to the route.

rest/src/main/kotlin/request/KtorRequestHandler.kt
```diff
@@ -39,7 +40,9 @@ public class KtorRequestHandler(
     override val token: String
 ) : RequestHandler {
     private val logger = KotlinLogging.logger("[R]:[KTOR]:[${requestRateLimiter.javaClass.simpleName}]")
-
+    override suspend fun <T> intercept(builder: RequestBuilder<T>): RequestBuilder<T> {
+        return builder.defaultInterception(builder.route, token)
+    }
```
Lukellmann, line 45:
```suggestion

```
not needed if we have a default implementation

Lukellmann (changes requested):

rest/src/main/kotlin/service/RestService.kt
```diff
@@ -10,18 +10,17 @@ import kotlin.contracts.contract
...
             .apply(builder)
-            .apply {
-                if (route.requiresAuthorizationHeader) {
-                    unencodedHeader(Authorization, "Bot ${requestHandler.token}")
-                }
-            }
-            .build()
-        return requestHandler.handle(request)
+        val interceptedBuilder = requestHandler.intercept(requestBuilder)
+        return requestHandler.handle(interceptedBuilder.build())
```
Lukellmann, line 24:
```suggestion
        val interceptedBuilder = with(requestHandler) { requestBuilder.intercept() }
        return requestHandler.handle(interceptedBuilder.build())
```
also change this if `intercept` would get a member extension

core/src/test/kotlin/regression/CacheMissRegression.kt
```diff
@@ -81,7 +81,7 @@ object FakeGateway : Gateway {
 }
 
 class CrashingHandler(val client: HttpClient, override val token: String) : RequestHandler {
-    override suspend fun <B : Any, R> handle(request: Request<B, R>): R {
+    override suspend fun <B : Any, R> handle(request: Request<B, R>, context: Any?): R {
```
Lukellmann, line 84: the huge problem with this context parameter is that we have to propagate it through all the way (from core to here)

rest/src/main/kotlin/request/RequestHandler.kt
```diff
@@ -15,5 +20,17 @@ public interface RequestHandler {
...
+        return builder
+    }
+}
+
+public fun  <T> RequestBuilder<T>.defaultInterception(route: Route<T>, token: String): RequestBuilder<T> {
+        if (route.requiresAuthorizationHeader) {
+            unencodedHeader(HttpHeaders.UserAgent, KordConstants.USER_AGENT)
+            unencodedHeader(HttpHeaders.Authorization, "Bot $token")
+        }
+    return this
```
Lukellmann, line 35:
```suggestion
    public suspend fun <T> intercept(builder: RequestBuilder<T>, context: Any? = null): RequestBuilder<T> =
        builder.defaultInterception(token)
}

public fun  <T> RequestBuilder<T>.defaultInterception(token: String): RequestBuilder<T> = apply {
    unencodedHeader(HttpHeaders.UserAgent, KordConstants.USER_AGENT)
    if (route.requiresAuthorizationHeader) {
        unencodedHeader(HttpHeaders.Authorization, "Bot $token")
    }
```
default interception in `RequestHandler.intercept` default implementation, always provide user agent, use `route` from `RequestBuilder`

rest/src/main/kotlin/service/RestService.kt
```diff
@@ -12,19 +12,17 @@ import kotlin.contracts.contract
 public abstract class RestService(@PublishedApi internal val requestHandler: RequestHandler) {
 
     @PublishedApi
-    internal suspend inline fun <T> call(route: Route<T>, builder: RequestBuilder<T>.() -> Unit = {}): T {
+    internal suspend inline fun <T> call(
+        route: Route<T>,
+        baseUrl: String = Route.baseUrl,
+        builder: RequestBuilder<T>.() -> Unit = {},
+    ): T {
```
Lukellmann, line 19:
```suggestion
   internal suspend inline fun <T> call(route: Route<T>, builder: RequestBuilder<T>.() -> Unit = {}): T {
```
we don't change the base url internally, this can be done with `RequestHandler.intercept`

Lukellmann (changes requested):

PR #552, opened 2022-03-05
fix small mistake in KordRestOnlyBuilder.kt

WinteryFox:
Fixes the chaining for one of the builder methods of KordRestOnlyBuilder

### DIFF
--- a/core/src/main/kotlin/builder/kord/KordRestOnlyBuilder.kt
+++ b/core/src/main/kotlin/builder/kord/KordRestOnlyBuilder.kt
@@ -47,7 +47,7 @@ public class KordRestOnlyBuilder(public val token: String) {
      * }
      * ```
      */
-    public fun requestHandler(handlerBuilder: (resources: ClientResources) -> RequestHandler) {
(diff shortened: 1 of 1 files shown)


Lukellmann:
Kord's builders are normally used in a Kotlin DSL style, they don't support the java-like builder chaining.
What's your usecase?

WinteryFox:
Really? This builder doesn't seem to accept DSL. There's no secondary constructor that accepts DSL either.
`KordRestOnlyBuilder("token").requestHandler { something }.build()`

Lukellmann:
It' supposed to be used this way:
```kotlin
val kord = Kord.restOnly {
    // ...
}
```

WinteryFox:
Ohh that makes more sense, should probably change the visibility of this builder then?

Lukellmann:
I don't see a problem with builders and their constructors being public.

WinteryFox:
Makes sense, it would be nice to leave a note with a reference to the correct usage though.

Lukellmann:
> Makes sense, it would be nice to leave a note with a reference to the correct usage though.

Would be nice if you were able to do this since you already opened this PR 😄

Lukellmann:
Thanks :)

PR #553, opened 2022-03-06
Update `enableEvent()`

Lukellmann:
Updates the extension function `enableEvents()` for `Intents.IntentsBuilder` to support all new events that were added to Kord to match [this list](https://discord.com/developers/docs/topics/gateway#list-of-intents).

This also includes little fixes:
- the `DirectMessages` intent will no longer be added to events that only need the `Guilds` intent
- `TypingStartEvent` will now add both `GuildMessageTyping` and `DirectMessageTyping`

Other changes:
- add `LoginBuilder.intents()` convenience function

### DIFF
--- a/core/src/main/kotlin/Util.kt
+++ b/core/src/main/kotlin/Util.kt
@@ -1,11 +1,13 @@
 package dev.kord.core
 
+import dev.kord.common.annotation.KordPreview
 import dev.kord.common.entity.Snowflake
 import dev.kord.core.entity.Entity
 import dev.kord.core.entity.KordEntity
 import dev.kord.core.entity.channel.thread.ThreadChannel
 import dev.kord.core.event.Event
 import dev.kord.core.event.channel.*
+import dev.kord.core.event.channel.thread.*
 import dev.kord.core.event.guild.*
 import dev.kord.core.event.message.*
 import dev.kord.core.event.role.RoleCreateEvent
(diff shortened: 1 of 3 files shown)


Lukellmann:
While working on this I also noticed that some events are missing in Kord, namely:
```
STAGE_INSTANCE_CREATE
STAGE_INSTANCE_UPDATE
STAGE_INSTANCE_DELETE

GUILD_STICKERS_UPDATE

INTEGRATION_CREATE
INTEGRATION_UPDATE
INTEGRATION_DELETE
```
We can address this in a separate PR.

core/src/main/kotlin/Util.kt
```diff
@@ -317,48 +331,105 @@ public fun Intents.IntentsBuilder.enableEvent(event: KClass<out Event>): Unit =
...
+    NewsChannelThreadCreateEvent::class,
+    TextChannelThreadCreateEvent::class,
+    UnknownChannelThreadCreateEvent::class,
+
+    ThreadUpdateEvent::class,
+    NewsChannelThreadUpdateEvent::class,
+    TextChannelThreadUpdateEvent::class,
+    UnknownChannelThreadUpdateEvent::class,
+
+    ThreadChannelDeleteEvent::class,
```
HopeBaron, line 370: ThreadChannelDeleteEvent is enough here, the other two are sub types

core/src/main/kotlin/Util.kt
```diff
@@ -317,48 +331,105 @@ public fun Intents.IntentsBuilder.enableEvent(event: KClass<out Event>): Unit =
...
     TextChannelDeleteEvent::class,
+    UnknownChannelDeleteEvent::class,
     VoiceChannelDeleteEvent::class,
 
-    ChannelPinsUpdateEvent::class,
-    -> {
-        +Guilds
-        +DirectMessages
-    }
+    ThreadChannelCreateEvent::class,
```
HopeBaron, line 360: Same here.

core/src/main/kotlin/Util.kt
```diff
@@ -317,48 +331,105 @@ public fun Intents.IntentsBuilder.enableEvent(event: KClass<out Event>): Unit =
...
+    NewsChannelThreadCreateEvent::class,
+    TextChannelThreadCreateEvent::class,
+    UnknownChannelThreadCreateEvent::class,
+
+    ThreadUpdateEvent::class,
```
HopeBaron, line 365: Same here

HopeBaron (changes requested):

core/src/main/kotlin/Util.kt
Lukellmann: This does not work, `when` uses `equals`, from `KClass.equals`:
> Returns true if this KClass instance represents the same Kotlin class as the class represented by other. On JVM this means that all of the following conditions are satisfied:
> 1. other has the same (fully qualified) Kotlin class name as this instance.
> 2. other's backing Class object is loaded with the same class loader as the Class object of this instance.
>
> [...]

Also see this Kotlin Playground: https://pl.kotl.in/H1e13i-bw

PR #554, opened 2022-03-08
Use gateway v10

Lukellmann:
The default rest version was bumped to 10 but we missed to also bump it for gateway.

`MessageCreateEvent` and `MessageUpdateEvent` now also add the `MessageContent` intent in `enableEvent()`.


Lukellmann:
Should we maybe put these versions into one place somehow? So we won't miss this again in the future.

HopeBaron:
I was thinking the same thing. we will look into this in a separate PR

PR #558, opened 2022-03-09
User Agent header

Lukellmann:
Discord's documentation states that "clients using the HTTP API must provide a valid User Agent" (https://discord.com/developers/docs/reference#user-agent) but Kord currently does not provide this header.

This PR adds this header.

### DIFF
--- a/common/src/main/kotlin/KordConstants.kt
+++ b/common/src/main/kotlin/KordConstants.kt
@@ -3,5 +3,5 @@ package dev.kord.common
 public object KordConstants {
 
     /** Kord's value for the [User Agent header](https://discord.com/developers/docs/reference#user-agent). */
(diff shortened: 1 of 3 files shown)


Lukellmann:
I left this as a draft for now as I see some open questions and problems with the current form of implementing it.
1. Hardcoding the version will require to change it for every release and does not support snapshot versions, is there a way to inject that at the build time of the artifacts?
2. Defining the User Agent value as a constant does not allow for some other library / bot to override this, is this something we should alllow?
3. Should we provide https://kord.dev or https://github.com/kordlib/kord for `$url`?

Lukellmann:
Which spelling should the user agent constant have? `UserAgent`, `userAgent` or `USER_AGENT`?
Kord isn't very consistent with this.

PR #559, opened 2022-03-09
Audit Log Change Key "type" can have integer or string values

Lukellmann:
In the case of some audit log events (e.g. `action_type` 80: `INTEGRATION_CREATE`) Audit Log Change for key `type` is of type string, in some other audit log events (e.g. `action_type` 13: `CHANNEL_OVERWRITE_CREATE`) Audit Log Change for key `type` is of type integer (see the [list of Audit Log Change Keys](https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-audit-log-change-key) for key name `type`).

This is the only Audit Log Change Key that can have different types for its value.

### DIFF
--- a/common/src/main/kotlin/entity/AuditLog.kt
+++ b/common/src/main/kotlin/entity/AuditLog.kt
@@ -8,8 +8,7 @@ import kotlinx.serialization.*
 import kotlinx.serialization.builtins.serializer
 import kotlinx.serialization.descriptors.*
 import kotlinx.serialization.encoding.*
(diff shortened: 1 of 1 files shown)


HopeBaron:
Making it an Any doesn't really provide any value :thinking:

Lukellmann:
I see three other options:
- some `IntOrString` sealed class
- always converting it to a String
- making it `JsonPrimitive`

HopeBaron:
Make it always a string

PR #563, opened 2022-03-10
Add strategy to mitigate loss of stacktrace through Ktors SuspendFunctionGun

DRSchlaubi:
Ktors SuspendFunctionGun often removes any useful information from a stacktrace

This PR tries to mitigate that by capturing the Stacktrace before Ktor gets called
AS discussed with @Lukellmann this is opt-in﻿

### DIFF
--- a/core/src/main/kotlin/builder/kord/KordBuilder.kt
+++ b/core/src/main/kotlin/builder/kord/KordBuilder.kt
@@ -75,6 +75,16 @@ public class KordBuilder(public val token: String) {
     private var cacheBuilder: KordCacheBuilder.(resources: ClientResources) -> Unit = {}
 
     /**
+     * Enables stack trace recovery on the currently defined [RequestHandler].
+     *
+     * @throws IllegalStateException if the [RequestHandler] is not a [KtorRequestHandler]
+     *
+     * @see StackTraceRecoveringKtorRequestHandler
+     * @see withStackTraceRecovery
+     */
+    private var stackTraceRecovery: Boolean = false
+
+    /**
      * Enable adding a [Runtime.addShutdownHook] to log out of the [Gateway] when the process is killed.
      */
     public var enableShutdownHook: Boolean = true
@@ -158,23 +168,6 @@ public class KordBuilder(public val token: String) {
     }
 
     /**
-     * Enables stack trace recovery on the currently defined [RequestHandler].
-     *
-     * @throws IllegalStateException if the [RequestHandler] is not a [KtorRequestHandler]
-     *
-     * @see StackTraceRecoveringKtorRequestHandler
-     * @see withStackTraceRecovery
-     */
-    public fun enableStackTraceRecovery() {
-        val parentBuilder = handlerBuilder
-        requestHandler {
-            val ktorRequestHandler = parentBuilder(it) as? KtorRequestHandler
-                ?: error("Stack trace recovery only works with KtorRequestHandlers")
-            ktorRequestHandler.withStackTraceRecovery()
-        }
-    }
-
-    /**
      * Configures the [DataCache] for caching.
      *
      *  ```
@@ -245,7 +238,17 @@ public class KordBuilder(public val token: String) {
 
         val resources =
             ClientResources(token, applicationId ?: getBotIdFromToken(token), shardsInfo, client, defaultStrategy)
-        val rest = RestClient(handlerBuilder(resources))
+        val rawRequestHandler = handlerBuilder(resources)
+        val requestHandler = if (stackTraceRecovery) {
+            if (rawRequestHandler is KtorRequestHandler) {
+                rawRequestHandler.withStackTraceRecovery()
+            } else {
+                error("stackTraceRecovery only works with KtorRequestHandlers, please set stackTraceRecovery = false or use a different RequestHandler")
+            }
+        } else {
+            rawRequestHandler
+        }
+        val rest = RestClient(requestHandler)
         val cache = KordCacheBuilder().apply { cacheBuilder(resources) }.build()
         cache.registerKordData()
         val gateway = run {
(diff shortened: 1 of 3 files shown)


BartArys (commented):
As I mentioned, masking the Throwable type with our own is problematic for error handling logic both in our code and userland.

Editing the cause of a caught exceptions, or editing its stacktrace would be problematic as well. 

Considering we documented `RequestHandler#handle` to only throw `RestRequestException`, it would be prudent to at least make our new exception an implementation of `RestRequestException` and throw that when we intercept a downstream `RestRequestException`. Perhaps going as far as creating our own version of `KtorRequestException` as well.

Other Throwables, and Errors, are undocumented behaviour. Users technically shouldn't expect these to be thrown, so I suppose it's okay if we type mask those.

Additionally, this is mostly a band-aid solution. Either until we better understand how to recover  Ktor stacktraces, or until the Ktor team better understands stacktrace recovery themselves. As such, it's probably better to mark the relevant code as `KordExperimental`.

rest/src/main/kotlin/request/StackTraceRecoveringKtorRequestHandler.kt
```diff
@@ -0,0 +1,58 @@
...
+
+        return try {
+            delegate.handle(request)
+        } catch (e: Throwable) {
+            throw stacktrace.apply {
+                sanitizeStackTrace()
+                cause = e
+            }
+        }
+    }
```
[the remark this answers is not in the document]
DRSchlaubi: We could try just re-throwing the existing exception and adding our context exception as the cause so the threw exc eption stays the same

rest/src/main/kotlin/request/StackTraceRecoveringKtorRequestHandler.kt
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Exception used to save the current stack trace before executing a request.
+ *
+ * @see StackTraceRecoveringKtorRequestHandler
+ */
+public class ContextException internal constructor() : RuntimeException() {
```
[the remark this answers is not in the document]
DRSchlaubi: I wasn't sure whether there was a valid use case to use that exception as a user or not

core/src/main/kotlin/builder/kord/KordBuilder.kt
```diff
@@ -159,6 +157,23 @@ public class KordBuilder(public val token: String) {
...
+     * @see withStackTraceRecovery
+     */
+    public fun enableStackTraceRecovery() {
+        val parentBuilder = handlerBuilder
+        requestHandler {
+            val ktorRequestHandler = parentBuilder(it) as? KtorRequestHandler
+                ?: error("Stack trace recovery only works with KtorRequestHandlers")
+            ktorRequestHandler.withStackTraceRecovery()
+        }
+    }
```
Lukellmann, line 175: I think this should set some sort of flag to use recovery that is then used at build time to replace the handler instead of replacing the `handlerBuilder`, the following would not work as intended because `requestHandler { }` replaces the recovering handler:
```kotlin
val kord = Kord("") {
    enableStackTraceRecovery()
    requestHandler { KtorRequestHandler(...) }
}
```
It could also just be a Boolean property in that case:
```kotlin
val kord = Kord("") {
    stackTraceRecovery = true
}
```

Apart from that: you only added it to `KordBuilder` and not `KordRestOnlyBuilder`.

rest/src/main/kotlin/request/StackTraceRecoveringKtorRequestHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    internal fun sanitizeStackTrace() {
+        // Remove artifacts of stack trace capturing
+        // at dev.kord.rest.request.StackTraceRecoveringKtorRequestHandler.handle(StackTraceRecoveringKtorRequestHandler.kt:7)
```
Lukellmann, line 45: this line number is already out of date

Lukellmann (changes requested):

core/src/main/kotlin/builder/kord/KordBuilder.kt
DRSchlaubi: With that flag, there would be no validation whether it actually does something or not, since it only works with the KtorRequestHandler

Lukellmann: If you put the logic inside `requestHandler { }` in `enableStackTraceRecovery` into the build function behind an if check there would.

DRSchlaubi: requestHandler currently doesn't have a builder it's just a function returning a RequestHandler, so that would require changing a lot of things

Lukellmann: Let me clarify what I mean:
```kotlin
class KordBuilder {

    var recoverStacktraces = false

    // ...

    fun build(): Kord {
        // ...
        val resources = ...
        var handler = handlerBuilder(resources)
        if (recoverStacktraces) {
            require(handler is KtorRequestHandler)
            handler = handler.withStackTraceRecovery
        }
        // ...
    }
}
```

rest/src/main/kotlin/request/StackTraceRecoveringKtorRequestHandler.kt
DRSchlaubi: Shouldn't have written docs ;)

Lukellmann: exactly :D

core/src/main/kotlin/builder/kord/KordBuilder.kt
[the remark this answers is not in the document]
DRSchlaubi: the new approach should also fix this

rest/src/main/kotlin/request/StackTraceRecoveringKtorRequestHandler.kt
Lukellmann: no but it could be replaced with `<linenumber>` or smth

PR #569, opened 2022-03-16
Fix Guild Scheduled Event Gateway Events

Lukellmann:
- fix deserialization of `GuildScheduledEventUserAdd` and `GuildScheduledEventUserRemove` in gateway module (deserialize object with 3 snowflakes instead of 3 flat snowflakes)
- actually dispatch `GuildScheduledEventUserAddEvent` and `GuildScheduledEventUserRemoveEvent` in core module
- add `getMember` and `getMemberOrNull` to `GuildScheduledEventUserEvent`
- fix return types of `withStrategy`

### DIFF
--- a/gateway/src/main/kotlin/Event.kt
+++ b/gateway/src/main/kotlin/Event.kt
@@ -427,15 +427,19 @@ public sealed class Event {
                     sequence
                 )
                 "GUILD_SCHEDULED_EVENT_USER_ADD" -> GuildScheduledEventUserAdd(
-                    decoder.decodeSerializableElement(descriptor, index, Snowflake.serializer()),
-                    decoder.decodeSerializableElement(descriptor, index, Snowflake.serializer()),
-                    decoder.decodeSerializableElement(descriptor, index, Snowflake.serializer()),
+                    data = decoder.decodeSerializableElement(
+                        descriptor,
+                        index,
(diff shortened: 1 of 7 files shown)


gateway/src/main/kotlin/Event.kt
```diff
@@ -734,19 +738,31 @@ public data class GuildScheduledEventUpdate(val event: DiscordGuildScheduledEven
...
+    @Serializable
+    public data class Data(
+        @SerialName("guild_scheduled_event_id")
+        val guildScheduledEventId: Snowflake,
+        @SerialName("user_id")
+        val userId: Snowflake,
+        @SerialName("guild_id")
+        val guildId: Snowflake,
+    )
+}
```
Lukellmann, line 752: I'm not sure about this nested `data class Data`, it's duplicated in `GuildScheduledEventUserRemove`, can someone think of a better name to share this between both?

HopeBaron: sounds like a ScheduledEventMetaData or (data)?

HopeBaron (approved):

gateway/src/main/kotlin/Event.kt
Lukellmann: renamed it to `GuildScheduledEventUserMetadata`, it's also shared now

PR #572, opened 2022-03-17
Add minValue and maxValue to ApplicationCommandOption

### LINKED ISSUE #556: Missing minValue and maxValue fields in ApplicationCommandOption
`ApplicationCommandOption` is missing the properties `minValue` and `maxValues` used for `INTEGER` and `NUMBER` command options:

https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure

DRSchlaubi:
Fix #556﻿

### DIFF
--- a/core/src/main/kotlin/behavior/interaction/AutoCompleteInteractionBehavior.kt
+++ b/core/src/main/kotlin/behavior/interaction/AutoCompleteInteractionBehavior.kt
@@ -4,8 +4,8 @@ import dev.kord.common.entity.Choice
 import dev.kord.common.entity.DiscordAutoComplete
 import dev.kord.core.entity.interaction.AutoCompleteInteraction
 import dev.kord.core.supplier.EntitySupplyStrategy
-import dev.kord.rest.builder.interaction.IntChoiceBuilder
-import dev.kord.rest.builder.interaction.NumberChoiceBuilder
+import dev.kord.rest.builder.interaction.IntegerOptionBuilder
+import dev.kord.rest.builder.interaction.NumberOptionBuilder
 import dev.kord.rest.builder.interaction.StringChoiceBuilder
 import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
@@ -28,9 +28,9 @@ public interface AutoCompleteInteractionBehavior : DataInteractionBehavior {
  *
  * The provided choices are only suggestions and the user can provide any other input as well.
  *
- * @see IntChoiceBuilder
+ * @see IntegerOptionBuilder
  */
-public suspend inline fun AutoCompleteInteractionBehavior.suggestInt(builder: IntChoiceBuilder.() -> Unit) {
+public suspend inline fun AutoCompleteInteractionBehavior.suggestInt(builder: IntegerOptionBuilder.() -> Unit) {
     contract {
         callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
     }
@@ -43,9 +43,9 @@ public suspend inline fun AutoCompleteInteractionBehavior.suggestInt(builder: In
  *
  * The provided choices are only suggestions and the user can provide any other input as well.
  *
- * @see NumberChoiceBuilder
+ * @see NumberOptionBuilder
  */
-public suspend inline fun AutoCompleteInteractionBehavior.suggestNumber(builder: NumberChoiceBuilder.() -> Unit) {
+public suspend inline fun AutoCompleteInteractionBehavior.suggestNumber(builder: NumberOptionBuilder.() -> Unit) {
     contract {
         callsInPlace(builder, InvocationKind.EXACTLY_ONCE)
     }
--- a/rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
+++ b/rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
@@ -111,11 +111,11 @@ public sealed class NumericOptionBuilder<T : Number>(
 }
 
 
-@Deprecated("Replaced by IntOptionBuilder", ReplaceWith("IntOptionBuilder"), DeprecationLevel.ERROR)
-public typealias IntChoiceBuilder = IntOptionBuilder
+@Deprecated("Replaced by IntegerOptionBuilder", ReplaceWith("IntegerOptionBuilder"), DeprecationLevel.ERROR)
+public typealias IntChoiceBuilder = IntegerOptionBuilder
 
 @KordDsl
-public class IntOptionBuilder(name: String, description: String) :
+public class IntegerOptionBuilder(name: String, description: String) :
     NumericOptionBuilder<Long>(name, description, ApplicationCommandOptionType.Integer) {
 
     override fun choice(name: String, value: Long) {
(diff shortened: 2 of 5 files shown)


rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
```diff
@@ -71,20 +73,62 @@ public sealed class BaseChoiceBuilder<T>(
...
+        default = _default,
+        autocomplete = _autocomplete,
+        minValue = _minValue.map { JsonPrimitive(it) },
+        maxValue = _maxValue.map { JsonPrimitive(it) },
+    )
+}
+
+
+@Deprecated("Replaced by IntOptionBuilder", ReplaceWith("IntOptionBuilder"))
+public typealias IntChoiceBuilder = IntOptionBuilder
```
HopeBaron, line 114: I guess this will break bytecode either way.

rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
```diff
@@ -71,20 +73,62 @@ public sealed class BaseChoiceBuilder<T>(
...
+    NumericOptionBuilder<Long>(name, description, ApplicationCommandOptionType.Integer) {
 
     override fun choice(name: String, value: Long) {
         if (choices == null) choices = mutableListOf()
         choices!!.add(Choice.IntChoice(name, value))
     }
 }
 
+@Deprecated("Replaced by IntOptionBuilder", ReplaceWith("NumberOptionBuilder"))
+public typealias NumberChoiceBuilder = NumberOptionBuilder
```
HopeBaron, line 127: Same here

rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
```diff
@@ -71,20 +73,62 @@ public sealed class BaseChoiceBuilder<T>(
...
+
+
+@Deprecated("Replaced by IntOptionBuilder", ReplaceWith("IntOptionBuilder"))
+public typealias IntChoiceBuilder = IntOptionBuilder
+
 @KordDsl
-public class IntChoiceBuilder(name: String, description: String) :
-    BaseChoiceBuilder<Long>(name, description, ApplicationCommandOptionType.Integer) {
+public class IntOptionBuilder(name: String, description: String) :
+    NumericOptionBuilder<Long>(name, description, ApplicationCommandOptionType.Integer) {
```
Lukellmann, line 118: If you already rename this, I would suggest `IntegerOptionBuilder`, it makes the distinction from `Int` clearer, we already have `ApplicationCommandOptionType.Integer`, `IntegerArgument` and `IntegerOptionValue`.

rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
Lukellmann, line 127:
```suggestion
@Deprecated("Replaced by IntOptionBuilder", ReplaceWith("NumberOptionBuilder"), DeprecationLevel.ERROR)
public typealias NumberChoiceBuilder = NumberOptionBuilder
```
same here

Lukellmann (changes requested):

rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
DRSchlaubi: I was supposed to not break the source code

rest/src/main/kotlin/builder/interaction/OptionsBuilder.kt
```diff
@@ -55,7 +57,7 @@ public sealed class BaseChoiceBuilder<T>(
     // TODO We can change these types to Optional<MutableList<Choice<T>>> and MutableList<Choice<T>> once
     //  https://youtrack.jetbrains.com/issue/KT-51045 is fixed.
     //  The bug from that issue prevents you from setting BaseChoiceBuilder<*>.choices to `null`.
-    private var _choices: Optional<MutableList<Choice<*>>> = Optional.Missing()
+    internal var _choices: Optional<MutableList<Choice<*>>> = Optional.Missing()
```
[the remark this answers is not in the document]
DRSchlaubi: I actually needed to revert this since changing it to protected makes ::_choices be a KProperty0, rather than a KMutableProperty0

core/src/main/kotlin/behavior/interaction/AutoCompleteInteractionBehavior.kt
```diff
@@ -28,9 +28,9 @@ public interface AutoCompleteInteractionBehavior : DataInteractionBehavior {
  *
  * The provided choices are only suggestions and the user can provide any other input as well.
  *
- * @see IntChoiceBuilder
+ * @see IntegerOptionBuilder
  */
-public suspend inline fun AutoCompleteInteractionBehavior.suggestInt(builder: IntChoiceBuilder.() -> Unit) {
+public suspend inline fun AutoCompleteInteractionBehavior.suggestInt(builder: IntegerOptionBuilder.() -> Unit) {
```
Lukellmann, line 33: this will also allow specifying min and max values (but they aren't used so it's probably fine)

Lukellmann (approved):

core/src/main/kotlin/behavior/interaction/AutoCompleteInteractionBehavior.kt
HopeBaron: Don't they allow defining ranges?

DRSchlaubi: AutoComplete suggestions provide a list of selectable inputs, so there is no range i guess

HopeBaron: should their be a BaseChoiceBuilder type for this then?

DRSchlaubi: IntegerOptionBuilder is used anyways, the only way would be to make 2 IntegerOptionBuilder and 2 NumberOptionBuilder, which seems a bit unnecessary

HopeBaron: what if we introduce an interface like, `RangedNumericOptionBuilder` and extends them; that only contains the min and max values
we won't introduce a new Double and Int values; instead we just use the one without values (interface) on the autocomplete functions and apply it to the one with ranges (complete type)

DRSchlaubi: That interfae is exactly what the NumericOptionBuilder is right now

the issue is that the int option builder is the same as the int choice builder

HopeBaron: Hmmm, we could go with this for now? (current implementation)
so we would need an additonal interface?

DRSchlaubi: We would need an additional class the interfaces are fine, but I doubt this will be a huge issue for now

HopeBaron: You are right, we can figure this out on the go

Issue #69, opened 2020-09-21
Status of Mutliplatform


BartArys:
This issue will function as a overview of the things we'll require or *like to have* before we can start moving Kord to mulltiplatform:

## Need to have
We'll not consider supporting the following platforms unless the following points are solved
* js
  * [x] [a proper datetime library, preferably maintained by jetbrains](https://github.com/Kotlin/kotlinx-datetime)
* native
  * [x] [a proper datetime library, preferably maintained by jetbrains](https://github.com/Kotlin/kotlinx-datetime)
  * [ ] [ktor support for ~~sockets~~ websockets](https://github.com/ktorio/ktor/issues/1215)
  * [ ] [support for multithreaded coroutines](https://github.com/Kotlin/kotlinx.coroutines/issues/462)
  * [ ] [stable memory model for native](https://blog.jetbrains.com/kotlin/2020/07/kotlin-native-memory-management-roadmap/) 

## Nice to have
These are some nice-to-haves, we can work around them though
* [ ] Multiplatform cache or concurrent collections
* [ ] [IR support for all platforms](https://kotlinlang.org/docs/reference/whatsnew14.html#unified-backends-and-extensibility)

DRSchlaubi:
https://github.com/Kotlin/kotlinx.coroutines/issues/462 got closed, so we actually have multithreaded coroutines on native now

DRSchlaubi:
Web socket support on native seems to exist (although I couldn't really test that yeet) however you cannot use TLS connections on native

silenium-dev:
Regarding the stable memory model for native, the new memory manager is now stable and used by default (https://kotlinlang.org/docs/native-memory-manager.html).

Lukellmann:
It's technically in Beta but so are Kotlin/Native and Kotlin Multiplatform: https://kotlinlang.org/docs/components-stability.html#current-stability-of-kotlin-components

DRSchlaubi:
Currently, a WIP MPP version of kord is on the `feature/müü` branch. Is supports almost all features (except voice) on both JVM and JS

DRSchlaubi:
After some checking with @Lukellmann I found out that K/N is ready, I propose using the following engines:
- Linux: CIO
- Apple systems: Darwing
- Windows: WinHttp

Xinerfapre:
Is it possible to add [Kotlin/Wasm](https://kotl.in/wasm) compatibility ?

DRSchlaubi:
[...]

PR #575, opened 2022-03-26
Change "dev.kord.rest.version" to be retrieved via `getProperty`, not `getenv`

MrPowerGamerBR:
`dev.kord.rest.version` looks and sounds like a Java System Property... but it is retrieved via `getenv`.

However environment variables use `UPPERCASED_VARIABLES` as the convention, so I don't think this was meant to be retrieved via `getenv`.


Lukellmann:
I'm not sure we should change this now, as I plan to bring a replacement in the form of more _publicly_ available configuration options (this is internal anyways).

DRSchlaubi:
Considering #69 I don't think it's a good idea to add more dependencies to the JVM stdlib because getProperty is not available on other platforms.

I think we should wait for @Lukellmann's replacement

MrPowerGamerBR:
Superseded by https://github.com/kordlib/kord/pull/601

PR #576, opened 2022-03-27
Version constant injection

Lukellmann:
This PR changes the version used in the user agent header from being hard-coded to being generated at build time using [this gradle plugin](https://github.com/gmazzo/gradle-buildconfig-plugin).

The plugin will generate the internal top-level property `dev.kord.common.BUILD_CONFIG_GENERATED_LIBRARY_VERSION` and assign the value of [`Library.version`](https://github.com/kordlib/kord/blob/features/version-injection/buildSrc/src/main/kotlin/Projects.kt#L12) to it. `KordConstants.KORD_VERSION` then makes this publicly accessible.

### DIFF
--- a/buildSrc/src/main/kotlin/Projects.kt
+++ b/buildSrc/src/main/kotlin/Projects.kt
@@ -1,23 +1,7 @@
-import org.gradle.api.Project
-import java.io.ByteArrayOutputStream
-
 /**
  * whether the process has been invoked by JitPack
  */
 val isJitPack get() = "true" == System.getenv("JITPACK")
-
-val Project.commitHash: String
-    get() = try {
-        ByteArrayOutputStream().use { out ->
-            exec {
-                commandLine("git", "rev-parse", "--short", "HEAD")
-                standardOutput = out
-            }
-            out.toString().trim()
-        }
-    } catch (e: Throwable) {
-        System.getenv("GITHUB_SHA") ?: "unknown"
-    }
 
 object Library {
     const val name = "kord"
@@ -35,6 +19,11 @@ object Library {
             }
 
         }
+
+    val commitHash get() = System.getenv("GITHUB_SHA") ?: "unknown"
+
+    // this environment variable isn't available out of the box, we set it ourselves
+    val shortCommitHash get() = System.getenv("SHORT_SHA") ?: "unknown"
 
     const val description = "Idiomatic Kotlin Wrapper for The Discord API"
     const val projectUrl = "https://github.com/kordlib/kord"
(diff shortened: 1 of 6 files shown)


Lukellmann:
@DRSchlaubi do you think `common/build.gradle.kts` is the right place for the plugin and its config? Or should I put this somewhere else?

DRSchlaubi:
Ideally all versions would be in the root file (just with apply false), but we already gave up on that when we used the precomposed scripts. We could add it to one of the scripts, so we can add it to multiple modules, but we only need it in common so it's fine

DRSchlaubi (changes requested):
Would be nice to have a commit hash as well, as we use a lot of snapshots

Lukellmann:
> Would be nice to have a commit hash as well, as we use a lot of snapshots

What would the format look like?

DRSchlaubi:
> > Would be nice to have a commit hash as well, as we use a lot of snapshots
> 
> What would the format look like?

Simply make an additional constant?

And then maybe also include it with (Commit hash: <hash>) in the user agent

Lukellmann:
> Simply make an additional constant?

Ah, I thought you wanted it to be part of the KORD_VERSION string.

Lukellmann:
Like this? https://github.com/kordlib/kord/pull/576/commits/85270e9c1f59709535b9a2cf93ccb85214357fb9

I also can't come up with documentation for `KORD_COMMIT_HASH`, any suggestions?

DRSchlaubi (approved):

buildSrc/src/main/kotlin/Projects.kt
```diff
@@ -1,11 +1,24 @@
...
-import org.gradle.kotlin.dsl.project
+import org.gradle.api.Project
+import java.io.ByteArrayOutputStream
 
 /**
  * whether the process has been invoked by JitPack
  */
 val isJitPack get() = "true" == System.getenv("JITPACK")
 
+val Project.commitHash: String
```
HopeBaron, line 9: Github allows you to get the commit hash, why is this logic code introduced.

Lukellmann: You are referring to the `GITHUB_SHA` environment variable, right? This one is the full hash, @DRSchlaubi wanted to have the short one if available (retrieved by `git rev-parse --short HEAD`).

HopeBaron: Github doesn't provide that? in the environment?

Lukellmann: I didn't search too long, maybe they do. Do you know more?

DRSchlaubi: > Github doesn't provide that? in the environment?

https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables only lists the long SHA

Lukellmann: Maybe we should compute the short hash in the actions instead (like [here](https://github.community/t/github-context-with-short-sha/17426)).
It would probably also be good to provide `KORD_COMMIT_HASH` and `KORD_SHORT_COMMIT_HASH`.

PR #577, opened 2022-03-29
Refactor common `RateLimiter`

Lukellmann:
Right now the only implementation of `dev.kord.common.ratelimit.RateLimiter` is `BucketRateLimiter` that uses a `Clock` for measuring intervals based on `Instant`s.

This PR introduces two new concrete implementations, namely `ClockIntervalRateLimiter` and `TimeSourceIntervalRateLimiter` with `AbstractIntervalRateLimiter` as their abstract base class.

`ClockIntervalRateLimiter` is a direct replacement for `BucketRateLimiter`, it behaves the same (compare [`BucketRateLimiter`](https://github.com/Lukellmann/kord/blob/604524544cb27dfff619409a392c7825fe3647e3/common/src/main/kotlin/ratelimit/BucketRateLimiter.kt) with [`AbstractIntervalRateLimiter`](https://github.com/Lukellmann/kord/blob/af1ed915d83c77a4f7ee1d7836d2c13e2daa800e/common/src/main/kotlin/ratelimit/AbstractIntervalRateLimiter.kt) and [`ClockIntervalRateLimiter`](https://github.com/Lukellmann/kord/blob/ba43112f1782623912f1dc63797c38bbd48332c8/common/src/main/kotlin/ratelimit/ClockIntervalRateLimiter.kt)).

A new `IntervalRateLimiter` interface is also introduced, which has a more restricted contract for `consume` than `RateLimiter`. The name `Interval...` instead of `Bucket...` also clarifies the distinction from the [`X-RateLimit-Bucket`](https://discord.com/developers/docs/topics/rate-limits#header-format-rate-limit-header-examples) header for requests rate limits which have nothing to do with the `RateLimiter` interface in `dev.kord.common.ratelimit` and is a more fitting name (the "buckets" the original name was referring to are actually time intervals).

The `ClockIntervalRateLimiter` and `TimeSourceIntervalRateLimiter` classes are `internal` and can be constructed using the `public` factory function `IntervalRateLimiter()` which has two overloads for `TimeSource` and `Clock`. This results in shorter names in places where `RateLimiter`s are constructed, the underlying interval measurement mechanism is more of an implementation detail.
When no `TimeSource` / `Clock` is specified explicitly, `TimeSource.Monotonic` is used by default for the following reasons:
`RateLimiter` uses time measurement for the duration of time intervals.
The `TimeSource` interface is intended for measuring intervals, it literally says this as the first thing in its documentation: "A source of time for measuring time intervals." (`TimeSource.Monotonic` being the most precise one available.)
`Clock` however is intended for use with real time moments (`Instant`s), which don't provide any value for `IntervalRateLimiter`s.

### DIFF
--- a/common/src/main/kotlin/ratelimit/AbstractIntervalRateLimiter.kt
+++ b/common/src/main/kotlin/ratelimit/AbstractIntervalRateLimiter.kt
@@ -4,17 +4,11 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlin.time.Duration
 
-/**
- * A [RateLimiter] that restricts the number of [consume] invocations in [intervals][interval] to a given [limit].
- * Exceeding this limit results in suspension until the next interval.
- *
- * @param limit The maximum number of [consume] invocations allowed for each interval.
- * @param interval The [Duration] of each interval.
- */
-public abstract class IntervalRateLimiter(
-    public val limit: Int,
-    public val interval: Duration,
-) : RateLimiter {
+/** Abstract base class for implementing [IntervalRateLimiter]. */
+public abstract class AbstractIntervalRateLimiter(
+    final override val limit: Int,
+    final override val interval: Duration,
+) : IntervalRateLimiter {
 
     /** Remaining number of [consume] invocations allowed in the current interval. */
     protected var remaining: Int = limit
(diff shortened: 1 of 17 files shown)


HopeBaron:
there is no need to make these internal imo

Lukellmann:
> there is no need to make these internal imo

They have a rather long name, so I introduced the `IntervalRateLimiter()` factory functions. But since they don't expose any more functions or properties than `IntervalRateLimiter`, I saw no reason to be able to use both `IntervalRateLimiter(120, 60.seconds, TimeSource.Monotonic)` and `TimeSourceIntervalRateLimiter(120, 60.seconds, TimeSource.Monotonic)` so I made them internal.

Lukellmann:
Plus the contract of `IntervalRateLimiter` defines their behavior, so the actual class is only an implementation detail.

PR #579, opened 2022-03-31
Nullable channel name

Lukellmann:
see https://github.com/discord/discord-api-docs/pull/4571, merged in https://github.com/discord/discord-api-docs/commit/bf135de7dd0595cb9a73c5debbb0425dd5a6090c


HopeBaron:
when would a channel name be null?

HopeBaron:
so groups has no name?

Lukellmann:
I think so, yes. But since it's already optional there is no change in core.

PR #586, opened 2022-04-03
Deserialize time as `Duration`

Lukellmann:
The `Duration` type is Kotlin's way to express time amounts.
However, it is not serializable by default and time is serialized in different units across Discord's API.

To make `Duration` usable in serializable classes, this PR introduces a serializer for every `DurationUnit` that (de-)serializes `Duration`s as an integer number of the respective `DurationUnit`.

It also changes the types of `rate_limt_per_user`, `max_age` and `afk_timeout` to `Duration`, using the new serializers. `ArchiveDuration` now also uses `Duration` as the backing value.

### DIFF
--- a/common/src/main/kotlin/serialization/DurationSerializers.kt
+++ b/common/src/main/kotlin/serialization/DurationSerializers.kt
@@ -1,6 +1,7 @@
 package dev.kord.common.serialization
 
 import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
 import kotlinx.serialization.descriptors.PrimitiveKind
 import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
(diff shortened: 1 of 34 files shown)


Lukellmann:
In places where I had to annotate type parameters like 
```kotlin
val rateLimitPerUser: Optional<@Serializable(with = DurationInWholeSecondsSerializer::class) Duration> = Optional.Missing()
```
I've added import aliases to make it shorter and more readable
```kotlin
val rateLimitPerUser: Optional<@Serializable(InWholeSeconds::class) Duration> = Optional.Missing()
```

common/src/main/kotlin/serialization/DurationSerializers.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlin.time.toDuration
+
+
+/** Serializer that encodes and decodes [Duration]s. */
+public sealed class DurationSerializer(private val unit: DurationUnit, name: String) : KSerializer<Duration> {
+
+    final override val descriptor: SerialDescriptor =
+        PrimitiveSerialDescriptor("dev.kord.common.serialization.$name", PrimitiveKind.LONG)
+
+    protected abstract fun Duration.convert(): Long
```
DRSchlaubi, line 21: Why have this method if toLong(DurationUnit) exists?

Lukellmann: Fair point, didn't notic tbh. I will replace it then.

Lukellmann:
I haven't changed the type to `Duration` in cases where time is expressed as an integer number of days yet.

I suggest to only do this in places where the time unit is not already part of the property name (like `DiscordIntegration.expireGracePeriod`) and don't change it where the unit is mentioned in the property name (like `GuildBanCreateRequest.deleteMessageDays`).

common/src/main/kotlin/serialization/DurationSerializers.kt
```diff
@@ -28,23 +29,64 @@ public sealed class DurationSerializer(private val unit: DurationUnit, name: Str
 }
 
 
+// nanoseconds
+
 /** Serializer that encodes and decodes [Duration]s in [whole nanoseconds][Duration.inWholeNanoseconds]. */
 public object DurationInWholeNanosecondsSerializer : DurationSerializer(NANOSECONDS, "DurationInWholeNanoseconds")
```
GreemDev, line 35: Is there a particular reason for specifying that it's whole time units? 
It's serialized as a long which, to my knowledge, is always whole; right?

Lukellmann: I started implementing this using the `Duration.inWhole[Unit]s` properties, so I also included the "Whole" part in the names.
Leaving that part out would also be an option.

If we ever needed to introduce serializers that serialize `Duration`s to floating point values, the "Whole" part in the names would also help to distinguish between them. However, if we would find a meaningful name for the floating point ones, "Whole" might not be necessary.

Lukellmann: But to be honest, I would also prefer `DurationInSeconds` over `DurationInWholeSeconds`.

Lukellmann: I've decided to remove "Whole" from the names, let's worry about this in the future when we have to.

PR #587, opened 2022-04-04
Update to Kotlin 1.6.20 and update other dependencies

NoComment1105:
No code changes, just dependency updates

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.8.0"
     id("org.jetbrains.dokka")
 
-    id("org.ajoberstar.git-publish") version "4.1.0"
+    id("org.ajoberstar.git-publish") version "3.0.1"
 
     signing
     `maven-publish`
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,14 +4,14 @@ import org.gradle.api.tasks.wrapper.Wrapper.DistributionType.ALL
 plugins {
     kotlin("jvm")
     kotlin("plugin.serialization")
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.6.0"
+    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.8.0"
     id("org.jetbrains.dokka")
 
-    id("org.ajoberstar.git-publish") version "2.1.3"
+    id("org.ajoberstar.git-publish") version "4.1.0"
 
     signing
     `maven-publish`
-    id("io.codearte.nexus-staging") version "0.22.0"
+    id("io.codearte.nexus-staging") version "0.30.0"
 }
 
 repositories {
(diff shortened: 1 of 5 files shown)


Lukellmann:
Where there some compile warnings for the `-Xopt-in` flag? I remember seeing those in my projects with 1.6.20 pre-release versions.

NoComment1105:
There were. I can fix them if you'd like. `-Xopt-in` has been deprecated for `-opt-in`

Lukellmann:
> There were. I can fix them if you'd like. `-Xopt-in` has been deprecated for `-opt-in`

would be nice 👍

NoComment1105:
All sorted

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-all.zip
```
Lukellmann, line 3: Can you also update `gradleVersion` in the root `build.gradle.kts`.
This is used by the `wrapper` task which is the [recommended way to upgrade gradle](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper).

NoComment1105: I already have :>

Lukellmann: oops, I'm blind :D

Lukellmann (approved):

Lukellmann:
ktor 2.0.0 is right around the corner (there is an extra PR for that), but maybe we should update to already released 1.6.8 in this one?

build.gradle.kts
```diff
@@ -4,14 +4,14 @@ import org.gradle.api.tasks.wrapper.Wrapper.DistributionType.ALL
 plugins {
     kotlin("jvm")
     kotlin("plugin.serialization")
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.6.0"
+    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.8.0"
     id("org.jetbrains.dokka")
 
-    id("org.ajoberstar.git-publish") version "2.1.3"
+    id("org.ajoberstar.git-publish") version "3.0.1"
```
NyCodeGHG, line 10:
```suggestion
    id("org.ajoberstar.git-publish") version "4.1.0"
```

NyCodeGHG (commented):
git-publish's latest version is 4.1.0, but it's not shown on plugins.gradle.org, because they now publish to maven central (https://github.com/ajoberstar/gradle-git-publish#newest-versions-are-on-maven-central)

NyCodeGHG:
building with java 11+ would be an option, while still compiling on jvm target 8

NoComment1105:
> ktor 2.0.0 is right around the corner (there is an extra PR for that), but maybe we should update to already released 1.6.8 in this one?

That can be done, sure.

NoComment1105:
Done! There was no code changing required.

Lukellmann (approved):

PR #594, opened 2022-04-14
Discord only accepts language IDs separated by dashes when registering application command localizations

MrPowerGamerBR:
```
Invalid Form Body {"1":{"options":{"0":{"description_localizations":{"pt_BR":{"_errors":[{"code":"BASE_TYPE_CHOICES","message":"Value must be one of ('fi', 'cs', 'th', 'it', 'hi', 'bg', 'da', 'hu', 'en-GB', 'pt-BR', 'ru', 'ja', 'uk', 'es-ES', 'hr', 'ro', 'fr', 'nl', 'de', 'no', 'zh-TW', 'ko', 'el', 'tr', 'zh-CN', 'pl', 'en-US', 'vi', 'lt', 'sv-SE')."}]}}},"1":{"description_localizations":{"pt_BR":{"_errors":[{"code":"BASE_TYPE_CHOICES","message":"Value must be one of ('fi', 'cs', 'th', 'it', 'hi', 'bg', 'da', 'hu', 'en-GB', 'pt-BR', 'ru', 'ja', 'uk', 'es-ES', 'hr', 'ro', 'fr', 'nl', 'de', 'no', 'zh-TW', 'ko', 'el', 'tr', 'zh-CN', 'pl', 'en-US', 'vi', 'lt', 'sv-SE')."}]}}}}}}
```

### DIFF
--- a/common/src/main/kotlin/Locale.kt
+++ b/common/src/main/kotlin/Locale.kt
@@ -236,7 +236,7 @@ public data class Locale(val language: String, val country: String? = null) {
         override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Locale", PrimitiveKind.STRING)
 
         override fun serialize(encoder: Encoder, value: Locale) {
-            encoder.encodeString("${value.language}${if (value.country != null) "_${value.country}" else ""}")
+            encoder.encodeString("${value.language}${if (value.country != null) "-${value.country}" else ""}")
         }
 
         override fun deserialize(decoder: Decoder): Locale = fromString(decoder.decodeString())


Lukellmann (approved):
Thanks for catching this, I've only used German localizations so far, which won't trigger this problem.

MrPowerGamerBR:
By the way @Lukellmann, I think that the RegEx should also be updated. The comment says " We accept both "_" and "-" as a separator, because Discord doesn't really document it", however the locale list documentation shows that all locales use a dash as a separator, so I don't think that they would magically change them to a underscore (unless if Discord mixes and matches the separators in different API endpoints?)

https://discord.com/developers/docs/reference#locales

Lukellmann:
@DRSchlaubi, you implemeted `Locale`, what do you think?

DRSchlaubi:
it wasnÄt documented at the time but still not all use dashes? am i missing something here?

PR #597, opened 2022-04-17
Stricter Locale parsing

Lukellmann:
See https://github.com/kordlib/kord/pull/594#issuecomment-1099719976 (only allow `-` as separator between language and country)

Additional changes:
- only allow `[a-z]` for language and `[A-Z]` for country (is this actually desirable?)
- throw `IllegalArgumentException` instead of `IllegalStateException`
- coerce empty country to `null` for unknown Locales and `JLocale` to `Locale` conversion

### DIFF
--- a/common/src/main/kotlin/Locale.kt
+++ b/common/src/main/kotlin/Locale.kt
@@ -219,11 +219,12 @@ public data class Locale(val language: String, val country: String? = null) {
          * Decodes the language from a `languageCode-countryCode` or `languageCode` format.
          *
          * This does not validate the actually languages and countries, it just validates the format.
+         *
+         * @throws IllegalArgumentException if [string] is not a valid format.
          */
         public fun fromString(string: String): Locale {
-            val match = languageTagFormat.matchEntire(string) ?: error("$string is not a valid Locale")
-            val (language) = match.destructured
-            val country = match.groupValues[2]
+            val match = requireNotNull(languageTagFormat.matchEntire(string)) { "$string is not a valid Locale" }
+            val (language, country) = match.destructured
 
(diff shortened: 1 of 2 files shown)


common/src/main/kotlin/Locale.kt
```diff
@@ -212,31 +212,31 @@ public data class Locale(val language: String, val country: String? = null) {
...
-            val match = languageTagFormat.matchEntire(string) ?: error("$string is not a valid Locale")
-            val (language) = match.destructured
-            val country = match.groupValues[2]
+            val match = requireNotNull(languageTagFormat.matchEntire(string)) { "$string is not a valid Locale" }
+            val (language, country) = match.destructured
 
             return ALL.firstOrNull { (l, c) ->
                 language == l && country == (c ?: "")
-            } ?: Locale(language, country)
+            } ?: Locale(language, country.takeIf { it.isNotEmpty() })
```
DRSchlaubi, line 231:
```suggestion
            } ?: Locale(language, country.ifBlank { null })
```

common/src/main/kotlin/Locale.kt
```diff
@@ -247,4 +247,4 @@ public data class Locale(val language: String, val country: String? = null) {
  * Converts this into a [Locale].
  */
 public val JLocale.kLocale: Locale
-    get() = Locale(language, country)
+    get() = Locale(language, country.takeIf { it.isNotEmpty() })
```
DRSchlaubi, line 250:
```suggestion
    get() = Locale(language, country.ifBlank { null })
```

DRSchlaubi (changes requested):

DRSchlaubi (approved):

PR #601, opened 2022-04-23
Add API version configuration

Lukellmann:
This PR allows to configure which versions of Discord's APIs Kord uses.
This is achieved by changing the properties `REST_VERSION`, `GATEWAY_VERSION` or `VOICE_GATEWAY_VERSION` of the `KordConfiguration` object:
```kotlin
object KordConfiguration {
    var REST_VERSION: Int
    var GATEWAY_VERSION: Int
    var VOICE_GATEWAY_VERSION: Int
}
```

Reading the REST version from the environment variable `dev.kord.rest.version` is removed.

`DefaultGateway.Companion.invoke` now no longer uses an unversioned gateway url.

### DIFF
--- a/common/src/main/kotlin/KordConfiguration.kt
+++ b/common/src/main/kotlin/KordConfiguration.kt
@@ -10,8 +10,10 @@ public object KordConfiguration {
     // https://github.com/Kotlin/kotlinx.atomicfu/issues/186
     // TODO use delegation when AtomicFU fixes this
 
+    private const val REST_GATEWAY_DEFAULT = 10
 
-    private val REST = atomic(10)
+
+    private val REST = atomic(REST_GATEWAY_DEFAULT)
 
     /**
      * The [version of Discord's REST API](https://discord.com/developers/docs/reference#api-versioning) Kord uses.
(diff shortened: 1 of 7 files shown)


Lukellmann:
This is a first draft of the configuration options I mentioned [here](https://github.com/kordlib/kord/pull/575#issuecomment-1079787757).

Would be nice to get some feedback on this. Are there more things that could be configured in this way?
(@HopeBaron @DRSchlaubi @MrPowerGamerBR)

Lukellmann:
If there are no concerns about the design of this (mutable properties in an object), I would say this is ready for review.

HopeBaron:
Looks fine :+1:

Lukellmann:
[This PR](https://github.com/discord/discord-api-docs/pull/5012) removed the separately listed Gateway version, should we also only have one property for this?

Lukellmann:
> [This PR](https://github.com/discord/discord-api-docs/pull/5012) removed the separately listed Gateway version, should we also only have one property for this?

I thought about this and we probably shouldn't - if you want to change the version for only one of them you should be free to do so.

HopeBaron:
Do you mean to have a fixed set of versions?

Lukellmann:
I meant always using the same version for gateway and rest, but I think we should allow changing them independently.

HopeBaron:
Yeah it looks fine; we just need to fix the conflicts to go

Lukellmann:
Will do today/tomorrow 👍

Lukellmann:
Done

PR #602, opened 2022-04-26
Nullable user_limit

Lukellmann:
The `user_limit` field of a channel object can be `null`, even though it is [only documented as optional](https://discord.com/developers/docs/resources/channel#channel-object-channel-structure).
This PR fixes this.


Lukellmann:
This seems to be because of https://discordstatus.com/incidents/spzyl9yb72hj.
I've created [this branch](https://github.com/kordlib/kord/tree/fixes/nullable-channel-fields) with more fixes applied for the meantime. It can be used with the `fixes-nullable-channel-fields-SNAPSHOT` version.

PR #604, opened 2022-04-28
feature/slash commands permissions v2

DRSchlaubi:
Update to slash command permissions v2 - Deprecate defaultPermission in builders - Add new types - Add new events - Add builders for @everyone and all channels - Mark bulk update endpoints as @KordUnsafe (as it is currently disabled)

Superseeds #483

### DIFF
--- a/core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
+++ b/core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
@@ -1,39 +1,37 @@
 package dev.kord.core.behavior
 
-import dev.kord.common.annotation.KordUnsafe
 import dev.kord.core.cache.data.ApplicationCommandData
 import dev.kord.core.entity.application.ChatInputCommandCommand
 import dev.kord.core.entity.application.GlobalChatInputCommand
 import dev.kord.core.entity.application.GuildChatInputCommand
 import dev.kord.rest.builder.interaction.ChatInputModifyBuilder
-import dev.kord.rest.builder.interaction.ChatInputModifyBuilderImpl
 
 
 public interface ChatInputCommandBehavior : ApplicationCommandBehavior {
 
-     public suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): ChatInputCommandCommand
+    public suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): ChatInputCommandCommand
 
 }
 
 
 public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, GuildApplicationCommandBehavior {
 
-     override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GuildChatInputCommand {
-          val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
-          val response = service.modifyGuildApplicationCommand(applicationId, guildId, id, request)
-          val data = ApplicationCommandData.from(response)
-          return GuildChatInputCommand(data, service)
-     }
+    override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GuildChatInputCommand {
+        val response = service.modifyGuildChatInputApplicationCommand(applicationId, guildId, id) {
+            builder()
+        }
+        val data = ApplicationCommandData.from(response)
+        return GuildChatInputCommand(data, service)
+    }
 }
 
 
-
-public interface GlobalChatInputCommandBehavior : ChatInputCommandBehavior,GlobalApplicationCommandBehavior {
-     @OptIn(KordUnsafe::class)
-     override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GlobalChatInputCommand {
-          val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
-          val response = service.modifyGlobalApplicationCommand(applicationId,id, request)
-          val data = ApplicationCommandData.from(response)
-          return GlobalChatInputCommand(data, service)
-     }
+public interface GlobalChatInputCommandBehavior : ChatInputCommandBehavior, GlobalApplicationCommandBehavior {
+    override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GlobalChatInputCommand {
+        val response = service.modifyGlobalChatInputApplicationCommand(applicationId, id) {
+            builder()
+        }
+        val data = ApplicationCommandData.from(response)
+        return GlobalChatInputCommand(data, service)
+    }
 }
--- a/core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
+++ b/core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
@@ -1,10 +1,12 @@
 package dev.kord.core.behavior
 
+import dev.kord.common.annotation.KordUnsafe
 import dev.kord.core.cache.data.ApplicationCommandData
 import dev.kord.core.entity.application.ChatInputCommandCommand
 import dev.kord.core.entity.application.GlobalChatInputCommand
 import dev.kord.core.entity.application.GuildChatInputCommand
 import dev.kord.rest.builder.interaction.ChatInputModifyBuilder
+import dev.kord.rest.builder.interaction.ChatInputModifyBuilderImpl
 
 
 public interface ChatInputCommandBehavior : ApplicationCommandBehavior {
@@ -16,8 +18,9 @@ public interface ChatInputCommandBehavior : ApplicationCommandBehavior {
 
 public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, GuildApplicationCommandBehavior {
 
+     @OptIn(KordUnsafe::class)
      override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GuildChatInputCommand {
-          val request = ChatInputModifyBuilder().apply { builder() }.toRequest()
+          val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
           val response = service.modifyGuildApplicationCommand(applicationId, guildId, id, request)
           val data = ApplicationCommandData.from(response)
           return GuildChatInputCommand(data, service)
@@ -27,9 +30,9 @@ public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, Guild
 
 
 public interface GlobalChatInputCommandBehavior : ChatInputCommandBehavior,GlobalApplicationCommandBehavior {
-
+     @OptIn(KordUnsafe::class)
      override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GlobalChatInputCommand {
-          val request = ChatInputModifyBuilder().apply { builder() }.toRequest()
+          val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
           val response = service.modifyGlobalApplicationCommand(applicationId,id, request)
           val data = ApplicationCommandData.from(response)
           return GlobalChatInputCommand(data, service)
--- a/core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
+++ b/core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
@@ -18,7 +18,6 @@ public interface ChatInputCommandBehavior : ApplicationCommandBehavior {
 
 public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, GuildApplicationCommandBehavior {
 
-     @OptIn(KordUnsafe::class)
      override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GuildChatInputCommand {
           val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
           val response = service.modifyGuildApplicationCommand(applicationId, guildId, id, request)
--- a/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
+++ b/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
@@ -41,7 +41,7 @@ public data class ApplicationCommandData(
                     guildId,
                     options.mapList { ApplicationCommandOptionData.from(it) },
                     defaultMemberPermissions,
-                    dmPermissions,
+                    dmPermission,
                     defaultPermission,
                     version
                 )
--- a/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
+++ b/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
@@ -19,9 +19,9 @@ public data class ApplicationCommandData(
     val guildId: OptionalSnowflake = OptionalSnowflake.Missing,
     val options: Optional<List<ApplicationCommandOptionData>> = Optional.Missing(),
     val defaultMemberPermissions: Optional<Permissions?> = Optional.Missing(),
-    val dmPermission: OptionalBoolean = OptionalBoolean.Missing,
+    val dmPermission: OptionalBoolean? = OptionalBoolean.Missing,
     @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'.")
-    val defaultPermission: OptionalBoolean = OptionalBoolean.Missing,
+    val defaultPermission: OptionalBoolean? = OptionalBoolean.Missing,
     val version: Snowflake
 ) {
     public companion object {
--- a/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
+++ b/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
@@ -3,11 +3,7 @@ package dev.kord.core.cache.data
 import dev.kord.cache.api.data.DataDescription
 import dev.kord.cache.api.data.description
 import dev.kord.common.entity.*
-import dev.kord.common.entity.optional.Optional
-import dev.kord.common.entity.optional.OptionalBoolean
-import dev.kord.common.entity.optional.OptionalSnowflake
-import dev.kord.common.entity.optional.mapList
-import dev.kord.common.entity.optional.orEmpty
+import dev.kord.common.entity.optional.*
 import kotlinx.serialization.Serializable
 
 @Serializable
@@ -19,6 +15,8 @@ public data class ApplicationCommandData(
     val description: String?,
     val guildId: OptionalSnowflake,
     val options: Optional<List<ApplicationCommandOptionData>>,
+    val defaultMemberPermissions: Optional<Permissions?> = Optional.Missing(),
+    val dmPermission: OptionalBoolean = OptionalBoolean.Missing,
     val defaultPermission: OptionalBoolean = OptionalBoolean.Missing,
     val version: Snowflake
 ) {
@@ -36,6 +34,8 @@ public data class ApplicationCommandData(
                     description,
                     guildId,
                     options.mapList { ApplicationCommandOptionData.from(it) },
+                    defaultMemberPermissions,
+                    dmPermissions,
                     defaultPermission,
                     version
                 )
--- a/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
+++ b/core/src/main/kotlin/cache/data/ApplicationCommandData.kt
@@ -20,6 +20,7 @@ public data class ApplicationCommandData(
     val options: Optional<List<ApplicationCommandOptionData>> = Optional.Missing(),
     val defaultMemberPermissions: Optional<Permissions?> = Optional.Missing(),
     val dmPermission: OptionalBoolean = OptionalBoolean.Missing,
+    @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'.")
     val defaultPermission: OptionalBoolean = OptionalBoolean.Missing,
     val version: Snowflake
 ) {
--- a/core/src/main/kotlin/entity/application/ApplicationCommand.kt
+++ b/core/src/main/kotlin/entity/application/ApplicationCommand.kt
@@ -56,7 +56,7 @@ public sealed interface ApplicationCommand : ApplicationCommandBehavior {
      * whether the command is enabled by default when the app is added to a guild.
      */
     @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'.")
-    public val defaultPermission: Boolean? get() = data.defaultPermission?.discordBoolean
+    public val defaultPermission: Boolean? get() = data.defaultPermission.value
 
 
 }
--- a/core/src/main/kotlin/entity/application/ApplicationCommand.kt
+++ b/core/src/main/kotlin/entity/application/ApplicationCommand.kt
@@ -64,9 +64,9 @@ public sealed interface ApplicationCommand : ApplicationCommandBehavior {
 
 public interface GlobalApplicationCommand : ApplicationCommand, GlobalApplicationCommandBehavior {
     /**
-     * Whether this command can be executed in dms or not
+     * Whether this command is available in DMs with the application.
      */
-    public val dmPermission: Boolean? get() = data.dmPermission.value
+    public val dmPermission: Boolean get() = data.dmPermission.orElse(true)
 }
 public class UnknownGlobalApplicationCommand(
     override val data: ApplicationCommandData,
--- a/core/src/main/kotlin/entity/application/ApplicationCommand.kt
+++ b/core/src/main/kotlin/entity/application/ApplicationCommand.kt
@@ -55,6 +55,7 @@ public sealed interface ApplicationCommand : ApplicationCommandBehavior {
     /**
      * whether the command is enabled by default when the app is added to a guild.
      */
+    @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'.")
     public val defaultPermission: Boolean? get() = data.defaultPermission.discordBoolean
 
 
--- a/core/src/main/kotlin/entity/application/ApplicationCommand.kt
+++ b/core/src/main/kotlin/entity/application/ApplicationCommand.kt
@@ -56,7 +56,7 @@ public sealed interface ApplicationCommand : ApplicationCommandBehavior {
      * whether the command is enabled by default when the app is added to a guild.
      */
     @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'.")
-    public val defaultPermission: Boolean? get() = data.defaultPermission.discordBoolean
+    public val defaultPermission: Boolean? get() = data.defaultPermission?.discordBoolean
 
 
 }
--- a/core/src/main/kotlin/entity/application/ApplicationCommand.kt
+++ b/core/src/main/kotlin/entity/application/ApplicationCommand.kt
@@ -2,12 +2,15 @@ package dev.kord.core.entity.application
 
 import dev.kord.common.entity.ApplicationCommandType
 import dev.kord.common.entity.ChannelType
+import dev.kord.common.entity.Permissions
 import dev.kord.common.entity.Snowflake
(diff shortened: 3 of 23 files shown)


Lukellmann:
> Mark bulk update endpoints as @KordUnsafe (as it is currently disabled)

Wouldn't deprecating be more appropriate? I think what they mean with "Disabled the batch editing endpoint" is they completely removed it (but maybe I'm wrong here).

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
```diff
@@ -114,31 +158,45 @@ public class ChatInputCreateBuilder(
...
+public interface ChatInputModifyBuilder : ApplicationCommandModifyBuilder, RootInputChatBuilder {
+    public var description: String?
+}
 
 @KordDsl
-public class ChatInputModifyBuilder : ApplicationCommandModifyBuilder, RootInputChatBuilder {
+public interface GlobalChatInputModifyBuilder : ChatInputModifyBuilder, GlobalApplicationCommandModifyBuilder
+
+@KordUnsafe
+public class ChatInputModifyBuilderImpl : GlobalChatInputModifyBuilder {
```
HopeBaron, line 179: Can we make a factory function for this instead? `object : Parent`

rest/src/main/kotlin/builder/interaction/ApplicationCommandStateHolder.kt
```diff
@@ -17,6 +18,11 @@ internal class ApplicationCommandModifyStateHolder {
 
     var options: Optional<MutableList<OptionsBuilder>> = Optional.Missing()
 
+    var defaultMemberPermissions: Optional<Permissions?> = Optional.Missing()
+    var dmPermission: OptionalBoolean? = OptionalBoolean.Missing
+
+
+    @Deprecated("danger default_permission will soon be deprecated. You can instead set default_member_permissions to \"0\" to disable the command by default and/or set dm_permission to false to disable globally-scoped commands inside of DMs with your app")
```
HopeBaron, line 25: Probably wanna set this to error level

HopeBaron (changes requested):

DRSchlaubi:
> > Mark bulk update endpoints as @KordUnsafe (as it is currently disabled)
> 
> Wouldn't deprecating be more appropriate? I think what they mean with "Disabled the batch editing endpoint" is they completely removed it (but maybe I'm wrong here).

Why would the endpoint itself still be documented then?

rest/src/main/kotlin/builder/interaction/ApplicationCommandStateHolder.kt
DRSchlaubi: Not if the API still accepts it

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
DRSchlaubi: How would that help?

rest/src/main/kotlin/builder/interaction/ApplicationCommandStateHolder.kt
HopeBaron: The fact that *it should be changed asap* implies that it should be marked as error to force people to move away from it

DRSchlaubi: If it still works with the Discord API I am not convinced that introducing a compiler error is the right thing to do

HopeBaron: I think you are right

Lukellmann:
> > > Mark bulk update endpoints as @KordUnsafe (as it is currently disabled)
> > 
> > Wouldn't deprecating be more appropriate? I think what they mean with "Disabled the batch editing endpoint" is they completely removed it (but maybe I'm wrong here).
> 
> Why would the endpoint itself still be documented then?

Idk, but I asked in ddevs and someone said it's removed permanently: https://discord.com/channels/613425648685547541/697489244649816084/969995779966124083

DRSchlaubi:
> > > > Mark bulk update endpoints as @KordUnsafe (as it is currently disabled)
> > > 
> > > 
> > > Wouldn't deprecating be more appropriate? I think what they mean with "Disabled the batch editing endpoint" is they completely removed it (but maybe I'm wrong here).
> > 
> > 
> > Why would the endpoint itself still be documented then?
> 
> Idk, but I asked in ddevs and someone said it's removed permanently: https://discord.com/channels/613425648685547541/697489244649816084/969995779966124083

They said that for ephemerals as well

Lukellmann:
> > > > > Mark bulk update endpoints as @KordUnsafe (as it is currently disabled)
> > > > 
> > > > 
> > > > Wouldn't deprecating be more appropriate? I think what they mean with "Disabled the batch editing endpoint" is they completely removed it (but maybe I'm wrong here).
> > > 
> > > 
> > > Why would the endpoint itself still be documented then?
> > 
> > Idk, but I asked in ddevs and someone said it's removed permanently: https://discord.com/channels/613425648685547541/697489244649816084/969995779966124083
> 
> They said that for ephemerals as well

Haha true :D

Lukellmann:
Permissions v2 bring ["an inversion of control from the previous command permissions system"](https://github.com/discord/discord-api-docs/discussions/4853):
Bots are no longer meant to edit permissions, they should just provide defaults. Changing them is up to moderators or OAuth apps authenticated on behalf of authorized users.

The question now is whether Kord should support editing permissions at all, since we don't support other OAuth stuff.

If it won't be supported, we should just deprecate all places that previously allowed editing (probably with error level, Discord removed this functionality entirely for bots), including the bulk edit endpoint I mentioned [here](https://github.com/kordlib/kord/pull/604#issuecomment-1111648208) (it's definitely not coming back for bot tokens).

I would prefer this, as the current form of this PR looks to me like we are including partial OAuth functionality in an ad hoc way (adding bearer tokens as a function parameter in some places). If we ever decide to add proper OAuth support, we can also add support for editing command permissions.

core/src/main/kotlin/entity/application/ApplicationCommand.kt
```diff
@@ -45,16 +47,27 @@ public sealed interface ApplicationCommand : ApplicationCommandBehavior {
...
 
 }
 
 
-public interface GlobalApplicationCommand : ApplicationCommand, GlobalApplicationCommandBehavior
+public interface GlobalApplicationCommand : ApplicationCommand, GlobalApplicationCommandBehavior {
+    /**
+     * Whether this command can be executed in dms or not
+     */
+    public val dmPermission: Boolean? get() = data.dmPermission.value
```
Lukellmann, line 69:
```suggestion
    /**
     * Whether this command is available in DMs with the application.
     */
    public val dmPermission: Boolean get() = data.dmPermission.orElse(true)
```
commands are usable in DMs by default

core/src/main/kotlin/cache/data/ApplicationCommandData.kt
```diff
@@ -18,6 +18,9 @@ public data class ApplicationCommandData(
     val descriptionLocalizations: Optional<Map<Locale, String>?> = Optional.Missing(),
     val guildId: OptionalSnowflake = OptionalSnowflake.Missing,
     val options: Optional<List<ApplicationCommandOptionData>> = Optional.Missing(),
+    val defaultMemberPermissions: Optional<Permissions?> = Optional.Missing(),
+    val dmPermission: OptionalBoolean = OptionalBoolean.Missing,
```
Lukellmann, line 22:
```suggestion
    val dmPermission: OptionalBoolean? = OptionalBoolean.Missing,
```
see common

core/src/main/kotlin/cache/data/ApplicationCommandData.kt
```diff
@@ -18,6 +18,9 @@ public data class ApplicationCommandData(
...
+    @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'.")
     val defaultPermission: OptionalBoolean = OptionalBoolean.Missing,
```
Lukellmann, line 24:
```suggestion
    val defaultPermission: OptionalBoolean? = OptionalBoolean.Missing,
```
see common

core/src/main/kotlin/event/interaction/ApplicationCommandPermissionsUpdateEvent.kt
```diff
@@ -0,0 +1,15 @@
...
+import dev.kord.core.event.Event
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.job
+
+public class ApplicationCommandPermissionsUpdateEvent(
+    public val permissions: ApplicationCommandPermissions,
+    override val kord: Kord,
+    override val shard: Int,
+    public val coroutineScope: CoroutineScope = CoroutineScope(kord.coroutineContext + SupervisorJob(kord.coroutineContext.job)),
```
Lukellmann, line 14:
```suggestion
    public val coroutineScope: CoroutineScope = kordCoroutineScope(kord),
```

core/src/main/kotlin/gateway/handler/InteractionEventHandler.kt
```diff
@@ -25,6 +27,20 @@ public class InteractionEventHandler(
...
             is ApplicationCommandDelete -> handle(event, shard, kord, coroutineScope)
+            is ApplicationCommandPermissionsUpdate -> {
+                val data = GuildApplicationCommandPermissionsData(
+                    event.permissions.id,
+                    event.permissions.applicationId,
+                    event.permissions.guildId,
+                    event.permissions.permissions.map {
+                        GuildApplicationCommandPermissionData.from(it)
+                    }
+                )
```
Lukellmann, line 38:
```suggestion
                val data = GuildApplicationCommandPermissionsData.from(event.permissions)
```

Lukellmann (changes requested):

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
Lukellmann: we can make this internal when we don't create this builder in core but call the functions of the service that take builder-lambdas instead

DRSchlaubi: We can also make the class internal if we don't call it in core

Lukellmann: exactly, that's what I wanted to say

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
```diff
@@ -118,17 +162,24 @@ public class ChatInputCreateBuilder(
...
 }
 
+@KordDsl
+public interface ChatInputModifyBuilder : LocalizedDescriptionModifyBuilder, ApplicationCommandModifyBuilder,
+    RootInputChatBuilder
 
 @KordDsl
-public class ChatInputModifyBuilder : LocalizedDescriptionModifyBuilder, ApplicationCommandModifyBuilder,
-    RootInputChatBuilder {
+public interface GlobalChatInputModifyBuilder : ChatInputModifyBuilder, GlobalApplicationCommandModifyBuilder
```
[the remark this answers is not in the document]
DRSchlaubi: I am quite sure these don't work in DMs right?

core/src/main/kotlin/entity/application/ApplicationCommand.kt
DRSchlaubi: Isn't damper mission null by default?

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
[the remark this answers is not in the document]
Lukellmann: Oh, good point, I'm actually not sure.

core/src/main/kotlin/entity/application/ApplicationCommand.kt
Lukellmann: Sure, but `null`/missing means it's available in DMs, so why expose null here in core? (if only chat input commands can be used in DMs, this should also be moved to the global chat input command entity)

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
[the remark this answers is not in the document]
Lukellmann: Both can be used in DMs:
grafik
grafik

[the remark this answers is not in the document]
DRSchlaubi: That's nonsense to have them but sure then I need to update that

core/src/main/kotlin/entity/application/ApplicationCommand.kt
DRSchlaubi: I don't think dmÜermission should show, whether it's available in DMs, it should show the state of the permission, and that can be null

rest/src/main/kotlin/builder/interaction/UserCommandBuilders.kt
```diff
@@ -34,13 +39,18 @@ public class UserCommandCreateBuilder(override var name: String) : ApplicationCo
...
+    @Deprecated("'defaultPermission' is deprecated in favor of 'defaultMemberPermissions' and 'dmPermission'. Setting 'defaultPermission' to false can be replaced by setting 'defaultMemberPermissions' to empty Permissions and 'dmPermission' to false ('dmPermission' is only available for global commands).")
     override var defaultPermission: Boolean? by state::defaultPermission.delegate()
 
     override fun toRequest(): ApplicationCommandCreateRequest {
         return ApplicationCommandCreateRequest(
             name = name,
             nameLocalizations = state.nameLocalizations,
             type = type,
+            defaultMemberPermissions = state.defaultMemberPermissions,
+            dmPermission = state.dmPermission,
```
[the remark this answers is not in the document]
DRSchlaubi: It will when the new extraction is there

core/src/main/kotlin/entity/application/ApplicationCommand.kt
Lukellmann: Hm ok, probably not very important anyway.

rest/src/main/kotlin/builder/interaction/InputChatBuilders.kt
DRSchlaubi: rest doesn't have these functions, because so many builder have different instances, we might should make a single builder impl for all commands and then use interfaces, like it is done with Global/guild but that is out of scope for this pr

core/src/main/kotlin/entity/application/ApplicationCommand.kt
Lukellmann: The reason for this suggestion was the description of the `dm_permission` field:
> Indicates whether the command is available in DMs with the app, only for globally-scoped commands. By default, commands are visible.

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
```diff
@@ -16,8 +18,9 @@ public interface ChatInputCommandBehavior : ApplicationCommandBehavior {
 
 public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, GuildApplicationCommandBehavior {
 
+     @OptIn(KordUnsafe::class)
      override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GuildChatInputCommand {
-          val request = ChatInputModifyBuilder().apply { builder() }.toRequest()
+          val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
           val response = service.modifyGuildApplicationCommand(applicationId, guildId, id, request)
```
Lukellmann, line 24:
```suggestion
          val response = service.modifyGuildChatInputApplicationCommand(applicationId, guildId, id) { builder() }
```
allows `ChatInputModifyBuilderImpl` to be internal

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
```diff
@@ -16,8 +18,9 @@ public interface ChatInputCommandBehavior : ApplicationCommandBehavior {
 
 public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, GuildApplicationCommandBehavior {
 
+     @OptIn(KordUnsafe::class)
```
Lukellmann, line 21:
```suggestion
```

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
```diff
@@ -27,9 +30,9 @@ public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, Guild
 
 
 public interface GlobalChatInputCommandBehavior : ChatInputCommandBehavior,GlobalApplicationCommandBehavior {
-
+     @OptIn(KordUnsafe::class)
```
Lukellmann, line 33:
```suggestion

```

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
```diff
@@ -27,9 +30,9 @@ public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, Guild
...
      override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GlobalChatInputCommand {
-          val request = ChatInputModifyBuilder().apply { builder() }.toRequest()
+          val request = ChatInputModifyBuilderImpl().apply { builder() }.toRequest()
           val response = service.modifyGlobalApplicationCommand(applicationId,id, request)
```
Lukellmann, line 36:
```suggestion
          val response = service.modifyGlobalChatInputApplicationCommand(applicationId, id) { builder() }
```
allows `ChatInputModifyBuilderImpl` to be internal

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
```diff
@@ -27,9 +30,9 @@ public interface GuildChatInputCommandBehavior : ChatInputCommandBehavior, Guild
 
 
 public interface GlobalChatInputCommandBehavior : ChatInputCommandBehavior,GlobalApplicationCommandBehavior {
-
+     @OptIn(KordUnsafe::class)
      override suspend fun edit(builder: suspend ChatInputModifyBuilder.() -> Unit): GlobalChatInputCommand {
```
Lukellmann, line 34: Can we find a way to use `GlobalChatInputModifyBuilder` here as the extension receiver (also for user/message)? Overriding breaks this.

Lukellmann (changes requested):

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
Lukellmann: a not so elegant way would be to provide additional overloads as inline extensions

Lukellmann: or we just give up the base interface methods like in `MultiApplicationCommandBuilder`

DRSchlaubi: > a not so elegant way would be to provide additional overloads as inline extensions (might not even work as members take precedence over extensions)

It does work, but the ide will always rank the non extension one first, so no one will find it

> or we just give up the base interface methods like in MultiApplicationCommandBuilder

I don't think that's a good idea

rest/src/main/kotlin/service/InteractionService.kt
```diff
@@ -341,117 +342,98 @@ public class InteractionService(requestHandler: RequestHandler) : RestService(re
         applicationIdGuildIdCommandId(applicationId, guildId, commandId)
     }
 
-    public suspend fun editApplicationCommandPermissions(
-        applicationId: Snowflake,
-        guildId: Snowflake,
-        commandId: Snowflake,
-        request: ApplicationCommandPermissionsEditRequest,
-    ): DiscordGuildApplicationCommandPermissions = call(Route.ApplicationCommandPermissionsPut) {
```
Lukellmann, line 349: delete the route too

rest/src/main/kotlin/service/InteractionService.kt
```diff
@@ -341,117 +342,98 @@ public class InteractionService(requestHandler: RequestHandler) : RestService(re
...
-        applicationIdGuildIdCommandId(applicationId, guildId, commandId)
-        body(ApplicationCommandPermissionsEditRequest.serializer(), request)
-    }
-
-    public suspend fun bulkEditApplicationCommandPermissions(
-        applicationId: Snowflake,
-        guildId: Snowflake,
-        request: List<PartialDiscordGuildApplicationCommandPermissions>,
-    ): List<DiscordGuildApplicationCommandPermissions> = call(Route.ApplicationCommandPermissionsBatchPut) {
```
Lukellmann, line 358: delete the route too

Lukellmann (changes requested):

Lukellmann:
Should we provide a convenient function to set `defaultMemberPermissions` to an empty `Permissions` instance? This would disable the command for everyone by default (see [here](https://discord.com/developers/docs/interactions/application-commands#application-command-permissions-object-using-default-permissions)).

rest/src/main/kotlin/builder/interaction/ApplicationCommandBuilders.kt
```diff
@@ -2,83 +2,59 @@ package dev.kord.rest.builder.interaction
...
+    public var dmPermission: Boolean?
+
+    /**
+     * Disables the command for everyone by default.
+     *
+     * **This only ensures that DM users cannot execute the command, any server owner can change this setting**
+     */
+    override fun disableCommand() {
+        super.disableCommand()
+        dmPermission = false
```
Lukellmann, line 42: because of the virtual override and the only implementation implementing the global builder, this will always set `dmPermission` to `false` (also for guild commands)

DRSchlaubi: Does this really matter? won't Discord just ignore this for guilds? Like this we would need to introduce this parameter to all builders

core/src/main/kotlin/behavior/ChatInputCommandBehavior.kt
Lukellmann: the solution might be to have members in the subtypes and an extension for the base type which takes the more generic builder

rest/src/main/kotlin/builder/interaction/ApplicationCommandBuilders.kt
Lukellmann: I haven't tested it, so I don't know if Discord will ignore it, but I would prefer to do this properly.

rest/src/main/kotlin/builder/interaction/MultiApplicationCommandBuilder.kt
```diff
@@ -6,25 +6,46 @@ import kotlin.contracts.InvocationKind
...
-public class MultiApplicationCommandBuilder {
+public sealed class MultiApplicationCommandBuilder {
     public val commands: MutableList<ApplicationCommandCreateBuilder> = mutableListOf()
 
-    public inline fun message(name: String, builder: MessageCommandCreateBuilder.() -> Unit = {}) {
+    public fun build(): List<ApplicationCommandCreateRequest> {
+        return commands.map { it.toRequest() }
+    }
+}
+
```
Lukellmann, line 16:
```suggestion
public inline fun MultiApplicationCommandBuilder.input(
    name: String,
    description: String,
    builder: ChatInputCreateBuilder.() -> Unit = {},
) {
    contract { callsInPlace(builder, InvocationKind.EXACTLY_ONCE) }
    commands += ChatInputCreateBuilderImpl(name, description).apply(builder)
}

public inline fun MultiApplicationCommandBuilder.message(
    name: String,
    builder: MessageCommandCreateBuilder.() -> Unit = {},
) {
    contract { callsInPlace(builder, InvocationKind.EXACTLY_ONCE) }
    commands += MessageCommandCreateBuilderImpl(name).apply(builder)
}


public inline fun MultiApplicationCommandBuilder.user(name: String, builder: UserCommandCreateBuilder.() -> Unit = {}) {
    contract { callsInPlace(builder, InvocationKind.EXACTLY_ONCE) }
    commands += UserCommandCreateBuilderImpl(name).apply(builder)
}

```
this is what I was referring to in this comment: https://github.com/kordlib/kord/pull/604#discussion_r867287095
the same could be done for the command behaviors

rest/src/main/kotlin/builder/interaction/MultiApplicationCommandBuilder.kt
```diff
@@ -6,25 +6,46 @@ import kotlin.contracts.InvocationKind
...
+public class GlobalMultiApplicationCommandBuilder : MultiApplicationCommandBuilder() {
```
Lukellmann, line 17:
```suggestion
@KordDsl
public class GlobalMultiApplicationCommandBuilder : MultiApplicationCommandBuilder() {
```

rest/src/main/kotlin/builder/interaction/MultiApplicationCommandBuilder.kt
```diff
@@ -6,25 +6,46 @@ import kotlin.contracts.InvocationKind
...
     }
 
+
+    public inline fun user(name: String, builder: GlobalUserCommandCreateBuilder.() -> Unit = {}) {
+        contract { callsInPlace(builder, InvocationKind.EXACTLY_ONCE) }
+        commands += UserCommandCreateBuilderImpl(name).apply(builder)
+    }
+}
+
+public class GuildMultiApplicationCommandBuilder : MultiApplicationCommandBuilder() {
```
Lukellmann, line 35:
```suggestion
@KordDsl
public class GuildMultiApplicationCommandBuilder : MultiApplicationCommandBuilder() {
```

rest/src/main/kotlin/builder/interaction/UserCommandBuilders.kt
```diff
@@ -3,44 +3,68 @@ package dev.kord.rest.builder.interaction
...
             defaultPermission = state.defaultPermission
         )
     }
 }
 
 @KordDsl
-public class UserCommandCreateBuilder(override var name: String) : ApplicationCommandCreateBuilder {
+public interface UserCommandCreateBuilder : ApplicationCommandCreateBuilder
+
+public interface GlobalUserCommandCreateBuilder : UserCommandCreateBuilder, GlobalApplicationCommandCreateBuilder
```
Lukellmann, line 45:
```suggestion
@KordDsl
public interface GlobalUserCommandCreateBuilder : UserCommandCreateBuilder, GlobalApplicationCommandCreateBuilder
```

rest/src/main/kotlin/route/Route.kt
```diff
@@ -810,7 +810,8 @@ public sealed class Route<T>(
         Route<DiscordGuildApplicationCommandPermissions>(
             HttpMethod.Put,
             "/applications/$ApplicationId/guilds/$GuildId/commands/$CommandId/permissions",
-            DiscordGuildApplicationCommandPermissions.serializer()
+            DiscordGuildApplicationCommandPermissions.serializer(),
+            false
```
Lukellmann, line 814: either remove/deprecate the permission edit routes or change this back, setting `requiresAuthorizationHeader` to `false` is simply wrong (it still requires a authorization header, but not with a bot token)

Lukellmann (changes requested):

rest/src/main/kotlin/builder/interaction/ApplicationCommandBuilders.kt
DRSchlaubi: Either we ignore it, we just don't have this function at all because it's less work to just do this manually, or we just drop dm support in the function

Lukellmann: > or we just drop dm support in the function

I think this is the easiest thing to do, if we also rename it to `disableCommandInGuild()` it should be clear that it doesn't do anything about dms

Another note: the docs of this method should refer to admins not server owners.

PR #614, opened 2022-06-05
Voice text

HopeBaron:


core/src/main/kotlin/entity/channel/VoiceChannel.kt
```diff
@@ -23,7 +25,7 @@ public class VoiceChannel(
     override val data: ChannelData,
     override val kord: Kord,
     override val supplier: EntitySupplier = kord.defaultSupplier
-) : CategorizableChannel, VoiceChannelBehavior {
+) : CategorizableChannel, TopGuildMessageChannel, VoiceChannelBehavior {
```
Lukellmann, line 28:
```suggestion
) : TopGuildMessageChannel, VoiceChannelBehavior {
```
`TopGuildMessageChannel` is already a `CategorizableChannel`

core/src/main/kotlin/entity/channel/VoiceChannel.kt
```diff
@@ -74,6 +76,10 @@ public class VoiceChannel(
 
     override suspend fun asChannelOrNull(): VoiceChannel = this
 
+    override suspend fun fetchChannelOrNull(): VoiceChannel? = supplier.getChannelOfOrNull(id)
+
+    override suspend fun fetchChannel(): VoiceChannel = supplier.getChannelOf(id)
```
Lukellmann, line 81: Why do we need these overrides?

Lukellmann (changes requested):

core/src/main/kotlin/entity/channel/VoiceChannel.kt
HopeBaron: to return a voice type channel

Lukellmann (approved):

NoComment1105 (approved):

PR #618, opened 2022-06-14
Revamp kord builder

HopeBaron:
This rewrites #551 introducing a new Proxy Kord builder instead of hacking over the current implementations;
More adjustments will be considered as of we've seen people use the `restOnly` implementation of Kord


core/src/main/kotlin/builder/kord/ProxyKordBuilder.kt
```diff
@@ -0,0 +1,84 @@
...
+import io.ktor.client.HttpClient
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableSharedFlow
+
+/**
+ * The rest only Kord builder. You probably want to invoke the [DSL builder][Kord.restOnly] instead.
+ */
+@KordExperimental
+public class KordProxyBuilder {
```
Lukellmann, line 26: This is almost the same as `KordRestOnlyBuilder`, I think it would be better to have a common base class instead of copy pasting.

I think we should also have a similar thing like the `Kord.restOnly()` function for this builder.

rest/src/main/kotlin/request/RequestHandler.kt
```diff
@@ -16,4 +22,11 @@ public interface RequestHandler {
      */
     @Throws(RestRequestException::class)
     public suspend fun <B : Any, R> handle(request: Request<B, R>): R
+    public suspend fun <T> intercept(builder: RequestBuilder<T>): RequestBuilder<T> = builder.apply {
```
Lukellmann, line 25: Does it need to be `suspend`?
Should it really return a `RequestBuilder` or just take it as an argument?

rest/src/main/kotlin/request/KtorRequestHandler.kt
```diff
@@ -75,7 +75,7 @@ public class KtorRequestHandler(
         headers.appendAll(request.headers)
 
         url {
-            url.takeFrom(dev.kord.rest.route.Route.baseUrl)
+            url.takeFrom(request.baseUrl)
```
Lukellmann, line 78: this should also be changed in the `CrashingHandler` test class

Lukellmann (changes requested):

core/src/main/kotlin/Kord.kt
```diff
@@ -403,6 +405,13 @@ public class Kord(
             return KordRestOnlyBuilder(token).apply(builder).build()
         }
     }
+
+    public inline fun proxy(builder: KordProxyBuilder.() -> Unit = {}): Kord {
```
Lukellmann, line 409: This has to be in the companion object like `restOnly`.
Docs would be nice too.
Maybe `@KordExperimental` too?

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+import dev.kord.core.gateway.DefaultMasterGateway
+import dev.kord.core.supplier.EntitySupplyStrategy
+import dev.kord.gateway.Gateway
+import dev.kord.gateway.builder.Shards
+import dev.kord.rest.service.RestClient
+import kotlinx.coroutines.flow.MutableSharedFlow
+
+/**
+ * The rest only Kord builder. You probably want to invoke the [DSL builder][Kord.restOnly] instead.
+ */
```
Lukellmann, line 17:
```suggestion
/**
 * The proxy Kord builder. You probably want to invoke the [DSL builder][Kord.proxy] instead.
 */
```

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+@KordExperimental
+public class KordProxyBuilder: RestOnlyBuilder() {
+
+    public override fun build(): Kord {
+        val client = httpClient.configure()
+
+        val resources = ClientResources(
+            "",
+            applicationId ?: Snowflake.min,
```
Lukellmann, line 26: I think it's problematic to provide `Snowflake.min` as a default, we use the application id in some places. Should we throw an exception instead?

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+import dev.kord.rest.service.RestClient
+import kotlinx.coroutines.flow.MutableSharedFlow
+
+/**
+ * The rest only Kord builder. You probably want to invoke the [DSL builder][Kord.restOnly] instead.
+ */
+@KordExperimental
+public class KordProxyBuilder: RestOnlyBuilder() {
+
+    public override fun build(): Kord {
```
Lukellmann, line 21: there is still a lot of duplicated logic, can we put more of it in the common base class?

core/src/main/kotlin/builder/kord/RestOnlyBuilder.kt
```diff
@@ -0,0 +1,44 @@
...
+    public var httpClient: HttpClient? = null
+
+    public var applicationId: Snowflake? = null
+
+    /**
+     * Configures the [RequestHandler] for the [RestClient].
+     *
+     * ```
+     * Kord(token) {
+     *     { resources -> KtorRequestHandler(resources.httpClient, ExclusionRequestRateLimiter()) }
```
Lukellmann, line 35:
```suggestion
     *     requestHandler { resources -> KtorRequestHandler(resources.httpClient, ExclusionRequestRateLimiter()) }
```

Lukellmann (changes requested):

rest/src/main/kotlin/request/RequestBuilder.kt
```diff
@@ -6,8 +6,9 @@ import dev.kord.rest.route.Route
 import io.ktor.http.*
 import kotlinx.serialization.SerializationStrategy
 
-public class RequestBuilder<T>(private val route: Route<T>, keySize: Int = 2) {
+public class RequestBuilder<T>(public var route: Route<T>, keySize: Int = 2) {
```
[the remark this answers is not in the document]
HopeBaron: I thought that someone may decide to change the route inside some custom logic later

rest/src/main/kotlin/request/RequestHandler.kt
HopeBaron: someone may need to do some suspending logic; or so I thought

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
HopeBaron: I can't think of a better alternative tbf; the id is required so

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
HopeBaron: You are right we can add duplicate code into the abstract class as well

rest/src/main/kotlin/request/RequestHandler.kt
Lukellmann: But what about the returned builder, is that needed?

HopeBaron: technically no since you are expected to edit the builder in place

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
HopeBaron: we can't cause someone may not supply any

HopeBaron:
I've fixed the issues we agreed on except the refactoring part; I think it can be in the next release

core/src/main/kotlin/builder/kord/KordProxyBuilder.kt
Lukellmann: but then the bot may not function correctly and throw in other places

HopeBaron: should we create a mocked version of some sorts?
@WinteryFox would you supply an applicationId here?
I mean as a Proxy Kord Builder we still need some applicationId to work on so @Lukellmann it does make sense to throw

WinteryFox: I do supply an application ID, you can just require it in the constructor. Since I'm currently still using the workaround of multiple Kord instances its fine for now, but should probably get a more long-term fix in the future.

rest/src/main/kotlin/request/RequestBuilder.kt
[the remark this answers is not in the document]
Lukellmann: I've converted it back to `val` (but left it public), if there really is a need for it to be a `var`, we can change it again in the future.

core/src/main/kotlin/Kord.kt
```diff
@@ -402,7 +403,26 @@ public class Kord(
...
+         * Builds a [Kord] instance configured by the [builder].
+         *
+         * The instance only allows for configuration of REST related APIs,
+         * interacting with the [gateway][Kord.gateway] or its [events][Kord.events] will result in no-ops.
+         *
+         * Similarly, [cache][Kord.cache] related functionality has been disabled and
+         * replaced with a no-op implementation.
+         */
+        @KordExperimental
+        public inline fun proxy(applicationId: Snowflake, builder: KordProxyBuilder.() -> Unit = {}): Kord {
```
Lukellmann, line 417: I think this still needs some better documentation to explain it's purpose. Do you have some suggestions?

Issue #613, opened 2022-06-03
Connection to VoiceGateway websocket throws HTTP 400 Error


sebampuero:
Whenever using `voiceChannel?.connect{}` the following is thrown:

`[DefaultDispatcher-worker-13] ERROR dev.kord.voice.gateway.DefaultVoiceGateway - No transformation found: class io.ktor.utils.io.ByteBufferChannel (Kotlin reflection is not available) -> class io.ktor.client.plugins.websocket.DefaultClientWebSocketSession (Kotlin reflection is not available)
with response from wss://russia3325.discord.media?v=4:
status: 400 Bad Request
response headers: 
Server: cloudflare
, Date: Fri, 03 Jun 2022 16:06:53 GMT
, Content-Type: text/html
, Content-Length: 155
, Connection: close
, CF-RAY: -`

The bot enters the voice channel , but it does not send any voice data when using an AudioProvider

lost-illusi0n:
#626 has been merged and should solve this. This can be closed.

PR #624, opened 2022-06-19
fix(voice): ktor 2 migrations

lost-illusi0n:
The move over to Ktor 2.0 has been merged into Kord, and everything compiles, however a few migrations were needed to be done to ensure voice was working. This PR takes care of that and should fix kordlib/kord#613 as well.

### DIFF
--- a/voice/src/main/kotlin/VoiceConnectionBuilder.kt
+++ b/voice/src/main/kotlin/VoiceConnectionBuilder.kt
@@ -142,7 +142,7 @@ public class VoiceConnectionBuilder(
             voiceState.sessionId
         ) to VoiceGatewayConfiguration(
             voiceServer.token,
-            "wss://${voiceServer.endpoint}?v=4"
+            "wss://${voiceServer.endpoint}/?v=4"
         )
     }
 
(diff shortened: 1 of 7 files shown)


voice/src/main/kotlin/VoiceConnectionBuilder.kt
```diff
@@ -142,7 +142,7 @@ public class VoiceConnectionBuilder(
             voiceState.sessionId
         ) to VoiceGatewayConfiguration(
             voiceServer.token,
-            "wss://${voiceServer.endpoint}?v=4"
+            "wss://${voiceServer.endpoint}/?v=4"
```
Lukellmann, line 145: Should this also be changed [here](https://github.com/kordlib/kord/blob/0.8.x/voice/src/main/kotlin/handlers/VoiceUpdateEventHandler.kt#L67)?

lost-illusi0n: Yes, good catch.

DRSchlaubi:
Replaced by #626

PR #632, opened 2022-06-27
Fix nonce deserialization (Alternative Solution)

### LINKED ISSUE #629: Message `nonce` field can be Integer or String, but Kord always deserializes it as a String
```json
{"d": {"id": "990134009365401670", "tts": false, "type": 0, "flags": 0, "nonce": 6870, "author": {"id": "330770985450078208", "bot": true, "avatar": "6668e8af53a272e4cd72888d2183c97f", "username": "Aperture", "public_flags": 65536, "discriminator": "8658", "avatar_decoration": null}, "embeds": [], "member": {"deaf": false, "mute": false, "nick": null, "user": {"id": "330770985450078208", "bot": true, "avatar": "6668e8af53a272e4cd72888d2183c97f", "username": "Aperture", "public_flags": 65536, "discriminator": "8658", "avatar_decoration": null}, "flags": 0, "roles": ["753439237856165889", "751233962428596274"], "avatar": null, "pending": false, "joined_at": "2020-09-10T02:18:32.579000+00:00", "premium_since": null, "communication_disabled_until": null}, "pinned": false, "content": "`[02:59:03]` :outbox_tray: RGN-Rafaelk#​4169 (`568682176087064576`) left the server", "guild_id": "751223482515325088", "mentions": [], "timestamp": "2022-06-25T05:59:03.633000+00:00", "channel_id": "752735099971436584", "components": [], "attachments": [], "mention_roles": [], "edited_timestamp": null, "mention_everyone": false, "referenced_message": null}, "s": 246086, "t": "MESSAGE_CREATE", "op": 0}
```

```
kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 113: Expected quotation mark '"', but had '0' instead at path: $.d.nonce
```

MrPowerGamerBR:
image

Fixes #629

Only #632 or #630 must be merged, they both fix the same issue, but in different ways.

### DIFF
--- a/common/src/main/kotlin/entity/AuditLog.kt
+++ b/common/src/main/kotlin/entity/AuditLog.kt
@@ -311,9 +311,9 @@ public sealed class AuditLogChangeKey<T>(public val name: String, public val ser
      */
     // Audit Log Change Key "type" has integer or string values, so we need some sort of union serializer
     // (see https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-audit-log-change-key)
+    // TODO use union type `String | Int` if Kotlin ever introduces them
     @SerialName("type")
-    public object Type : AuditLogChangeKey<String>("type", IntOrStringSerializer) {
-        // TODO use union type `String | Int` if Kotlin ever introduces them
+    public object Type : AuditLogChangeKey<String>("type", IntOrStringSerializer)
     }
 
     @SerialName("enable_emoticons")
--- a/common/src/main/kotlin/entity/AuditLog.kt
+++ b/common/src/main/kotlin/entity/AuditLog.kt
@@ -314,7 +314,6 @@ public sealed class AuditLogChangeKey<T>(public val name: String, public val ser
     // TODO use union type `String | Int` if Kotlin ever introduces them
     @SerialName("type")
     public object Type : AuditLogChangeKey<String>("type", IntOrStringSerializer)
-    }
 
     @SerialName("enable_emoticons")
     public object EnableEmoticons : AuditLogChangeKey<Boolean>("enable_emoticons", serializer())
--- a/common/src/main/kotlin/entity/AuditLog.kt
+++ b/common/src/main/kotlin/entity/AuditLog.kt
@@ -5,6 +5,7 @@ import dev.kord.common.entity.optional.OptionalSnowflake
 import dev.kord.common.entity.optional.orEmpty
 import dev.kord.common.serialization.DurationInDaysSerializer
 import dev.kord.common.serialization.DurationInSecondsSerializer
+import dev.kord.common.serialization.IntOrStringSerializer
 import kotlinx.datetime.Instant
 import kotlinx.serialization.*
 import kotlinx.serialization.builtins.serializer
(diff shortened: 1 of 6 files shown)


common/src/main/kotlin/entity/DiscordMessage.kt
```diff
@@ -114,6 +121,26 @@ public data class DiscordMessage(
     val thread: Optional<DiscordChannel> = Optional.Missing()
 )
 
+internal object NonceAsStringSerializer : KSerializer<String> {
```
Lukellmann, line 124: I've implemented something similar [here](https://github.com/kordlib/kord/blob/0.8.x/common/src/main/kotlin/entity/AuditLog.kt#L319). I suggest a refactor to use the same serializer in both places.
A good place to put this serializer would be [here](https://github.com/kordlib/kord/tree/0.8.x/common/src/main/kotlin/serialization).

Lukellmann (changes requested):
I would also go for this solution, this doesn't introduce yet another type.

common/src/main/kotlin/entity/AuditLog.kt
```diff
@@ -310,39 +309,11 @@ public sealed class AuditLogChangeKey<T>(public val name: String, public val ser
...
-                val jsonPrimitive = value.toIntOrNull()?.let { JsonPrimitive(it) } ?: JsonPrimitive(value)
-                encoder.encodeSerializableValue(backingSerializer, jsonPrimitive)
-            }
-
-            override fun deserialize(decoder: Decoder): String {
-                val jsonPrimitive = decoder.decodeSerializableValue(backingSerializer)
-                return if (jsonPrimitive.isString) jsonPrimitive.content else jsonPrimitive.int.toString()
-            }
-        }
     }
```
Lukellmann, line 317:
```suggestion
```
the block is no longer needed

common/src/main/kotlin/entity/DiscordMessage.kt
```diff
@@ -13,6 +14,13 @@ import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonContentPolymorphicSerializer
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.int
+import kotlinx.serialization.json.intOrNull
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
```
Lukellmann, line 23:
```suggestion
```
we no longer need these imports

Lukellmann (commented):
looks good overall, just some minor suggestions

MrPowerGamerBR:
For some reason the "Add suggestion to batch" was grayed out, that's why I needed to apply all suggestions one by one 😭

Lukellmann:
> For some reason the "Add suggestion to batch" was grayed out, that's why I needed to apply all suggestions one by one 😭

iirc applying batch suggestions is only available in the "Files changed" tab

PR #634, opened 2022-07-01
Fix threads cache

Lukellmann:
This PR fixes a bug that threads weren't cached when receiving Guild events.
It also deprecates `Guild.threads` (it does not behave like one might expect) and instead replaces it with `GuildBehavior.cachedThreads`.

### DIFF
--- a/core/src/main/kotlin/behavior/GuildBehavior.kt
+++ b/core/src/main/kotlin/behavior/GuildBehavior.kt
@@ -68,12 +68,25 @@ public interface GuildBehavior : KordEntity, Strategizable {
      * Returns all active public and private threads in this guild
      * Threads are ordered by their id, in descending order.
      *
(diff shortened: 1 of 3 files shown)


HopeBaron:
Is there a way we could get these through rest?
if so we'd go with transforming this into a function that accepts a supplier

HopeBaron:
your primary concern is weather you can cache it from gateway or not; it's up to the user to do it.

Lukellmann:
What exactly does this mean for this PR?

HopeBaron:
Looking back at this this looks just fine

Lukellmann:
Can be merged then?

gdude2002:
Would be great to see a merge for this

gdude2002:
Thanks o7

PR #635, opened 2022-07-01
Stricter compilation, clean up opt-ins

Lukellmann:
This PR changes some compilation settings and opt-in behavior.

### Stricter compilation
- enable [progressive mode](https://kotlinlang.org/docs/whatsnew13.html#progressive-mode) - will give us fixes earlier
- enable [`allWarningsAsErrors`](https://kotlinlang.org/docs/gradle.html#attributes-common-to-jvm-js-and-js-dce) - will force us to care about warnings

### Clean up opt-ins
- remove project-wide opt-in for `ExperimentalStdlibApi` - we don't use it
- add project-wide opt-in for our own `KordPreview`, `KordExperimental` and `KordVoice` - we know where we can use our own experimental stuff and this way we don't need to opt-in manually everywhere
- remove project-wide opt-in for `ExperimentalCoroutinesApi` - only keep it in "test" source sets for `runTest {}`

### DIFF
--- a/buildSrc/src/main/kotlin/Compiler.kt
+++ b/buildSrc/src/main/kotlin/Compiler.kt
@@ -1,5 +1,4 @@
 object CompilerArguments {
-    const val coroutines = "-opt-in=kotlinx.coroutines.ExperimentalCoroutinesApi"
     const val time = "-opt-in=kotlin.time.ExperimentalTime"
     const val contracts = "-opt-in=kotlin.contracts.ExperimentalContracts"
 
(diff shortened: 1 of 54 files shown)


Lukellmann:
> Looks fine Im just a bit confused about the freeCompilerArgs change in the build script

In core or in the common build script?
And what exactly causes your confusion?

HopeBaron:
Why the compile block was removed

Lukellmann:
The only thing it did was adding the opt-in for experimental stdlib, however we didn't use it. We can add it back if there is a need for it in a lot of places or just use individual opt-in.

PR #638, opened 2022-07-02
Update application command fields

Lukellmann:
- `default_member_permissions` no longer optional
- `dm_permission` no longer nullable

see https://github.com/discord/discord-api-docs/pull/4991, merged in https://github.com/discord/discord-api-docs/commit/ad81ed31372245d2b4f7006924bd8dd68130a87f

### DIFF
--- a/core/src/main/kotlin/entity/application/ApplicationCommand.kt
+++ b/core/src/main/kotlin/entity/application/ApplicationCommand.kt
@@ -50,7 +50,7 @@ public sealed interface ApplicationCommand : ApplicationCommandBehavior {
     /**
      * Set of [Permissions] required to execute this command unless a server admin changed them.
      */
-    public val defaultMemberPermissions: Permissions? get() = data.defaultMemberPermissions.value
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/entity/application/ApplicationCommand.kt
```diff
@@ -66,7 +66,7 @@ public interface GlobalApplicationCommand : ApplicationCommand, GlobalApplicatio
     /**
      * Whether this command is available in DMs with the application.
      */
-    public val dmPermission: Boolean get() = data.dmPermission?.value ?: true
+    public val dmPermission: Boolean get() = data.dmPermission.orElse(true)
```
HopeBaron, line 69: I guess this should be or else false cause if it's absent it means it's not permited in Dms?

Lukellmann: [This table](https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-structure) suggest the opposite:
> By default, commands are visible.

PR #633, opened 2022-06-29
Add `app_permissions` field

MrPowerGamerBR:
https://discord.com/developers/docs/change-log#calculated-permissions-in-interaction-payloads

PR #646, opened 2022-07-05
Add `app_permissions` field 

Lukellmann:
Replaces #633 (credits to @MrPowerGamerBR).

see https://github.com/discord/discord-api-docs/pull/5131, merged in https://github.com/discord/discord-api-docs/commit/f4e3955f673c7f26b665f64411d8353bda7d0525

### DIFF
--- a/core/src/main/kotlin/entity/interaction/GuildInteraction.kt
+++ b/core/src/main/kotlin/entity/interaction/GuildInteraction.kt
@@ -12,11 +12,20 @@ public sealed interface GuildInteraction : Interaction, GuildInteractionBehavior
 
     override val guildId: Snowflake get() = data.guildId.value!!
 
-    /** Overridden permissions of the interaction [invoker][user] in the [channel][GuildInteractionBehavior.channel]. */
+    /**
+     * [Permissions] the [interaction invoker][user] has within the [channel][GuildInteractionBehavior.channel] the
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/entity/interaction/GuildInteraction.kt
```diff
@@ -12,11 +12,20 @@ public sealed interface GuildInteraction : Interaction, GuildInteractionBehavior
 
     override val guildId: Snowflake get() = data.guildId.value!!
 
-    /** Overridden permissions of the interaction [invoker][user] in the [channel][GuildInteractionBehavior.channel]. */
+    /**
+     * [Permissions] the [interaction invoker][user] has within the [channel][GuildInteractionBehavior.channel] the
+     * interaction was sent from.
+     */
     public val permissions: Permissions get() = data.permissions.value!!
```
Lukellmann, line 19: Should we rename this to `userPermissions` or `memberPermissions`? That way it would be harder to confuse it with `appPermissions`.

HopeBaron (approved):

core/src/main/kotlin/entity/interaction/GuildInteraction.kt
HopeBaron: `app` tells the difference between both doesn't it

PR #652, opened 2022-07-10
Update dependencies

Lukellmann:
- Binary compatibility validator 0.9.0 -> 0.10.1
- Kotlin 1.7.0 -> 1.7.10
- Dokka 1.6.21 -> 1.7.0
- AtomicFU 0.17.3 -> 0.18.2
- gradle-buildconfig-plugin 3.0.3 -> 3.1.0
- kotlinx-datetime 0.3.3 -> 0.4.0
- kotlinx.coroutines 1.6.2 -> 1.6.3

- don't set java `sourceCompatibility`, the warning doesn't show anymore

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -1,6 +1,8 @@
 plugins {
     groovy
     `kotlin-dsl`
+    kotlin("jvm") version "1.7.10"
+    kotlin("plugin.serialization") version "1.7.10"
 }
 
 repositories {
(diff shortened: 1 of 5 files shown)


buildSrc/build.gradle.kts
```diff
@@ -9,10 +13,14 @@ repositories {
...
+    implementation(kotlin("serialization"))
+    implementation("org.jetbrains.dokka", "dokka-gradle-plugin", "1.7.0")
+    implementation("org.jetbrains.kotlinx", "atomicfu-gradle-plugin", "0.18.2")
     implementation(gradleApi())
     implementation(localGroovy())
 }
+
+tasks.withType<KotlinCompile> {
+    kotlinOptions.languageVersion = "1.5"
+}
```
Lukellmann, line 26: Do we need this? Everything needs to work fine without it.

NoComment1105: That should be set to 1.7, if you do keep it. If I remember correctly there is something it does, but I don't fill remember

HopeBaron: languageVersion | Provide source compatibility with the specified version of Kotlin | "1.4" (DEPRECATED), "1.5", "1.6", "1.7"
-- | -- | --


I think it should stay on 1.5

PR #657, opened 2022-07-23
feature(voice): add helper to move/update voice connections

lost-illusi0n:


MrPowerGamerBR:
Context
image

voice/src/main/kotlin/VoiceConnectionBuilder.kt
```diff
@@ -96,6 +98,12 @@ public class VoiceConnectionBuilder(
      */
     public var streams: Streams? = null
 
+    /**
+     * The amount of time the connection should wait before assuming the voice connection has been closed instead of
+     * moved.
+     */
+    public var connectionDetachDuration: Duration = 100.milliseconds
```
Lukellmann, line 105: Shouldn't this default to the 5 seconds it was before? This way the default behavior stays the same.

lost-illusi0n: I was thinking that, but I'm really not sure why I choose 5 seconds before. It truly is unbelievably long for no real benefit. I can only see confusing behavior from this, regardless if it was the default before. Such as if a user kicked the bot and then told it to rejoin within those 5 seconds, the same connection would be used, which is a bit counter-intuitive. And there still is a 5-second delay between the bot being kicked and the bot actually knowing it was kicked, which IMO is pretty bad. Imagine you want to send a message to bot kick, that message would be delayed by 5 seconds. I've done tests and a smaller threshold is completely fine and makes the behavior much more expected.

Lukellmann: Makes sense 👍

PR #660, opened 2022-07-28
Reconfigure Dokka and suppress documentation for deprecated declarations

Lukellmann:
This PR reconfigures and updates Dokka:
- Dokka 1.7.0 -> 1.7.10
- also configure `dokkaHtmlPartial` tasks, only `dokkaHtml` tasks were configured before (but they aren't usedby multi-module projects like kord, see [here](https://kotlin.github.io/dokka/1.7.10/user_guide/gradle/usage/#multi-module-projects))
- change branch for source link (was `master` before which is no longer the default branch)
- add `externalDocumentationLink()` for kotlinx.coroutines, kotlinx.serialization and ktor
- exclude `TweetNaclFast` from documentation
- [exclude](https://kotlinlang.org/docs/kotlin-doc.html#suppress) deprecated declarations with level `ERROR` from documentation - level `HIDDEN` is excluded by default, level `WARNING` should probably not be insta-removed

This PR also includes two fixes unrelated to Dokka:
- fix package for `MessageInteractionData` (was `cache.data` but should be `dev.kord.core.cache.data`)
- use `Library.projectUrl` instead of `Library.description` for `url` in pom for publishing, currently this can lead to something like [this](https://search.maven.org/artifact/dev.kord/kord-core/0.8.0-M15/jar): 
grafik


buildSrc/src/main/kotlin/kord-module.gradle.kts
```diff
@@ -57,21 +59,31 @@ tasks {
...
+        dokkaSourceSets.configureEach {
 
-                    remoteLineSuffix.set("#L")
-                }
+            jdkVersion.set(Jvm.targetInt)
 
-                jdkVersion.set(8)
+            sourceLink {
+                localDirectory.set(file("src/main/kotlin"))
+                remoteUrl.set(URL("https://github.com/kordlib/kord/blob/${Library.commitHashOrDefault("0.8.x")}/${project.name}/src/main/kotlin/"))
```
Lukellmann, line 75: I don't like that the default branch is hard-coded to `0.8.x` here, it was hard-coded to `master` before which became outdated some time ago.
Any ideas on how this can be avoided?

common/src/main/kotlin/entity/DiscordMessage.kt
```diff
@@ -790,22 +790,27 @@ public sealed class MessageType(public val code: Int) {
...
     public object GuildBoostTier1 : MessageType(9)
 
+    /** @suppress */
     @Deprecated("Renamed to 'GuildBoostTier2'.", ReplaceWith("GuildBoostTier2"))
     public object UserPremiumGuildSubscriptionTwo : MessageType(10)
     public object GuildBoostTier2 : MessageType(10)
 
+    /** @suppress */
     @Deprecated("Renamed to 'GuildBoostTier3'.", ReplaceWith("GuildBoostTier3"))
     public object UserPremiumGuildSubscriptionThree : MessageType(11)
```
Lukellmann, line 815: I suppressed these even though they only have level `WARNING`, because they are simple renames.

buildSrc/src/main/kotlin/kord-module.gradle.kts
DRSchlaubi: Maybe just use the current branch?

Lukellmann: But that's also just available in the GitHub CI (where we have the commit hash too), right?

Lukellmann: Also turns out that using the commit hash breaks source links for https://dokka.kord.dev as the github actions that generate the docs use the commit hash of that repo and not kord's commit hash.

PR #665, opened 2022-08-09
Breaking: `Event` is no longer a `CoroutineScope`

Lukellmann:
# Problem

Currently, the `dev.kord.core.event.Event` interface extends `CoroutineScope`, but this is problematic.

The reason is that events don't have a well-defined lifecycle, as we never call `CoroutineScope.cancel()` for them (and there is no obvious place to do so). The [documentation](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/) even highlights that the "**key part of custom usage of `CoroutineScope` is cancelling it at the end of the lifecycle**".

This wouldn't be too much of a problem in itself, but it leads to a memory leak, as @Stonedestroyer discovered.

### Memory Leak Details

The way we create the `CoroutineScope` instances for use by `Event`s (via delegation) is the following:
```kotlin
fun kordCoroutineScope(kord: Kord) = CoroutineScope(kord.coroutineContext + SupervisorJob(kord.coroutineContext.job))
```

This means that we create a parent-child hierarchy between the `Job`s of the scope of `Kord` (parent) and the newly created scope of `Event` (child).

The hierarchy has a relevant impact: the parent now holds a reference to the child. When the child is cancelled, this reference is released so that the child can be garbage collected.
But we actually never cancel the child, which results in a memory leak as the parent keeps accumulating references to children that will never be released.

It can even happen that multiple scopes will leak for a single event, as [this for loop](https://github.com/kordlib/kord/blob/8b875195b7225e410bbb916c9bc5309ab0629d5b/core/src/main/kotlin/gateway/handler/DefaultGatewayEventInterceptor.kt#L33) creates a new scope with the described parent-child hierarchy for every listener it tries to call.

See also [this thread](https://discord.com/channels/556525343595298817/1005983735746936913) on the Kord Discord Server.

# Solution

As a solution, this PR removes `CoroutineScope` as a supertype of `Event`.

This is a breaking change, however there are easy replacements for cases where `Event` was used as a `CoroutineScope`:
- for coroutines that need a local scope, the [`coroutineScope { ... }`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html) scoping function can be used, it "is designed for _parallel decomposition_ of work" and enforces structured concurrency
- for coroutines that need a scope which is more global, the `Kord` instance can be used, it is also a `CoroutineScope` but one that has a well-defined lifecycle (see `Kord.shutdown()`)

# Migration

Code like this
```kotlin
kord.on<SomeEvent> {
    launch { /* ... */ } // launched in event scope
}
```
can be replaced relatively easy with
```kotlin
kord.on<SomeEvent> {
    coroutineScope {
        launch { /* ... */ } // for coroutines that need a local scope, coroutineScope {} waits for its children
    }
    // or
    kord.launch { /* ... */ } // for coroutines that need a more global scope
}
```

If you used the `CoroutineScope` of `Event` as a way to associate a custom context to an event, take a look at #667.

### DIFF
--- a/core/src/main/kotlin/gateway/handler/BaseGatewayEventHandler.kt
+++ b/core/src/main/kotlin/gateway/handler/BaseGatewayEventHandler.kt
@@ -2,8 +2,6 @@ package dev.kord.core.gateway.handler
 
 import dev.kord.cache.api.DataCache
 import dev.kord.core.Kord
-import kotlinx.coroutines.CoroutineScope
-import kotlin.coroutines.CoroutineContext
 import dev.kord.core.event.Event as CoreEvent
 import dev.kord.gateway.Event as GatewayEvent
 
@@ -11,6 +9,6 @@ public abstract class BaseGatewayEventHandler(
     protected val cache: DataCache
 ) {
 
-    public abstract suspend fun handle(event: GatewayEvent, shard: Int, kord: Kord, coroutineScope: CoroutineScope): CoreEvent?
+    public abstract suspend fun handle(event: GatewayEvent, shard: Int, kord: Kord): CoreEvent?
 
 }
--- a/core/src/main/kotlin/gateway/handler/DefaultGatewayEventInterceptor.kt
+++ b/core/src/main/kotlin/gateway/handler/DefaultGatewayEventInterceptor.kt
@@ -2,19 +2,14 @@ package dev.kord.core.gateway.handler
 
 import dev.kord.cache.api.DataCache
 import dev.kord.core.Kord
-import dev.kord.core.event.kordCoroutineScope
 import dev.kord.core.gateway.ShardEvent
 import io.ktor.util.logging.*
-import kotlinx.coroutines.CoroutineScope
 import mu.KotlinLogging
 import dev.kord.core.event.Event as CoreEvent
 
 private val logger = KotlinLogging.logger { }
 
-public class DefaultGatewayEventInterceptor(
-    cache: DataCache,
-    private val eventScope: ((ShardEvent, Kord) -> CoroutineScope) = { _, kord -> kordCoroutineScope(kord) }
-) : GatewayEventInterceptor {
+public class DefaultGatewayEventInterceptor(cache: DataCache) : GatewayEventInterceptor {
 
     private val listeners = listOf(
         MessageEventHandler(cache),
@@ -31,12 +26,7 @@ public class DefaultGatewayEventInterceptor(
     override suspend fun handle(event: ShardEvent, kord: Kord): CoreEvent? {
         return runCatching {
             for (listener in listeners) {
-                val coreEvent = listener.handle(
-                    event.event,
-                    event.shard,
-                    kord,
-                    eventScope.invoke(event, kord)
-                )
+                val coreEvent = listener.handle(event.event, event.shard, kord)
                 if (coreEvent != null) {
                     return coreEvent
                 }
--- a/core/src/main/kotlin/Kord.kt
+++ b/core/src/main/kotlin/Kord.kt
@@ -631,6 +631,6 @@ public inline fun <reified T : Event> Kord.on(
 ): Job =
     events.buffer(CoroutineChannel.UNLIMITED).filterIsInstance<T>()
         .onEach { event ->
-            scope.launch(event.coroutineContext) { runCatching { consumer(event) }.onFailure { kordLogger.catching(it) } }
+            scope.launch { runCatching { consumer(event) }.onFailure { kordLogger.catching(it) } }
         }
         .launchIn(scope)
--- a/core/src/main/kotlin/event/Event.kt
+++ b/core/src/main/kotlin/event/Event.kt
@@ -2,12 +2,8 @@ package dev.kord.core.event
 
 import dev.kord.core.Kord
 import dev.kord.gateway.Gateway
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.job
-import kotlin.coroutines.CoroutineContext
 
-public interface Event : CoroutineScope {
+public interface Event {
     /**
      * The Gateway that spawned this event.
      */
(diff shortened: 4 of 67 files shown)


core/src/main/kotlin/gateway/handler/BaseGatewayEventHandler.kt
```diff
@@ -2,15 +2,13 @@ package dev.kord.core.gateway.handler
...
-import kotlin.coroutines.CoroutineContext
 import dev.kord.core.event.Event as CoreEvent
 import dev.kord.gateway.Event as GatewayEvent
 
 public abstract class BaseGatewayEventHandler(
     protected val cache: DataCache
 ) {
 
-    public abstract suspend fun handle(event: GatewayEvent, shard: Int, kord: Kord, coroutineScope: CoroutineScope): CoreEvent?
+    public abstract suspend fun handle(event: GatewayEvent, shard: Int, kord: Kord): CoreEvent?
```
Lukellmann, line 12: A handler no longer needs a `CoroutineScope`.

core/src/main/kotlin/gateway/handler/DefaultGatewayEventInterceptor.kt
```diff
@@ -2,19 +2,14 @@ package dev.kord.core.gateway.handler
...
 import mu.KotlinLogging
 import dev.kord.core.event.Event as CoreEvent
 
 private val logger = KotlinLogging.logger { }
 
-public class DefaultGatewayEventInterceptor(
-    cache: DataCache,
-    private val eventScope: ((ShardEvent, Kord) -> CoroutineScope) = { _, kord -> kordCoroutineScope(kord) }
-) : GatewayEventInterceptor {
+public class DefaultGatewayEventInterceptor(cache: DataCache) : GatewayEventInterceptor {
```
Lukellmann, line 12: The interceptor no longer needs to construct `CoroutineScope`s.

HopeBaron (approved):

gdude2002:
It's possible that this will be an irreversibly breaking change for KordEx, as our event handling launches event handlers in the coroutine scope provided by the event.

This might seem a bit odd at first, but because Kord doesn't provide a way to add or modify contextual event data, this was the hack that @ByteAlex used to make KordEx even close to viable with his Dispers gateway.

Before this PR is merged, I would prefer to see ways to make Kord objects viable without logging in, and ways to intercept and modify events, as has been promised by the team for months.

HopeBaron:
I've been short on time lately to look into anything that would keep the sealed nature of the events while allowing for extensiblity of the events.
Essentially we could assume that we can open the event's tree AKA open the event classes
while keeping its inherited properties on our side closed.

As per my last comment on Discord it seems to be that we will have to open the events some day 
However I believe this would've lots of disadvantages down the line.
CC @BartArys @Lukellmann

gdude2002:
One way to band-aid events for people with frameworks would be to assign a UUID to each event. Another (potentially more accessible) option would be to add a property which is a `MutableMap<String, Any>`, since that would be removed at the end of the event's lifecycle - that's what KordEx did for its check contexts and, while it feels like a bit of a hack, it's been pretty invaluable for some users.

You can see precisely what that entailed [here](https://github.com/Kord-Extensions/kord-extensions/blob/develop/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Maps.kt#L22).

Stonedestroyer:
After 6 hours of testing with an example bot linked at https://github.com/Stonedestroyer/kord-example on a token with 3500 servers I can confirm that the coroutines are getting garbage cleaned and the bot is staying at a consistent 1.7 GB RAM compared to before where it was spiking always.

HopeBaron:
> After 6 hours of testing with an example bot linked at https://github.com/Stonedestroyer/kord-example on a token with 3500 servers I can confirm that the coroutines are getting garbage cleaned and the bot is staying at a consistent 1.7 GB RAM compared to before where it was spiking always.

That's nice to hear

HopeBaron:
> One way to band-aid events for people with frameworks would be to assign a UUID to each event. Another (potentially more accessible) option would be to add a property which is a `MutableMap<String, Any>`, since that would be removed at the end of the event's lifecycle - that's what KordEx did for its check contexts and, while it feels like a bit of a hack, it's been pretty invaluable for some users.
> 
> You can see precisely what that entailed [here](https://github.com/Kord-Extensions/kord-extensions/blob/develop/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Maps.kt#L22).

You are suggesting to add a map like so 
```kt
interface Event { 
  ...
  val extras: Map<String, Any>
}
```
But how would the user add these extras

gdude2002:
No, I'm suggesting a MutableMap, not just a Map.

Getters and setters can be added via extension functions.

HopeBaron:
I've modfied the example just before you commented
Tho in a real scenario wouldn't you wrap those `extras` into another objects later on?

gdude2002:
Not really - this is really something that Kotlin doesn't have a good solution for, so I don't see why a set of extension functions for get/set and early handling of the event to add data wouldn't work well enough. You don't get compiler-enforced safety for the cast, but that ultimately doesn't matter with some well-written functions.

Lukellmann:
> This might seem a bit odd at first, but because Kord doesn't provide a way to add or modify contextual event data, this was the hack that @ByteAlex used to make KordEx even close to viable with his Dispers gateway.

Can you explain this hack a bit more? I'd like to understand the problem better.

gdude2002:
So, as you may have heard, the really big bots don't actually connect directly to Discord. Most of them use some kind of gateway proxy or load balancing for command processing, which ultimately means that bots don't have access to an actual Discord token. For this reason, those bots often need to attach extra context to an event before the command handlers see it - for example, to pass in data that the command may need, but that isn't provided in the event, since the bot has no way of requesting it later.

This was the primary reason @ByteAlex suggested the way KordEx now handles coroutine contexts for events - there are (even more hacky) ways to associate data with a specific coroutine context, so he was assigning a new context to every event he was passing into KordEx, in order to store and later retrieve that data.

Lukellmann:
> One way to band-aid events for people with frameworks would be to assign a UUID to each event.

Wouldn't the object identity of the event serve the same purpose?

Lukellmann:
My last 2 commits made the diff clearer, reviews should be easier now.

PR #666, opened 2022-08-11
Use gateway url provided in `Ready` event for resuming

Lukellmann:
relevant doc commits:
- https://github.com/discord/discord-api-docs/commit/30b37e576d3457f5401c50dea96e8bc5bfdd675b
- https://github.com/discord/discord-api-docs/commit/58d6ec4ba99018c2fc104428c0309081b14d9fb3


gateway/src/main/kotlin/handler/HandshakeHandler.kt
```diff
@@ -19,31 +19,37 @@ internal class HandshakeHandler(
...
+                .apply { parameters.appendMissing(initialUrl.parameters) }
+                .build()
+
+            resumeContext.value = ResumeContext(event.data.sessionId, resumeUrl)
+        }
+
         on<Close.SessionReset> {
-            session.update { null }
+            resumeContext.value = null
         }
```
Lukellmann, line 53: I think it might be necessary to reset the context on some other `Close` events, e.g. `UserClose` (resuming after the user closed the gateway seems wrong). I'm new to gateway code, so I might be wrong here, what do you think?

gateway/src/main/kotlin/handler/HandshakeHandler.kt
```diff
@@ -19,31 +19,37 @@ internal class HandshakeHandler(
...
 
+    override fun start() {
         on<Hello> {
-            reconnectRetry.reset() //connected and read without problems, resetting retry counter
+            reconnectRetry.reset() // connected and read without problems, resetting retry counter
             identifyRateLimiter.consume {
-                if (sessionStart) send(identify)
-                else send(resume)
+                send(resumeOrIdentify)
             }
```
Lukellmann, line 39: Do we also need to rate-limit resumes?

[This](https://discord.com/developers/docs/topics/gateway#rate-limiting) suggests we do not:
> Clients also have a limit of [concurrent](https://discord.com/developers/docs/topics/gateway#session-start-limit-object) [Identify](https://discord.com/developers/docs/topics/gateway#identify) requests allowed per 5 seconds. If you hit this limit, the Gateway will respond with an [Opcode 9 Invalid Session](https://discord.com/developers/docs/topics/gateway#invalid-session).

And there is also [this](https://discord.com/developers/docs/topics/gateway#identifying):
> Clients are limited to 1000 `IDENTIFY` calls to the websocket in a 24-hour period. This limit is global and across all shards, but does not include `RESUME` calls.

gateway/src/main/kotlin/handler/HandshakeHandler.kt
HopeBaron: I do assume that this is for the resumeGatewayUrl new property but why not just store that new URL in place of the old gateway url

Lukellmann: we need to keep the old one because we have to connect to it when we don't resume a session but instead start a new one

HopeBaron: Oh, this looks fine then

HopeBaron:
I reviewed the PR a second time and it looks fine; but it's a draft so I'll be waiting til it's complete

Lukellmann:
I made it a draft because of this concern:
> I think it might be necessary to reset the context on some other `Close` events, e.g. `UserClose` (resuming after the user stopped the gateway seems wrong). I'm new to gateway code, so I might be wrong here, what do you think?

But if say this isn't needed, then this PR is ready.

HopeBaron:
The resume will not need to modify anything because the reset or not is handled by the loop conditions
all what changed in this PR is where should we connect on resumes not how we should connect so I guess we are good to go if you have tested it and left the bot running for a bit with, say "turn on turn off network to test how would that go

Lukellmann:
Problem is that I can't really test rn because they only send the default url until September 12 (which means we connect to the same as normal).

HopeBaron:
In that case guess there is no need to rush and merge as this may be subject to change

Lukellmann:
Converted this back to a draft for now

Lukellmann:
I checked rn, I'm still not receiving a session-specific URL. Will try again later today/tomorrow.

Lukellmann:
> I checked rn, I'm still not receiving a session-specific URL. Will try again later today/tomorrow.

Still no luck... 😞

gateway/src/main/kotlin/handler/HandshakeHandler.kt
Lukellmann: out of scope for this PR

Lukellmann:
Still nothing, what should we do?

Lukellmann:
I was finally able to receive a different url and test it - seems to work as intended

gateway/src/main/kotlin/handler/HandshakeHandler.kt
Lukellmann: In fact this is solved by #693

PR #670, opened 2022-08-15
Fix MessageStickerType constructor initialization deadlock

### LINKED ISSUE #669: `MessageStickerType` Initialization Deadlock
Sometimes, randomly, my application threads gets stuck on `waiting on the Class initialization monitor for dev.kord.common.entity.MessageStickerType`, which seems to indicate that there is a deadlock.

The application is deserializing Gateway Events on multiple threads.

Thread dump of the application: [kord-messagestickertype-deadlock-jstack.txt](https://github.com/kordlib/kord/files/9339926/kord-messagestickertype-deadlock-jstack.txt)

Example code that triggers the bug:
```kotlin
fun main() {
    val a = thread {
        println(MessageStickerType)
    }

    val b = thread {
        println(MessageStickerType.APNG)
    }
}
```

MrPowerGamerBR:
Instead of initializing the set during class load, the set will be created when calling the `values` list.

Fixes #669

Other Kord classes also handles `values` lists in the same manner (`Image`, `DiscordChannel`, `DiscordMessage`, etc), so I've decided to keep it in the same way.

However, Lost and Lukellmann pointed out that using `by lazy` also fixes the issue, and I have tested it and it also does work, so maybe this is something to keep in mind for the future to avoid unnecessary set allocations.

### DIFF
--- a/common/src/main/kotlin/entity/DiscordMessage.kt
+++ b/common/src/main/kotlin/entity/DiscordMessage.kt
@@ -171,7 +171,8 @@ public sealed class MessageStickerType(public val value: Int) {
     public object LOTTIE : MessageStickerType(3)
 
     public companion object {
-        public val values: Set<MessageStickerType> = setOf(PNG, APNG, LOTTIE)
+        public val values: Set<MessageStickerType>
+            get() = setOf(PNG, APNG, LOTTIE)
     }
 
     internal object Serializer : KSerializer<MessageStickerType> {


Lukellmann (approved):

Lukellmann:
> However, Lost and Lukellmann pointed out that using `by lazy` also fixes the issue, and I have tested it and it also does work, so maybe this is something to keep in mind for the future to avoid unnecessary set allocations.

I already considered this, see [here](https://github.com/kordlib/kord/blob/4ef0576b2011ca15f0bdd689d372c1bef5e3c7bf/ksp-processors/src/main/kotlin/kordenum/KordEnumGeneration.kt#L215) :)

PR #671, opened 2022-08-16
Don't log token when sending Resume

Lukellmann:
`DefaultGateway` logs the JSON payloads it sends when trace logging is enabled. It also hides the bot token in case of an `Identify` command. But it doesn't do this for `Resume` commands, which also includes the bot token.

This PR fixes this and also updates some `toString` implementations of data classes that contain a bot token.

It also makes the `Json` instance used by `DefaultGateway` a property of the companion object since it can be shared safely.


gateway/src/main/kotlin/Command.kt
```diff
@@ -80,7 +80,7 @@ public sealed class Command {
 
 @Serializable
 internal data class Identify(
-    internal val token: String,
```
Lukellmann, line 83: reason: this doesn't provide any safety, internal visibility becomes `public` on jvm anyway

Lukellmann:
Should we do this for voice gateway too @lost-illusi0n?

gateway/src/main/kotlin/Command.kt
```diff
@@ -80,7 +80,7 @@ public sealed class Command {
...
+    val token: String,
```
Lukellmann, line 83: reason: this doesn't provide any safety, internal visibility becomes `public` on jvm anyway

lost-illusi0n:
I'm not against it. Might as well be consistent across the entirety of Kord.

Lukellmann:
Alright, just asking cause I don't know how sensitive voice tokens are.

lost-illusi0n:
Probably sensitive enough, it is a token after all. I believe someone with access to it would be able to hijack the voice connection for a server (within a certain timeframe), though I haven't tried it :p

gateway/src/main/kotlin/DefaultGateway.kt
```diff
@@ -187,8 +181,37 @@ public class DefaultGateway(private val data: DefaultGatewayData) : Gateway {
...
+                            }
+                            else -> json
+                        }
+                    }
+
+                    else -> json
+                }
+
+                "Gateway <<< $credentialFreeJson"
+            }
```
Lukellmann, line 211: This is quite heavy for just not logging the token, voice does it like [this](https://github.com/kordlib/kord/blob/bebfcc74efae44d2e578b6d90076f1265eb9e049/voice/src/main/kotlin/gateway/DefaultVoiceGateway.kt#L149) to hide [`secret_key`](https://discord.com/developers/docs/topics/voice-connections#establishing-a-voice-udp-connection-example-session-description-payload), should we do it like that too instead?

Lukellmann:
Other token types this PR didn't touch yet are interaction and webhook tokens, what about them?

MrPowerGamerBR:
> Other token types this PR didn't touch yet are interaction and webhook tokens, what about them?

I think that, for consistency, it would be nice to hide them too, because they are sensitive too, because with interaction/webhook tokens you can send whatever message you want.

By the way, wouldn't it be better to filter the sensitive data on the logger end, instead of replacing all entities with a custom `toString()` implementation? While replacing the `toString()` implemention works, someone could forget to update it if new variables are added to the entities. Filtering on the logger side would make it easier to filter in other places were the token might get accidentally printed to the console, however you do have the disadvantage that it may filter something that it shouldn't be filtered. https://www.baeldung.com/logback-mask-sensitive-data

Lukellmann:
Hm, how would we filter it on the logger side without a logger implementation (kord doesn't provide one)?

MrPowerGamerBR:
> Hm, how would we filter it on the logger end without a logger implementation (kord doesn't provide one)?

Yeah, that's one of the issues with this approach: The user would need to configure their logger implementation (logback, log4j, etc) to filter tokens

lost-illusi0n:
Not quite, Kord depends on kotlin-logging for logging, which simply wraps around the SLF4J api. I'm aware of the log4j api having an AbstractLogger class that can be implemented, which would then strip out the token(s) from any outgoing log message, and delegate it to the actual logger. I assume this can similarly be done with the [slf4j abstract logger](https://www.slf4j.org/api/org/slf4j/helpers/AbstractLogger.html). Might be worth taking a look at.

Lukellmann:
Mentioned trace level in title to not scare people.

gateway/src/main/kotlin/Command.kt
lukellmann: changed already in #730

PR #672, opened 2022-08-17
Update dependencies and binary compatibility validator workflow

Lukellmann:
- Binary compatibility validator 0.10.1 -> 0.11.0
- AtomicFU 0.18.2 -> 0.18.3
- Ktor 2.0.3 -> 2.1.0
- kotlinx.coroutines 1.6.3 -> 1.6.4
- JUnit 5 5.8.2 -> 5.9.0
- MockK 1.12.4 -> 1.12.5
- Change workflow for binary compatibility validator like recommended [here](https://github.com/Kotlin/binary-compatibility-validator#workflow).
We will now always have to run `apiDump` when we change public API, otherwise tests will fail. This will keep the diff in `.api` files in the same commit as the actual changes. It ensures we are aware that we changed something and will prevent unwanted breaking changes by mistake. (Previously, we only dumped API before a release, so it was hard to track which commits changed which part of the API.)
This is also the normal way the validator is used by other libs (kxser, kxcoro, ...) so new contributors that worked with it before will not have to use it differently than they are used to.
(This PR also updates api dumps to let tests pass.)

### DIFF
--- a/.github/workflows/deployment-ci.yml
+++ b/.github/workflows/deployment-ci.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Build with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: --stacktrace --info build -x apiCheck
+          arguments: --stacktrace --info build
 
 
   release:
@@ -74,14 +74,7 @@ jobs:
       - name: Build with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: --stacktrace --info build -x apiCheck
-
-      - name: API Checking with Gradle
-        uses: gradle/gradle-build-action@v2
-        with:
-          arguments: apiCheck
-
-
+          arguments: --stacktrace --info build
 
       - name: Publish with Gradle
         uses: gradle/gradle-build-action@v2
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,7 +4,7 @@ import org.gradle.api.tasks.wrapper.Wrapper.DistributionType.ALL
 plugins {
     kotlin("jvm")
     kotlin("plugin.serialization")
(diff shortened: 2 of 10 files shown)


.github/workflows/deployment-ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Build with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: --stacktrace --info build -x apiCheck
```
HopeBaron, line 40: This will cause the apiCheck to happen on snapshots which we don''t want

.github/workflows/deployment-ci.yml
```diff
@@ -74,14 +74,7 @@ jobs:
...
```
HopeBaron, line 77: we want apiCheck to happen on releases

HopeBaron (changes requested):

.github/workflows/deployment-ci.yml
Lukellmann: I know, and I wanted to change this.

We would now always have to update the api dump when we change api, which will tie the diff in .api files with the actual change to the same commit. This will always make us aware that we changed something and will prevent unwanted breaking changes by mistake.

This is also the normal way the validator is used by other libs (kxser, kxcoro, ...) so new contributors that worked with it before will not have to use it differently than they are used to.

.github/workflows/deployment-ci.yml
Lukellmann: it will always be executed, it's part of the `check` pipeline, see https://github.com/Kotlin/binary-compatibility-validator#tasks

.github/workflows/deployment-ci.yml
Lukellmann: I update the PR description to better state the reason for this change.

PR #673, opened 2022-08-20
Use GitHub Pages Deploy Action

Lukellmann:
This PR replaces the [gradle-git-publish](https://github.com/ajoberstar/gradle-git-publish) plugin with [GitHub Pages Deploy Action](https://github.com/marketplace/actions/deploy-to-github-pages).

Since GitHub Pages Deploy Action uses the always available [`GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication), we don't longer need to provide a personal access token for authentication. If this gets merged, the `GRGIT_USER` secret can be removed.
This also fixes the issue that we couldn't upgrade the gradle-git-publish plugin because its newer versions are built for jdk 11 while kord is still using jdk 8.

This was inspired by @DRSchlaubi's [kord-dokka](https://github.com/DRSchlaubi/kord-dokka) repo.

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -1,10 +1,10 @@
-# This workflow will upload docs to the gh-pages branch whenever something is pushed to 0.8.x.
+# This workflow will build and push docs to the gh-pages branch whenever something is pushed to 0.8.x.
 
 name: Docs
 on:
   push:
     branches:
-      - docs # todo set back to 0.8.x after testing
+      - 0.8.x
 
 jobs:
   update_docs:
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -4,25 +4,28 @@ name: Docs
 on:
   push:
     branches:
-      - 0.8.x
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-ci.yml
```diff
@@ -7,22 +7,32 @@ on:
...
-    name: update docs
+  update_docs:
+    name: Update docs
     runs-on: ubuntu-latest
+
     env:
-      GRGIT_USER: ${{ secrets.GRGIT_USER }}
-      ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+      GITHUB_TAG_NAME: ${{ github.event.release.tag_name }}
+      GITHUB_BRANCH_NAME: ${{ github.ref }}
```
Lukellmann, line 16: Providing these environment variables will show the version the docs were built from in the dokka header (they are queried [here](https://github.com/kordlib/kord/blob/b5d5dc82f85d145741d71629e179cba1125592db/buildSrc/src/main/kotlin/Projects.kt#L9)):

dokka header

PR #675, opened 2022-08-29
Serialize unavailable guilds properly

### LINKED ISSUE #627: Kord fails to deserialize a `GuildCreate` event with a unavailable guild
```
Caused by: kotlinx.serialization.MissingFieldException: Fields [name, icon, owner_id, region, afk_channel_id, afk_timeout, verification_level, default_message_notifications, explicit_content_filter, roles, emojis, features, mfa_level, application_id, system_channel_id, system_channel_flags, rules_channel_id, vanity_url_code, description, banner, premium_tier, preferred_locale, public_updates_channel_id, nsfw_level, premium_progress_bar_enabled] are required for type with serial name 'dev.kord.common.entity.DiscordGuild', but they were missing at path: $.d
        at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:40)
        at kotlinx.serialization.encoding.AbstractDecoder.decodeSerializableValue(AbstractDecoder.kt:43)
        at kotlinx.serialization.encoding.AbstractDecoder.decodeSerializableElement(AbstractDecoder.kt:70)
        at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableElement(StreamingJsonDecoder.kt:111)
        at kotlinx.serialization.encoding.CompositeDecoder$DefaultImpls.decodeSerializableElement$default(Decoding.kt:537)
        at dev.kord.gateway.Event$DeserializationStrategy.getByDispatchEvent(Event.kt:167)
        at dev.kord.gateway.Event$DeserializationStrategy.deserialize(Event.kt:80)
        at dev.kord.gateway.Event$DeserializationStrategy.deserialize(Event.kt:48)
        at kotlinx.serialization.json.internal.PolymorphicKt.decodeSerializableValuePolymorphic(Polymorphic.kt:59)
        at kotlinx.serialization.json.internal.StreamingJsonDecoder.decodeSerializableValue(StreamingJsonDecoder.kt:38)
        ... 9 common frames omitted
```

Guild:
```json
{"d": {"id": "989200553198375033", "unavailable": true}, "s": 27007, "t": "GUILD_CREATE", "op": 0}
```

HopeBaron:
fixes #627


gateway/src/main/kotlin/Event.kt
```diff
@@ -591,7 +591,7 @@ public data class ChannelDelete(val channel: DiscordChannel, override val sequen
 public data class ChannelPinsUpdate(val pins: DiscordPinsUpdateData, override val sequence: Int?) : DispatchEvent()
 
 public data class TypingStart(val data: DiscordTyping, override val sequence: Int?) : DispatchEvent()
-public data class GuildCreate(val guild: DiscordGuild, override val sequence: Int?) : DispatchEvent()
+public data class GuildCreate(val guild: DiscordPartialGuild, override val sequence: Int?) : DispatchEvent()
```
Lukellmann, line 594: I'm not sure if we should use `DiscordPartialGuild` for this. This rather seems to be a union of `DiscordUnavailableGuild` and `DiscordGuild`. We might instead introduce a sealed superclass for these two and decide on whether to deserialize it as a `DiscordGuild` or a `DiscordUnavailableGuild` based on the presence of the `unavailable` field.

Lukellmann (changes requested):

gateway/src/main/kotlin/Event.kt
Lukellmann: So something like this:
```kotlin
// supertype for DiscordUnavailableGuild and DiscordGuild
@Serializable(with = DiscordBaseGuild.Serializer::class)
public sealed class DiscordBaseGuild {
    public abstract val id: Snowflake

    internal object Serializer : JsonContentPolymorphicSerializer<DiscordBaseGuild>(DiscordBaseGuild::class) {
        override fun selectDeserializer(element: JsonElement) =
            when (element.jsonObject["unavailable"]?.jsonPrimitive?.boolean) {
                true -> DiscordUnavailableGuild.serializer()
                false, null -> DiscordGuild.serializer()
            }
    }
}
```
`DiscordGuild.unavailable` could then also be removed.

Lukellmann: This at least works for the available case in the basic test I did (unavailable is hard to reproduce).

HopeBaron: do we even need a super type :thinking:

Lukellmann: I mean what should we put in the event class if there is no supertype? We don't have union types.

Lukellmann: this also supports my suggestion: https://github.com/discord/discord-api-docs/commit/745685b683dbcfbfe7e8818d25ed63d3eb6343f3

PR #686, opened 2022-09-11
Generate kord enums with KSP

Lukellmann:
This PR adds [KSP](https://kotlinlang.org/docs/ksp-overview.html) and [KotlinPoet](https://github.com/square/kotlinpoet) to the project and introduces an annotation to generate "kord enums" (classes like `ChannelType`). All these classes are now generated with this annotation.

We have a lot of sealed classes like `ChannelType` (representing [this](https://discord.com/developers/docs/resources/channel#channel-object-channel-types)) that are very similar. They all have:
- a single property that represents the value (the value discord actually uses)
- a serializer
- a set of object subtypes that represent all known values
- an Unknown subclass that we use for values that aren't documented / added yet

I gave them the name "kord enums" because they are essentially enums with the ability to represent unknown cases.

Every time we add a new one of these classes (e.g. when discord adds a new feature), we have to write a lot of boilerplate to make the class similar to the existing ones. But it's easy to bring in slight variations which leads to inconsistent code over time.
Adding new subtypes to these classes can also be error-prone, since it requires adding the new type and also adding it in the serializer which can be easy to forget (see https://github.com/kordlib/kord/commit/b75e33fd31f14f759f041ea94a452a87e38fbd38).

This PR introduces a KSP processor to solve these problems. It allows us to define these classes via annotations which reduces boilerplate, keeps consistency across these classes and ensures a single point of truth for the value (we had the value in the subtype declaration and serializer before).

Another benefit is that when we want to change something about all these classes, we just have to adjust the processor.

In the future we might use KSP to also generate other boilerplate heavy parts of kord (flags, builders etc.).

This PR also
- implements `equals`, `hashCode` and `toString` for every kord enum.
- adds `WebhookType.Application` (was missing before).
- replaces `DiscordGuildApplicationCommandPermission.Type` with `ApplicationCommandPermissionType` because of technical reasons (can't generate nested classes).
- removes deprecation from `EmbedType`.
- adds a new `entries` property in the companion object of every generated kord enum with type `List<KordEnumType>` that is a list of all known entries. The name is taken from Kotlin's upcoming [`Enum.entries`](https://github.com/Kotlin/KEEP/pull/284/files).
- fixes a bug that deleted public threads would produce `UnknownChannelThreadDeleteEvent` instead of `TextChannelThreadDeleteEvent`.


Lukellmann:
This is ready

HopeBaron:
does this mean that all Add x code are now redundant? 
I've not looked at the code yet tho

Lukellmann:
Add x code? What do you mean?

HopeBaron:
Error codes from other PRs

PR #691, opened 2022-09-15
Fix default avatar URL not working

### LINKED ISSUE #689: Default avatar not working
When the user hasn't set a default avatar and you are trying to get their default discord avatar using `member.defaultAvatar` this returns an invalid URL.

Root cause: PR #401 
This PR converted from by default using PNG for all avatars to default using WEBP. This is a good change in general and should remain like this for custom avatars.

But for some reason Discord decided not to make WEBP variants for their default avatars, as shown here:
https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints

Solution:
For `defaultAvatar` an exception has to be made so that it always returns PNG.

MrPowerGamerBR:
This PR changes that now the format must be provided when initializing the `Icon` class, instead of relying on the `animated` property.

Fixes #689

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -6401,8 +6401,7 @@ public final class dev/kord/core/entity/GuildWidget : dev/kord/core/KordObject, 
 }
 
 public abstract class dev/kord/core/entity/Icon : dev/kord/core/KordObject {
-	public synthetic fun <init> (ZLdev/kord/rest/route/CdnUrl;Ldev/kord/core/Kord;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun getAnimated ()Z
+	public synthetic fun <init> (Ldev/kord/rest/Image$Format;Ldev/kord/rest/route/CdnUrl;Ldev/kord/core/Kord;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getCdnUrl ()Ldev/kord/rest/route/CdnUrl;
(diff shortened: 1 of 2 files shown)


core/api/core.api
```diff
@@ -6401,8 +6401,7 @@ public final class dev/kord/core/entity/GuildWidget : dev/kord/core/KordObject,
 }
 
 public abstract class dev/kord/core/entity/Icon : dev/kord/core/KordObject {
-	public synthetic fun <init> (ZLdev/kord/rest/route/CdnUrl;Ldev/kord/core/Kord;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun getAnimated ()Z
```
Lukellmann, line 6405: Could you bring this back for binary compatibility? Not sure if it should be deprecated though.

Lukellmann (changes requested):

Lukellmann (approved):
thanks

MrPowerGamerBR:
I readded the `animated` property @Lukellmann. :3 

Originally I added it by just checking if the `Image.Format` is GIF, however, to future proof the code, I've decided to keep it in the constructor itself.

The reason why I made it that way: Stickers aren't supported as a `Icon` yet, but when they are supported, the code would need to be changed because animated stickers can be `LOTTIE` or `APNG` files, so both aren't GIFs.

And static stickers are also `PNG`! So a simple format check wouldn't have worked. (Discord Docs doesn't have a distinction between `PNG` and `APNG` in their CDN docs)

PR #692, opened 2022-09-17
Parameter for deleting old messages

dookon:


core/src/main/kotlin/behavior/channel/GuildMessageChannelBehavior.kt
```diff
@@ -23,9 +23,14 @@ public interface GuildMessageChannelBehavior : GuildChannelBehavior, MessageChan
      * Messages older than 14 days will be deleted individually.
      *
      * @param reason the reason showing up in the audit log
+     * @param deleteOlderMessages If the messages are from a week ago they will be deleted individually
```
Lukellmann, line 26: should be two weeks

Lukellmann (changes requested):
Three things:
- Could you put the parameter before `reason`, `reason` is the last parameter everywhere else?
- Could you make the default value `true` to get the same behavior as before when the parameter isn't passed?
- Could you provide a hidden deprecated adapter for binary compatibility like this:
```kotlin
public suspend fun bulkDelete(messages: Iterable<Snowflake>, reason: String? = null) = bulkDelete(messages, deleteOlderMessages = true, reason)
```

core/src/main/kotlin/behavior/channel/GuildMessageChannelBehavior.kt
dookon: Alright

core/src/main/kotlin/behavior/channel/GuildMessageChannelBehavior.kt
```diff
@@ -17,15 +17,31 @@ import kotlin.time.Duration.Companion.days
...
+    }
+
+    /**
+     * Requests to bulk delete the [messages].
+     * Messages older than 14 days will be deleted individually.
+     *
+     * @param reason the reason showing up in the audit log
+     * @param deleteOlderMessages whether messages from two weeks ago will be deleted
+     * @throws [RestRequestException] if something went wrong during the request.
+     */
```
Lukellmann, line 39:
```suggestion
    /**
     * Requests to bulk delete the [messages].
     *
     * If [deleteOlderMessages] is `true` (default), messages older than 14
     * days will be deleted individually. Otherwise they will be ignored.
     *
     * @param reason the reason showing up in the audit log
     * @throws [RestRequestException] if something went wrong during the request.
     */
```
Include docs for the param in the text.

Lukellmann (approved):
Thanks :)

dookon:
> Thanks :)

No problem haha, thanks for the suggestions

PR #693, opened 2022-09-20
Properly implement Identify rate limiting

### LINKED ISSUE #625: Properly implement login rate limiting
Currently we connect all shards at once and rate limit the identify command, which causes session resets and timeouts (see #624)

A better solution would be to implement rate limiting for logins

Lukellmann:
Previously Kord simply used `dev.kord.common.ratelimit.RateLimiter` to limit the rate at which Identify requests were sent to the Gateway.
However, this approach didn't wait for Gateway connections to fully start before allowing another connection to identify (it would simply wait 5 seconds instead) and didn't take the [additional requirements](https://discord.com/developers/docs/topics/gateway#sharding-max-concurrency) for multiple shards into account.
This faulty implementation of Identify rate limiting resulted in Invalid Session (opcode 9) events (see #625 and [this thread](https://discord.com/channels/556525343595298817/1021384687337353216)).

To solve this, the new `dev.kord.gateway.ratelimit.IdentifyRateLimiter` interface was introduced. The default implementation uses [`rate_limit_keys`](https://discord.com/developers/docs/topics/gateway#sharding-max-concurrency) to keep track of which shards are currently identifying and only clears them when the corresponding Ready event (or Invalid Session / close / timeout in the case of failure) is received.
An instance can be obtained by calling the `IdentifyRateLimiter` factory function.

`DefaultGateway` now also no longer opens WebSocket connections before the `IdentifyRateLimiter` has been consumed to further ensure rate limits are not exceeded.

see https://discord.com/developers/docs/topics/gateway#rate-limiting

fixes #625


core/src/main/kotlin/gateway/MasterGateway.kt
```diff
@@ -22,12 +29,44 @@ public interface MasterGateway {
...
+                    readyListeners.clear()
+                }
+
+                // make sure we don't miss the event by executing until first suspension point before starting gateway
+                readyListeners += launch(start = UNDISPATCHED) { gateway.events.first { it is Ready } }
+
+                val config = configuration.copy(shard = configuration.shard.copy(index = shardId))
+                launch { gateway.start(config) }
+
+                previousRateLimitKey = rateLimitKey
```
Lukellmann, line 66: Should we use this instead of `previousRateLimitKey`:
```kotlin
var previousRateLimitBucket = -1
// ...
    val rateLimitBucket = shardId / maxConcurrency

    if (rateLimitBucket > previousRateLimitBucket) {
        // ...
    }
    // ...
    previousRateLimitBucket = rateLimitBucket
```
This would only behave differently in a situation like [this](https://discord.com/channels/613425648685547541/697489244649816084/1021565107949551676).

DRSchlaubi:
I'll review this after testing it on my bot

core/src/main/kotlin/gateway/MasterGateway.kt
Lukellmann: For the test cases this would mean the `expectedBuckets` would change for ``fun `example 2 but without shards 15-30`()`` from `1` to `2` and for ``fun `randomly distributed shards`()`` from `3` to `4`

common/src/main/kotlin/ratelimit/ClockIntervalRateLimiter.kt
```diff
@@ -35,4 +35,6 @@ internal class ClockIntervalRateLimiter(
             else -> remainingConsumes -= 1
         }
     }
+
+    override fun toString() = "ClockIntervalRateLimiter(limit=$limit, interval=$interval, clock=$clock)"
```
Lukellmann, line 39: I've added these `toString()`s during debugging - might as well keep them

core/src/main/kotlin/gateway/MasterGateway.kt
Lukellmann: no longer under consideration

Lukellmann:
I think this is ready now, could you also take a look @MrPowerGamerBR?

PR #694, opened 2022-09-29
Add support for initializing a gateway connection with resume session data

MrPowerGamerBR:
Adds support for `resume_gateway_url`, the `resumeGatewayUrl` parameter is a `(String) -> (String)` instead of a `String` because the `resume_gateway_url` varies between sessions, so the user may want to override the gateway resume URL on a case by case basis (idk who would want to do that tho lmao)

This PR also adds support for initializing a gateway connection with resume session data, this allows you to store your session data somewhere (example: in a file) and, after restarting your bot, you can resume the session instead of reidentifying.

This is VERY useful for large bots, because you can have (almost) zero downtime without consuming `IDENTIFY`.

Here's an example of how it works:
```kotlin
suspend fun main() {
    val gateway = DefaultGateway {}

    var sessionId: String? = null
    var resumeGatewayUrl: String? = null
    var sequenceId: Int? = null

    gateway.events
        .filterIsInstance<Ready>()
        .onEach {
            sessionId = it.data.sessionId
            resumeGatewayUrl = it.data.resumeGatewayUrl
        }
        .launchIn(gateway)

    gateway.events
        .filterIsInstance<DispatchEvent>()
        .onEach {
            sequenceId = it.sequence
        }
        .launchIn(gateway)

    gateway.events
        .filterIsInstance<MessageCreate>()
        .filter { it.message.content.startsWith("!!reconnect") }
        .onEach {
            println("Stopping gateway connection...")
            gateway.stop(CloseReason.Codes.SERVICE_RESTART.code) // Using code 1000 (which is what Kord uses) cancels the current session, using service restart doesn't!
            println("Restarting gateway session $sessionId, using $resumeGatewayUrl and seq number $sequenceId")
            gateway.start("BotTokenHereYay") {
                resumeSession = ResumeSession(Session(sessionId!!, "$resumeGatewayUrl/?v=${KordConfiguration.GATEWAY_VERSION}&encoding=json&compress=zlib-stream"), sequenceId!!)

                @OptIn(PrivilegedIntent::class)
                intents = Intents.all
            }
        }
        .launchIn(gateway)

    gateway.start("BotTokenHereYay") {
        @OptIn(PrivilegedIntent::class)
        intents = Intents.all
    }

    awaitCancellation()
}
```

The API dump has a bunch of random stuff, idk why

### DIFF
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -91,6 +91,24 @@ public interface Gateway : CoroutineScope {
      */
     public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving")): GatewayResumeConfiguration
 
+    /**
+     * Closes the Gateway with code 1012, suspending until the underlying webSocket is closed.
+     *
+     * The session won't be invalidated, and you will be able to [resume] the gateway session. The session will be invalidated by Discord after a few minutes.
+     *
+     * @return the gateway session information, if there was any successful identify before closing the connection.
+     */
+    public suspend fun stopForResume(): GatewayResumeConfiguration = stop(WebSocketCloseReason(1012, "service restart"))
+
+    /**
+     * Closes the Gateway with code 1000, suspending until the underlying webSocket is closed.
+     *
+     * The session will be invalidated, and you won't be able to [resume] it. The bot will appear offline in the member list.
+     *
+     * @return the gateway session information, if there was any successful identify before closing the connection.
+     */
+    public suspend fun stopAndInvalidateSession(): GatewayResumeConfiguration = stop(WebSocketCloseReason(1000, "leaving"))
+
     public companion object {
         private object None : Gateway {
 
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -40,12 +40,29 @@ public interface Gateway : CoroutineScope {
     public val ping: StateFlow<Duration?>
 
     /**
-     * Starts a reconnection gateway connection with the given [configuration].
+     * Starts a reconnecting gateway connection with the given [configuration].
+     *
+     * After connecting, an identify request will be sent and a session will be created.
+     * If you want to resume an already started session, use [resume].
+     *
      * This function will suspend until the lifecycle of the gateway has ended.
      *
      * @param configuration the configuration for this gateway session.
      */
     public suspend fun start(configuration: GatewayConfiguration)
+
+    /**
+     * Starts a reconnecting gateway connection with the given [configuration].
+     *
+     * After connecting, a resume request will be sent and the session will be resumed if the session is valid.
+     * If the session was invalidated, an identify request will be sent and a session will be created.
+     * If you want to start a new session, use [start].
+     *
+     * This function will suspend until the lifecycle of the gateway has ended.
+     *
+     * @param configuration the configuration for this gateway session.
+     */
+    public suspend fun resume(configuration: GatewayResumeConfiguration)
 
     /**
      * Sends a [Command] to the gateway, suspending until the message has been sent.
@@ -64,9 +81,14 @@ public interface Gateway : CoroutineScope {
     public suspend fun detach()
 
     /**
-     * Closes the Gateway and ends the current session, suspending until the underlying webSocket is closed.
+     * Closes the Gateway, suspending until the underlying webSocket is closed.
+     *
+     * By default, the current gateway session will be closed. If you want to keep the session alive to reconnect later,
+     * change the [closeReason] to use a different [WebSocketCloseReason.code] that isn't 1000 or 1001.
+     *
+     * @param closeReason the close reason that will be used when stopping the WebSocket connection.
      */
-    public suspend fun stop()
+    public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving"))
 
     public companion object {
         private object None : Gateway {
@@ -83,7 +105,9 @@ public interface Gateway : CoroutineScope {
 
             override suspend fun start(configuration: GatewayConfiguration) {}
 
-            override suspend fun stop() {}
+            override suspend fun resume(configuration: GatewayResumeConfiguration) {}
+
+            override suspend fun stop(closeReason: WebSocketCloseReason) {}
 
             override suspend fun detach() {
                 (this as CoroutineScope).cancel()
@@ -112,6 +136,10 @@ public suspend inline fun Gateway.editPresence(builder: PresenceBuilder.() -> Un
 
 /**
  * Starts a reconnecting gateway connection with the given parameters.
+ *
+ * After connecting, an identify request will be sent and a session will be created.
+ * If you want to resume an already started session, use [resume].
+ *
  * This function will suspend until the lifecycle of the gateway has ended.
  *
  * ```kotlin
@@ -139,6 +167,52 @@ public suspend inline fun Gateway.start(token: String, config: GatewayConfigurat
     val builder = GatewayConfigurationBuilder(token)
     builder.apply(config)
     start(builder.build())
+}
+
+/**
+ * Starts a reconnecting gateway connection with the given parameters.
+ *
+ * After connecting, a resume request will be sent and the session will be resumed if the session is valid.
+ * If the session was invalidated, an identify request will be sent and a session will be created.
+ * If you want to start a new session, use [start].
+ *
+ * This function will suspend until the lifecycle of the gateway has ended.
+ *
+ * ```kotlin
+ * gateway.resume("your_token", "gatewaySessionId", "https://discord-resume-url-here.discord.com/", 0) {
+ *     shard = DiscordShard(0,1)
+ *
+ *     presence {
+ *         afk = false
+ *         status = Status.Online
+ *         watching("you :eyes:")
+ *     }
+ *
+ * }
+ *
+ * //gateway has disconnected
+ * ```
+ *
+ * @param token     The Discord token of the bot.
+ * @param sessionId The gateway session ID.
+ * @param resumeUrl The gateway session resume URL.
+ * @param sequence  The last received sequence number of this gateway session.
+ * @param config additional configuration for the gateway.
+ */
+public suspend inline fun Gateway.resume(token: String, sessionId: String, resumeUrl: String, sequence: Int, config: GatewayConfigurationBuilder.() -> Unit = {}) {
+    contract {
+        callsInPlace(config, InvocationKind.EXACTLY_ONCE)
+    }
+    val builder = GatewayConfigurationBuilder(token)
+    builder.apply(config)
+    resume(
+        GatewayResumeConfiguration(
+            sessionId,
+            resumeUrl,
+            sequence,
+            builder.build()
+        )
+    )
 }
 
 /**
@@ -222,6 +296,8 @@ public fun Gateway.requestGuildMembers(request: RequestGuildMembers): Flow<Guild
             return@transformWhile (it.data.chunkIndex + 1) < it.data.chunkCount
         }// 0 <= chunk_index < chunk_count
 }
+
+public data class WebSocketCloseReason(val code: Short, val message: String)
 
 /**
  * Enum representation of Discord's [Gateway close event codes](https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway-gateway-close-event-codes).
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -87,8 +87,9 @@ public interface Gateway : CoroutineScope {
      * change the [closeReason] to use a different [WebSocketCloseReason.code] that isn't 1000 or 1001.
      *
      * @param closeReason the close reason that will be used when stopping the WebSocket connection.
+     * @return the gateway session information, if there was any successful identify before closing the connection.
      */
-    public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving"))
+    public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving")): GatewaySession?
 
     public companion object {
         private object None : Gateway {
@@ -107,7 +108,7 @@ public interface Gateway : CoroutineScope {
 
             override suspend fun resume(configuration: GatewayResumeConfiguration) {}
 
-            override suspend fun stop(closeReason: WebSocketCloseReason) {}
+            override suspend fun stop(closeReason: WebSocketCloseReason): GatewaySession? { return null }
 
             override suspend fun detach() {
                 (this as CoroutineScope).cancel()
@@ -194,12 +195,10 @@ public suspend inline fun Gateway.start(token: String, config: GatewayConfigurat
  * ```
  *
  * @param token     The Discord token of the bot.
- * @param sessionId The gateway session ID.
- * @param resumeUrl The gateway session resume URL.
- * @param sequence  The last received sequence number of this gateway session.
+ * @param session   The gateway session information.
  * @param config additional configuration for the gateway.
  */
-public suspend inline fun Gateway.resume(token: String, sessionId: String, resumeUrl: String, sequence: Int, config: GatewayConfigurationBuilder.() -> Unit = {}) {
+public suspend inline fun Gateway.resume(token: String, session: GatewaySession, config: GatewayConfigurationBuilder.() -> Unit = {}) {
     contract {
         callsInPlace(config, InvocationKind.EXACTLY_ONCE)
     }
@@ -207,9 +206,7 @@ public suspend inline fun Gateway.resume(token: String, sessionId: String, resum
     builder.apply(config)
     resume(
         GatewayResumeConfiguration(
-            sessionId,
-            resumeUrl,
-            sequence,
+            session,
             builder.build()
         )
     )
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -89,7 +89,7 @@ public interface Gateway : CoroutineScope {
      * @param closeReason the close reason that will be used when stopping the WebSocket connection.
      * @return the gateway session information, if there was any successful identify before closing the connection.
      */
-    public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving")): GatewaySession?
+    public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving")): GatewayResumeConfiguration
 
     public companion object {
         private object None : Gateway {
@@ -108,7 +108,7 @@ public interface Gateway : CoroutineScope {
 
             override suspend fun resume(configuration: GatewayResumeConfiguration) {}
 
-            override suspend fun stop(closeReason: WebSocketCloseReason): GatewaySession? { return null }
+            override suspend fun stop(closeReason: WebSocketCloseReason): GatewayResumeConfiguration { error("Can't stop this!") }
 
             override suspend fun detach() {
                 (this as CoroutineScope).cancel()
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -80,6 +80,11 @@ public interface Gateway : CoroutineScope {
      */
     public suspend fun detach()
 
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    public suspend fun stop() {
+        stop(WebSocketCloseReason(1000, "leaving"))
+    }
+
     /**
      * Closes the Gateway, suspending until the underlying WebSocket is closed.
      *
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -54,7 +54,7 @@ public interface Gateway : CoroutineScope {
     /**
      * Starts a reconnecting gateway connection with the given [configuration].
      *
-     * After connecting, a resume request will be sent and the session will be resumed if the session is valid.
+     * After connecting, a [Resume] request will be sent and the session will be resumed if the session is valid.
      * If the session was invalidated, an identify request will be sent and a session will be created.
      * If you want to start a new session, use [start].
      *
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -156,7 +156,7 @@ public suspend inline fun Gateway.editPresence(builder: PresenceBuilder.() -> Un
 /**
  * Starts a reconnecting gateway connection with the given parameters.
  *
- * After connecting, an identify request will be sent and a session will be created.
+ * After connecting, an [Identify] request will be sent and a session will be created.
  * If you want to resume an already started session, use [resume].
  *
  * This function will suspend until the lifecycle of the gateway has ended.
--- a/gateway/src/main/kotlin/Gateway.kt
+++ b/gateway/src/main/kotlin/Gateway.kt
@@ -99,24 +99,6 @@ public interface Gateway : CoroutineScope {
      */
     public suspend fun stop(closeReason: WebSocketCloseReason = WebSocketCloseReason(1000, "leaving")): GatewayResumeConfiguration
 
-    /**
-     * Closes the Gateway with code 1012, suspending until the underlying webSocket is closed.
-     *
-     * The session won't be invalidated, and you will be able to [resume] the gateway session. The session will be invalidated by Discord after a few minutes.
-     *
-     * @return the gateway session information, if there was any successful identify before closing the connection.
-     */
-    public suspend fun stopForResume(): GatewayResumeConfiguration = stop(WebSocketCloseReason(1012, "service restart"))
-
-    /**
-     * Closes the Gateway with code 1000, suspending until the underlying webSocket is closed.
-     *
-     * The session will be invalidated, and you won't be able to [resume] it. The bot will appear offline in the member list.
-     *
-     * @return the gateway session information, if there was any successful identify before closing the connection.
-     */
-    public suspend fun stopAndInvalidateSession(): GatewayResumeConfiguration = stop(WebSocketCloseReason(1000, "leaving"))
-
     public companion object {
         private object None : Gateway {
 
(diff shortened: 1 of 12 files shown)


MrPowerGamerBR:
While implementing, I found a bug in the `DefaultGateway`: If you `.stop()` and then `.start()` the same gateway connection...

```kotlin
00:35:39.005 [DefaultDispatcher-worker-14] ERROR dev.kord.gateway.DefaultGateway - invalid distance too far back
java.util.zip.ZipException: invalid distance too far back
	at java.base/java.util.zip.InflaterOutputStream.write(InflaterOutputStream.java:267)
	at java.base/java.io.FilterOutputStream.write(FilterOutputStream.java:108)
	at dev.kord.gateway.DefaultGateway.deflateData(DefaultGateway.kt:184)
	at dev.kord.gateway.DefaultGateway.read(DefaultGateway.kt:194)
	at dev.kord.gateway.DefaultGateway.access$read(DefaultGateway.kt:65)
	at dev.kord.gateway.DefaultGateway$readSocket$2.emit(DefaultGateway.kt:174)
	at dev.kord.gateway.DefaultGateway$readSocket$2.emit(DefaultGateway.kt:171)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:62)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.access$emitAllImpl$FlowKt__ChannelsKt(Channels.kt:1)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Channels.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
	Suppressed: java.util.zip.ZipException: invalid distance too far back
		at java.base/java.util.zip.InflaterOutputStream.flush(InflaterOutputStream.java:169)
		at java.base/java.util.zip.InflaterOutputStream.finish(InflaterOutputStream.java:186)
		at java.base/java.util.zip.InflaterOutputStream.close(InflaterOutputStream.java:129)
		at kotlin.io.CloseableKt.closeFinally(Closeable.kt:60)
		at dev.kord.gateway.DefaultGateway.deflateData(DefaultGateway.kt:183)
		... 13 common frames omitted
```

While trying to investigate, it looks like the previous gateway connection is still active, even if we sent a `.close(...)` to it!

I changed the `readSocket` to
```kotlin
    private suspend fun readSocket() {
        val test = UUID.randomUUID()

        socket.incoming.asFlow().buffer(Channel.UNLIMITED).collect {
            println("UUID: $test $it")
            when (it) {
                is Frame.Binary, is Frame.Text -> read(it)
                else -> { /*ignore*/
                }
            }
        }
    }
```

``` 
UUID: c9d4ed82-504e-4be6-aebb-1b4d6311e0f8 Frame BINARY (fin=true, buffer len = 19)
Sequence ID: 503
Sequence ID: 504
UUID: c9d4ed82-504e-4be6-aebb-1b4d6311e0f8 Frame BINARY (fin=true, buffer len = 64)
Sequence ID: 505
UUID: c9d4ed82-504e-4be6-aebb-1b4d6311e0f8 Frame BINARY (fin=true, buffer len = 683)
UUID: c9d4ed82-504e-4be6-aebb-1b4d6311e0f8 Frame BINARY (fin=true, buffer len = 90)
Sequence ID: 506
Stopping gateway connection...
Sequence ID: 507
Restarting gateway session c703770c837e443edd07b3c7a6c8a306, using wss://gateway-us-east1-c.discord.gg and seq number 507
creating new inflater
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 114)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 15)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 134)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 33)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 49)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 571)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 115)
Sequence ID: 508
Sequence ID: 509
Sequence ID: 510
Sequence ID: 511
Sequence ID: 512
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 322)
Sequence ID: 513
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 180)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 132)
Sequence ID: 514
Sequence ID: 515
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 92)
Sequence ID: 516
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 103)
Sequence ID: 517
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 80)
Sequence ID: 518
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 511)
Sequence ID: 519
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 90)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 30)
Sequence ID: 520
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 82)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 223)
Sequence ID: 521
Sequence ID: 522
Sequence ID: 523
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 31)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 109)
Sequence ID: 524
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 56)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 97)
Sequence ID: 525
Sequence ID: 526
Sequence ID: 527
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 81)
Sequence ID: 528
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 54)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 218)
Sequence ID: 529
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 59)
Sequence ID: 530
Sequence ID: 531
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 34)
Sequence ID: 532
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 31)
Sequence ID: 533
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 129)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 58)
Sequence ID: 534
Sequence ID: 535
creating new inflater
UUID: bc8ecd8b-428d-42ef-bca1-aedad42732e2 Frame BINARY (fin=true, buffer len = 114)
UUID: 1d416030-a96d-42cf-b524-692df7a6aa0e Frame BINARY (fin=true, buffer len = 90)
00:39:15.949 [DefaultDispatcher-worker-14] ERROR dev.kord.gateway.DefaultGateway - invalid distance too far back
java.util.zip.ZipException: invalid distance too far back
	at java.base/java.util.zip.InflaterOutputStream.write(InflaterOutputStream.java:267)
	at java.base/java.io.FilterOutputStream.write(FilterOutputStream.java:108)
	at dev.kord.gateway.DefaultGateway.deflateData(DefaultGateway.kt:184)
	at dev.kord.gateway.DefaultGateway.read(DefaultGateway.kt:194)
	at dev.kord.gateway.DefaultGateway.access$read(DefaultGateway.kt:65)
	at dev.kord.gateway.DefaultGateway$readSocket$2.emit(DefaultGateway.kt:174)
	at dev.kord.gateway.DefaultGateway$readSocket$2.emit(DefaultGateway.kt:171)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:62)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.access$emitAllImpl$FlowKt__ChannelsKt(Channels.kt:1)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Channels.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
	Suppressed: java.util.zip.ZipException: invalid distance too far back
		at java.base/java.util.zip.InflaterOutputStream.flush(InflaterOutputStream.java:169)
		at java.base/java.util.zip.InflaterOutputStream.finish(InflaterOutputStream.java:186)
		at java.base/java.util.zip.InflaterOutputStream.close(InflaterOutputStream.java:129)
		at kotlin.io.CloseableKt.closeFinally(Closeable.kt:60)
		at dev.kord.gateway.DefaultGateway.deflateData(DefaultGateway.kt:183)
		... 13 common frames omitted
UUID: bc8ecd8b-428d-42ef-bca1-aedad42732e2 Frame BINARY (fin=true, buffer len = 15)
00:39:16.034 [DefaultDispatcher-worker-2] ERROR dev.kord.gateway.DefaultGateway - invalid distance too far back
java.util.zip.ZipException: invalid distance too far back
	at java.base/java.util.zip.InflaterOutputStream.write(InflaterOutputStream.java:267)
	at java.base/java.io.FilterOutputStream.write(FilterOutputStream.java:108)
	at dev.kord.gateway.DefaultGateway.deflateData(DefaultGateway.kt:184)
	at dev.kord.gateway.DefaultGateway.read(DefaultGateway.kt:194)
	at dev.kord.gateway.DefaultGateway.access$read(DefaultGateway.kt:65)
	at dev.kord.gateway.DefaultGateway$readSocket$2.emit(DefaultGateway.kt:174)
	at dev.kord.gateway.DefaultGateway$readSocket$2.emit(DefaultGateway.kt:171)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.emitAllImpl$FlowKt__ChannelsKt(Channels.kt:62)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt.access$emitAllImpl$FlowKt__ChannelsKt(Channels.kt:1)
	at kotlinx.coroutines.flow.FlowKt__ChannelsKt$emitAllImpl$1.invokeSuspend(Channels.kt)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
	Suppressed: java.util.zip.ZipException: invalid distance too far back
		at java.base/java.util.zip.InflaterOutputStream.flush(InflaterOutputStream.java:169)
		at java.base/java.util.zip.InflaterOutputStream.finish(InflaterOutputStream.java:186)
		at java.base/java.util.zip.InflaterOutputStream.close(InflaterOutputStream.java:129)
		at kotlin.io.CloseableKt.closeFinally(Closeable.kt:60)
		at dev.kord.gateway.DefaultGateway.deflateData(DefaultGateway.kt:183)
		... 13 common frames omitted
```

MrPowerGamerBR:
It makes sense why this happens, it is creating a new gateway connection because of `data.reconnectRetry.hasNext && state.value is State.Running`, and because I used `.start`, the state is changed to `Running`, so it starts the connection all over again PLUS a new connection due to the `.start` call.

To fix this, I made the `DefaultGateway` track what's the current restart ID and, if it is different, it won't try restarting the connection.

There's probably a better way to implement this, but hey, it works.

Lukellmann:
Two things:

#

Could you try to merge #666 into this since it already adds support for `resume_gateway_url`

#

I think instead of hacking this functionality into the existing methods, we should rather introduce a pair of new ones like this:
```kotlin
public interface Gateway : CoroutineScope {
    // ...

    public suspend fun stopAndResumeLater(): GatewayResumeConfiguration

    public suspend fun resume(configuration: GatewayResumeConfiguration)
}

@Serializable // for convenience, users might want to serialize this
public data class GatewayResumeConfiguration(
    val sessionId: String,
    val resumeUrl: String,
    val sequence: Int,
    val startConfiguration: GatewayConfiguration, // this is the same as the last one passed to start()
)
```
`resume` would behave exactly like `start` (suspend until the lifecycle is ended etc.) but initially sends resume instead of identify. Later identifies will then use `startConfiguration`.

This will also allow core to add support for this when handling multiple shards (similar methods should be added to `Kord` as well):
```kotlin
public interface MasterGateway {
    // ...

    public suspend fun stopAndResumeLaterAll(): Map<Int, GatewayResumeConfiguration> = coroutineScope {
        val results = Channel<Pair<Int, GatewayResumeConfiguration>>()

        for ((shardId, gateway) in gateways) {
            launch {
                val resumeConfig = gateway.stopAndResumeLater()
                results.send(shardId to resumeConfig)
            }
        }

        buildMap {
            repeat(gateways.size) {
                val (shardId, resumeConfig) = results.receive()
                put(shardId, resumeConfig)
            }
        }
    }

    public suspend fun resumeAll(configurations: Map<Int, GatewayResumeConfiguration>) {
        require(gateways.keys == configurations.keys) { "Shards don't match, expected ${configurations.keys}, got ${gateways.keys}" }
        return coroutineScope {
            for ((shardId, gateway) in gateways) {
                launch { gateway.resume(configurations[shardId]!!) }
            }
        }
}
```
This will simplify your original usage example significantly (code when using core would be similar):
```kotlin
suspend fun main() {
    val gateway = DefaultGateway {}

    gateway.events
        .filterIsInstance<MessageCreate>()
        .filter { it.message.content.startsWith("!!reconnect") }
        .onEach {
            println("Stopping gateway connection...")
            val resumeConfig = gateway.stopAndResumeLater()
            println("Restarting gateway session ${resumeConfig.sessionId}, using ${resumeConfig.resumeUrl} and seq number ${resumeConfig.sequence}")
            gateway.resume(resumeConfig)
        }
        .launchIn(gateway)

    // initial start
    gateway.start("BotTokenHereYay") {
        @OptIn(PrivilegedIntent::class)
        intents = Intents.all
    }

    awaitCancellation()
}
```

MrPowerGamerBR:
Didn't notice that there was already a PR open related to the `resume_gateway_url`, whoops!

I will do your suggestions later, my implementation was kinda hacky because I didn't want to change the code *too* much. In fact, my original idea was to implement only the resuming sessions part, because I wanted to resume a session if the bot restarted instead of reidentifying.

Lukellmann:
> The API dump has a bunch of random stuff, idk why

```
- ...Lkotlin...
+ ...SLkotlin...
```
is because the additional `Short` param you added (remember jvms `short` primitive is `S` in api dumps)

In the (data) classes were you added a property, the constructor, `copy` and `componentN` signatures changed accordingly.

`Gateway$DefaultImpls.stop$default` was added because of the default value for the new `Short` param.

Lukellmann:
#666 was merged, so you can now base this PR on it

MrPowerGamerBR:
Done @Lukellmann! Instead of implementing a `stopAndResumeLater` function, I thought it would've made more sense to make `stop` return the current gateway session data.

Lukellmann:
> Instead of implementing a `stopAndResumeLater` function, I thought it would've made more sense to make `stop` return the current gateway session data.

The problem I see with this approach is that you need to know about the [details of close codes](https://discord.com/developers/docs/topics/gateway#initiating-a-disconnect) so you pass a right one that allows resuming. I agree that `stopAndResumeLater` isn't the best name for that function, maybe something like `stopForResume` (or something even better you can come up with)?

Also why not directly return `GatewayResumeConfiguration` instead of `GatewaySession`? This could be directly passed to `resume` without recreating the whole config.

MrPowerGamerBR:
> Also why not directly return GatewayResumeConfiguration instead of GatewaySession? This could be directly passed to resume without recreating the whole config.

Because the session can be null, if you called `.stop` before a `Ready` event has been received.

Of course, in that case it could be a `GatewayResumeConfiguration` with null session parameters... Hmm, not a bad idea in handsight.

> The problem I see with this approach is that you need to know about the [details of close codes](https://discord.com/developers/docs/topics/gateway#initiating-a-disconnect) so you pass a right one that allows resuming. I agree that `stopAndResumeLater` isn't the best name for that function, maybe something like `stopForResume` (or something even better you can come up with)?

I don't think it is a bad idea, considering that the gateway module is a "low level" implementation of the gateway.

It could be better if it was a predefined list of close codes, just like the `GatewayCloseCodes` enum.

Lukellmann:
> > The problem I see with this approach is that you need to know about the [details of close codes](https://discord.com/developers/docs/topics/gateway#initiating-a-disconnect) so you pass a right one that allows resuming. I agree that `stopAndResumeLater` isn't the best name for that function, maybe something like `stopForResume` (or something even better you can come up with)?
> 
> I don't think it is a bad idea to allow the user to pass their own close code when stopping the gateway, considering that the gateway module is a "low level" implementation of the gateway.
> 
> It could be better if it was a predefined list of close codes, just like the `GatewayCloseCodes` enum.

Yeah, we can allow specifying the close code but should probably provide a default for both closing with invalidating the session and closing without invalidating the session (aka allowing resuming). Providing both defaults would mean providing two methods.

MrPowerGamerBR:
Now the `stop` functions return `GatewayResumeConfiguration`, yay!

I've decided to name the functions `stopForResume` and `stopAndInvalidateSession`.

I'm not sure if they are the *best* names for it. `stopAndResumeLater` makes you think that the gateway instance will automatically resume later (that's what I thought when you said it tbh).

Discord docs also doesn't help us a lot with how should we name it:

> When you close the connection to the gateway with close code 1000 or 1001, your session will be invalidated and your bot will appear offline.

> If you simply close the TCP connection or use a different close code, the session will remain active and timeout after a few minutes. This can be useful when you're [Resuming](https://discord.com/developers/docs/topics/gateway#resuming) the previous session.

The only alternative that I thought of is `stopWithoutInvalidatingSession`, but that seems kinda cumbersome to read, but at least it seems more similar to `stopAndInvalidateSession`.

MrPowerGamerBR:
I made a change to not consume the `identifyRateLimiter` when resuming a gateway connection, because, as far as I know, resuming do not have any rate limits, only identifying has.

<html><body>


max_concurrency | integer | Number of identify requests allowed per 5 seconds
-- | -- | --



</body>
</html>

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -64,9 +81,33 @@ public interface Gateway : CoroutineScope {
...
+    public suspend fun stopForResume(): GatewayResumeConfiguration = stop(WebSocketCloseReason(1012, "service restart"))
+
+    /**
+     * Closes the Gateway with code 1000, suspending until the underlying webSocket is closed.
+     *
+     * The session will be invalidated, and you won't be able to [resume] it. The bot will appear offline in the member list.
+     *
+     * @return the gateway session information, if there was any successful identify before closing the connection.
      */
-    public suspend fun stop()
```
Lukellmann, line 69: could you add this for binary compatibility:
```kotlin
public suspend fun stop() {
    stop(WebSocketCloseReason(1000, "leaving"))
}
```

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -40,13 +40,30 @@ public interface Gateway : CoroutineScope {
     public val ping: StateFlow<Duration?>
 
     /**
-     * Starts a reconnection gateway connection with the given [configuration].
+     * Starts a reconnecting gateway connection with the given [configuration].
+     *
+     * After connecting, an identify request will be sent and a session will be created.
```
Lukellmann, line 45:
```suggestion
     * After connecting, an [Identify] request will be sent and a session will be created.
```

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -40,13 +40,30 @@ public interface Gateway : CoroutineScope {
...
      * This function will suspend until the lifecycle of the gateway has ended.
      *
      * @param configuration the configuration for this gateway session.
      */
     public suspend fun start(configuration: GatewayConfiguration)
 
+    /**
+     * Starts a reconnecting gateway connection with the given [configuration].
+     *
+     * After connecting, a resume request will be sent and the session will be resumed if the session is valid.
```
Lukellmann, line 57:
```suggestion
     * After connecting, a [Resume] request will be sent and the session will be resumed if the session is valid.
```

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -64,9 +81,33 @@ public interface Gateway : CoroutineScope {
...
 
     /**
-     * Closes the Gateway and ends the current session, suspending until the underlying webSocket is closed.
+     * Closes the Gateway, suspending until the underlying webSocket is closed.
+     *
+     * By default, the current gateway session will be closed. If you want to keep the session alive to reconnect later,
+     * change the [closeReason] to use a different [WebSocketCloseReason.code] that isn't 1000 or 1001.
+     *
+     * @param closeReason the close reason that will be used when stopping the WebSocket connection.
+     * @return the gateway session information, if there was any successful identify before closing the connection.
```
Lukellmann, line 90:
```suggestion
     * Closes the Gateway, suspending until the underlying WebSocket is closed.
     *
     * By default, the current gateway session will be closed. If you want to keep the session alive to [resume] later,
     * change the [closeReason] to use a different [WebSocketCloseReason.code] that isn't `1000` or `1001`.
     *
     * Returns a [GatewayResumeConfiguration] that can be passed to [resume] later. The
     * [session][GatewayResumeConfiguration.session] will be non-null if there was any successful [Identify]
     * request before closing the connection.
     *
     * @param closeReason the close reason that will be used when closing the WebSocket connection.
```

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -64,9 +81,33 @@ public interface Gateway : CoroutineScope {
...
+
+    /**
+     * Closes the Gateway with code 1000, suspending until the underlying webSocket is closed.
+     *
+     * The session will be invalidated, and you won't be able to [resume] it. The bot will appear offline in the member list.
+     *
+     * @return the gateway session information, if there was any successful identify before closing the connection.
      */
-    public suspend fun stop()
+    public suspend fun stopAndInvalidateSession(): GatewayResumeConfiguration = stop(WebSocketCloseReason(1000, "leaving"))
```
Lukellmann, line 110: these should probably be extensions instead

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -112,6 +155,10 @@ public suspend inline fun Gateway.editPresence(builder: PresenceBuilder.() -> Un
 
 /**
  * Starts a reconnecting gateway connection with the given parameters.
+ *
+ * After connecting, an identify request will be sent and a session will be created.
```
Lukellmann, line 159:
```suggestion
 * After connecting, an [Identify] request will be sent and a session will be created.
```

gateway/src/main/kotlin/Gateway.kt
```diff
@@ -141,6 +188,48 @@ public suspend inline fun Gateway.start(token: String, config: GatewayConfigurat
...
+ * }
+ *
+ * //gateway has disconnected
+ * ```
+ *
+ * @param token     The Discord token of the bot.
+ * @param session   The gateway session information.
+ * @param config additional configuration for the gateway.
+ */
+public suspend inline fun Gateway.resume(token: String, session: GatewaySession, config: GatewayConfigurationBuilder.() -> Unit = {}) {
```
Lukellmann, line 219: Why do we need this function?

gateway/src/main/kotlin/GatewayResumeConfiguration.kt
```diff
@@ -0,0 +1,9 @@
+package dev.kord.gateway
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+public data class GatewayResumeConfiguration(
```
Lukellmann, line 6: I think `GatewayConfiguration`, `GatewayResumeConfiguration` and `GatewaySession` should be in the same file.

gateway/src/main/kotlin/handler/HandshakeHandler.kt
```diff
@@ -34,7 +34,11 @@ internal class HandshakeHandler(
     override fun start() {
         on<Hello> {
             reconnectRetry.reset() // connected and read without problems, resetting retry counter
-            identifyRateLimiter.consume {
+            
+            val resumeContext = resumeContext.value
+            if (resumeContext != null) {
+                send(Resume(configuration.token, resumeContext.sessionId, sequence.value ?: 0))
+            } else identifyRateLimiter.consume {
```
Lukellmann, line 41: #693 already fixes this, so no need to also do it here

Lukellmann (changes requested):

gateway/src/main/kotlin/Gateway.kt
MrPowerGamerBR: If you want to resume a session with a different gateway configuration, example: You want to change the bot's presence for the next time the bot reidentifies.

Lukellmann: Did not think about this 👍

MrPowerGamerBR:
Done :3

Lukellmann:
I forgot: what about adding this functionality to core?

MrPowerGamerBR:
Will do that later :)

HopeBaron:
Closed to be included in a up-to-date PR

PR #702, opened 2022-10-17
fix: make LinearRetry behave linearly for all inputs

### LINKED ISSUE #596: LinearRetry is not linear between start and finish
I assume linear retry should go like this:
image

**Case maxTries = 1, what duration should the delay be**
Unsure what the correct behaviour would be (e.g. firstBackOff or maxBackOff or the median between them ?)
Currently it is delay(maxBackOff-firstBackOff)

**Case maxTries = Int.MAX_VALUE**
It should probably the start at firstBackOff and progress linearly towards maxBackOff so that Int.MAX_VALUE increments were done to the delay to get there.
-> currently it starts at 0ms 

You can replicate this behaviour with the following code
```kotlin
    val retrier = LinearRetry(1.seconds,  60.seconds, Int.MAX_VALUE)
    runBlocking {
        while(retrier.hasNext) {
            retrier.retry()
            println(System.currentTimeMillis())
        }
    }
```

ToxicMushroom:
fixes #596

### DIFF
--- a/gateway/src/main/kotlin/retry/LinearRetry.kt
+++ b/gateway/src/main/kotlin/retry/LinearRetry.kt
@@ -28,7 +28,7 @@ public class LinearRetry(
         require(
             maxBackoff.minus(firstBackoff).isPositive()
         ) { "maxBackoff ${maxBackoff.inWholeMilliseconds} ms needs to be bigger than firstBackoff ${firstBackoff.inWholeMilliseconds} ms" }
-        require(maxTries > 1) { "maxTries needs to be positive but was $maxTries" }
+        require(maxTries > 1) { "maxTries needs to be greater than 1 but was $maxTries" }
     }
 
     private val tries = atomic(0)
--- a/gateway/src/main/kotlin/retry/LinearRetry.kt
+++ b/gateway/src/main/kotlin/retry/LinearRetry.kt
@@ -42,9 +42,12 @@ public class LinearRetry(
     override suspend fun retry() {
         if (!hasNext) error("max retries exceeded")
 
-        tries.incrementAndGet()
-        var diff = (maxBackoff - firstBackoff).inWholeMilliseconds / maxTries
-        diff *= tries.value
+        // tries/maxTries ratio * (backOffDiff) = retryProgress
+        val retryProgress =
+            ((tries.getAndIncrement() / (maxTries - 1).toDouble())
+                    * (maxBackoff.inWholeMilliseconds - firstBackoff.inWholeMilliseconds)).toLong()
+        val diff = firstBackoff.inWholeMilliseconds + retryProgress
+
         linearRetryLogger.trace { "retry attempt ${tries.value}, delaying for $diff ms" }
         delay(diff)
     }
--- a/gateway/src/main/kotlin/retry/LinearRetry.kt
+++ b/gateway/src/main/kotlin/retry/LinearRetry.kt
@@ -5,6 +5,7 @@ import kotlinx.atomicfu.update
 import kotlinx.coroutines.delay
 import mu.KotlinLogging
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
 
 private val linearRetryLogger = KotlinLogging.logger { }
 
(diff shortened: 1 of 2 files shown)


gateway/src/test/kotlin/helper/LinearRetryTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import dev.kord.gateway.retry.LinearRetry
+import kotlinx.coroutines.runBlocking
+import kotlin.test.Test
+import kotlin.time.Duration.Companion.milliseconds
+
+class LinearRetryTest {
+
+    @Test
+    fun testLinearity() {
```
Lukellmann, line 11: Could you maybe also add tests for extreme cases you described in the issue? (like `1`, `Int.MAX_VALUE`, ...).
Also to make these test independent of real time, the delay-skipping `runTest {}` can be used with `currentTime` for getting the current virtual time.

Lukellmann (changes requested):

gateway/src/test/kotlin/helper/LinearRetryTest.kt
ToxicMushroom: oh thanks for the tip I didn't know that existed

gateway/src/main/kotlin/retry/LinearRetry.kt
```diff
@@ -42,10 +43,13 @@ public class LinearRetry(
...
         if (!hasNext) error("max retries exceeded")
 
-        tries.incrementAndGet()
-        var diff = (maxBackoff - firstBackoff).inWholeMilliseconds / maxTries
-        diff *= tries.value
-        linearRetryLogger.trace { "retry attempt ${tries.value}, delaying for $diff ms" }
+        // tries/maxTries ratio * (backOffDiff) = retryProgress
+        val ratio = tries.getAndIncrement() / (maxTries - 1).toDouble()
+        val retryProgress =
+            (ratio * (maxBackoff - firstBackoff).inWholeMilliseconds).toLong().milliseconds
```
Lukellmann, line 49: If you import [`kotlin.time.times`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.time/times.html), you can also avoid the conversion here.

gateway/src/test/kotlin/helper/LinearRetryTest.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlin.time.TimeSource
+import kotlin.time.measureTime
+
+class LinearRetryTest {
+
+    @Test
+    fun testLinearity() = runTest {
+        val linearRetry = LinearRetry(1.seconds, 10.seconds, 10)
+        var i = 0
+        val elapsed = TimeSource.Monotonic.measureTime {
```
Lukellmann, line 18: Using `TimeSource.measureTime` won't work, it doesn't use the virtual [`currentTime`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/current-time.html).

gateway/src/main/kotlin/retry/LinearRetry.kt
ToxicMushroom: oh cool, intellij didn't show that so I assumed it didn't exist

gateway/src/main/kotlin/retry/LinearRetry.kt
```diff
@@ -27,7 +28,7 @@ public class LinearRetry(
         require(
             maxBackoff.minus(firstBackoff).isPositive()
         ) { "maxBackoff ${maxBackoff.inWholeMilliseconds} ms needs to be bigger than firstBackoff ${firstBackoff.inWholeMilliseconds} ms" }
-        require(maxTries > 0) { "maxTries needs to be positive but was $maxTries" }
+        require(maxTries > 1) { "maxTries needs to be positive but was $maxTries" }
```
Lukellmann, line 31: Just realized that the message should also be updated (`1` is also positive).

gateway/src/test/kotlin/helper/LinearRetryTest.kt
```diff
@@ -0,0 +1,44 @@
...
+        val start = currentTime
+
+        while (linearRetry.hasNext) {
+            linearRetry.retry()
+            i++
+        }
+
+        val end = currentTime
+        val elapsed = (end-start).milliseconds
+        assert(elapsed >= 55.seconds)
```
Lukellmann, line 25: These assertions should use `==` instead of `>=` or `<=` - virtual time is always exact and predictable.

Lukellmann (commented):
Just two last things - then we should be good to go.

Lukellmann (approved):
Thanks for your efforts!

PR #703, opened 2022-10-22
Discord endpoints

dookon:


Lukellmann:
Tests failed because of the Binary compatibility validator - you have to run `./gradlew clean apiDump` and commit the changes to make the tests pass.

dookon:
@Lukellmann Do you know a good way to make Icon deprecated? Because this way I did will break everyone's code

Lukellmann:
> @Lukellmann Do you know a good way to make Icon deprecated? Because this way I did will break everyone's code

We can simply keep the old `Icon` code and slap `@Deprecated` on everything and then gradually increase the deprecation level to nudge people into replacing it with `Asset` - this will also preserve binary compatibility.

Lukellmann (changes requested):

core/src/main/kotlin/entity/Icon.kt
```diff
@@ -40,19 +40,63 @@ public sealed class Icon(
         return "Icon(type=${javaClass.name},format=$format,animated=$animated,cdnUrl=$cdnUrl,kord=$kord)"
     }
 
+    @Deprecated(
+        "Icon class does not cover all CDN endpoints",
+        ReplaceWith("Asset.Emoji"),
+        DeprecationLevel.WARNING
+    )
     public class EmojiIcon(animated: Boolean, emojiId: Snowflake, kord: Kord) :
```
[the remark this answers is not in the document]
dookon: I did this to avoid binary compatibility errors.

core/src/main/kotlin/entity/Icon.kt
```diff
@@ -40,19 +40,63 @@ public sealed class Icon(
...
+        DeprecationLevel.WARNING
+    )
     public class DefaultUserAvatar(discriminator: Int, kord: Kord) :
-        Icon(Image.Format.PNG /* Discord Default Avatars only support PNG */, false, DiscordCdn.defaultAvatar(discriminator), kord)
+        Icon(
+            Image.Format.PNG /* Discord Default Avatars only support PNG */,
+            false,
+            DiscordCdn.defaultAvatar(discriminator),
+            kord
+        )
```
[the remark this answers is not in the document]
dookon: no problem

core/src/main/kotlin/entity/Icon.kt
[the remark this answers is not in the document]
Lukellmann: I think we should use the other approach I proposed my other comment, we want to remove `Icon` completely in the end and that won't work without compatibility issues if the new `Asset` has it as its super type.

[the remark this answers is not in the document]
dookon: Got it!

core/api/core.api
```diff
@@ -6267,7 +6338,7 @@ public final class dev/kord/core/entity/GuildEmoji : dev/kord/core/entity/KordEn
 	public final fun getData ()Ldev/kord/core/cache/data/EmojiData;
 	public final fun getGuildId ()Ldev/kord/common/entity/Snowflake;
 	public fun getId ()Ldev/kord/common/entity/Snowflake;
-	public final fun getImage ()Ldev/kord/core/entity/Icon;
+	public final fun getImage ()Ldev/kord/core/entity/Asset;
```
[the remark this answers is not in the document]
Lukellmann: Yes that'a right: you can have two methods on jvm with same name and parameters when they have a different return type. In the bytecode they are referenced by their full signature - including return type.
So to preserve binary compatibility we have to keep the old method signature.
But since you can't have two methods (or properties which become functions on jvm) with the same name and parameters but a different return type in Kotlin, we have to use this pattern with `@JvmName` and `@Deprecated(..., level = HIDDEN)`.
If you add one of them and look at the resulting API dump, you'll see that the method will be there twice with the same name but a different return type - the only other difference being that the one marked as deprecated with `HIDDEN` level is synthetic, meaning it isn't visible to newly written source but is still available to already compiled bytecode which keeps the binary compatibility.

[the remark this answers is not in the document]
dookon: Got it, thanks

dookon:
Why is this class ([Avatar](https://github.com/kordlib/kord/blob/0.8.x/core/src/main/kotlin/entity/User.kt#L88)) here?

Lukellmann:
Idk, is it used anywhere?

dookon:
No, I also searched for codes that import it

dookon:
Can I remove? Or mark as deprecated?

Lukellmann:
Deprecation would be the way to go then (the functionality is provided by `Asset` instead, right?).

dookon:
Yes

NoComment1105 (approved):

core/src/main/kotlin/entity/Icon.kt
```diff
@@ -43,16 +46,11 @@ public sealed class Icon(
     public class EmojiIcon(animated: Boolean, emojiId: Snowflake, kord: Kord) :
         Icon(if (animated) Image.Format.GIF else Image.Format.WEBP, animated, DiscordCdn.emoji(emojiId), kord)
 
-    public class DefaultUserAvatar(discriminator: Int, kord: Kord) :
```
Lukellmann, line 46: again, could you restore the formatting here? this will make it easier to review this pr

core/src/main/kotlin/entity/Asset.kt
```diff
@@ -0,0 +1,90 @@
+package dev.kord.core.entity
+
+import dev.kord.common.entity.Snowflake
+import dev.kord.core.Kord
+import dev.kord.core.KordObject
+import dev.kord.rest.Image
+import dev.kord.rest.route.CdnUrl
+import dev.kord.rest.route.DiscordCdn
+
+public sealed class Asset(
```
Lukellmann, line 10: shouldn't this instead be a normal final class with factory functions in the companion object? i don't see any values in the subclasses.

so instead of
```kt
public class DefaultUserAvatar(discriminator: Int, kord: Kord) :
    Asset(Image.Format.PNG, DiscordCdn.defaultAvatar(discriminator), kord)
```
we would have
```kotlin
public companion object {
    public fun defaultUserAvatar(discriminator: Int, kord: Kord): Asset =
        Asset(Image.Format.PNG, DiscordCdn.defaultAvatar(discriminator), kord)
}
```

core/src/main/kotlin/entity/Asset.kt
```diff
@@ -0,0 +1,90 @@
...
+    public val format: Image.Format,
```
Lukellmann, line 11: why is the default format png when it can't be animated and webp if the hash says it's not animated?

i think there should be other constructors instead:
the primary constructor as it is right now but with a default value for format (png or webp, whichever is the right default) and another constructors like this:
```kotlin
constructor(animated: Boolean, cdnUrl: CdnUrl, kord: Kord) :
    this(format = if (anmiated) GIF else <default format>, cdnUrl, kord)
```

core/src/main/kotlin/entity/Asset.kt
```diff
@@ -0,0 +1,90 @@
...
+    override val kord: Kord
+) : KordObject {
+    public val url: String
+        get() = cdnUrl.toUrl {
+            this.format = this@Asset.format
+        }
+
+    public val animated: Boolean get() = format is Image.Format.GIF
+
+    public suspend fun getImage(): Image = Image.fromUrl(kord.resources.httpClient, cdnUrl.toUrl())
```
Lukellmann, line 22:
```suggestion
    public suspend fun getImage(): Image = Image.fromUrl(kord.resources.httpClient, url)
```
this should use `Asset.format`, otherwise the default `Image.Fornat.WEBP` will be used (see `CdnUrl.toUrl()`

core/src/main/kotlin/entity/Asset.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    public suspend fun getImage(size: Image.Size): Image =
+        Image.fromUrl(kord.resources.httpClient, cdnUrl.toUrl {
+            this.size = size
+        })
```
Lukellmann, line 27:
```suggestion
    public suspend fun getImage(size: Image.Size): Image = getImage(format, size)
```
same as above

core/src/main/kotlin/entity/Asset.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    public class Emoji(animated: Boolean, emojiId: Snowflake, kord: Kord) : Asset(if (animated) Image.Format.GIF else Image.Format.WEBP, DiscordCdn.emoji(emojiId), kord)
+
+    public class DefaultUserAvatar(discriminator: Int, kord: Kord) :
+        Asset(Image.Format.PNG, DiscordCdn.defaultAvatar(discriminator), kord)
+
+    public class UserAvatar(userId: Snowflake, avatarHash: String, kord: Kord) :
+        Asset(if (avatarHash.startsWith("a_")) Image.Format.GIF else Image.Format.WEBP, DiscordCdn.userAvatar(userId, avatarHash), kord)
```
Lukellmann, line 46: i think the `hash.startsWith("a_")` logic should be extracted into a function, it's duplicated a lot

core/src/main/kotlin/entity/Guild.kt
```diff
@@ -354,12 +363,15 @@ public class Guild(
     /**
      * Gets the banner url in the specified format.
      */
+    @Deprecated("Old method", ReplaceWith("Guild#banner"), DeprecationLevel.WARNING)
```
Lukellmann, line 366:
```suggestion
    @Deprecated("Old method", ReplaceWith("this.banner"), DeprecationLevel.WARNING)
```
this way an automatic replacement can be performed by intellij (applies to other `ReplaceWith`s as well)

core/src/main/kotlin/entity/GuildScheduledEvent.kt
```diff
@@ -117,6 +117,8 @@ public class GuildScheduledEvent(
     /** The cover image hash of this event. */
     public val imageHash: String? get() = data.image.value
 
+    public val cover: Asset? get() = data.image.value?.let { Asset.GuildEventCover(guildId, it, kord) }
```
Lukellmann, line 120: i think `image` would be the better name - it's consistent with the already existing `imageHash`

core/src/main/kotlin/entity/PartialGuild.kt
```diff
@@ -85,10 +87,14 @@ public class PartialGuild(
...
     public val splashHash: String? get() = data.splash.value
 
+    public val splash: Asset? get() = data.splash.value?.let { Asset.GuildSplash(data.id, it, kord) }
+
+    public val banner: Asset? get() = data.banner.value?.let { Asset.GuildBanner(data.id, it, kord) }
 
     /**
      * Gets the discovery splash url in the specified [format], if present.
      */
+    @Deprecated("PartialGuild has no discovery splash field", level = DeprecationLevel.ERROR)
```
Lukellmann, line 97: yes it does? there is `splash` and `splashHash` after all

Lukellmann (changes requested):

Lukellmann:
any updates on this @oluiss? i've gone ahead and merged `0.9.x` with the mpp changes into this so it's easier for you to work with :)

oluiss:
I had even forgotten about this PR, I can apply the suggestions today

rest/src/main/kotlin/route/DiscordCdn.kt
```diff
@@ -15,5 +15,33 @@ public object DiscordCdn {
...
+        CdnUrl("$BASE_URL/discovery-splashes/$guildId/$hash")
+
+    public fun guildBanner(guildId: Snowflake, hash: String): CdnUrl = CdnUrl("$BASE_URL/banners/$guildId/$hash")
+
+    public fun userBanner(userId: Snowflake, hash: String): CdnUrl = CdnUrl("$BASE_URL/banners/$userId/$hash")
+
+    public fun applicationIcon(applicationId: Snowflake, hash: String): CdnUrl =
+        CdnUrl("$BASE_URL/app-icons/$applicationId/$hash")
+
+    public fun applicationCover(applicationId: Snowflake, hash: String): CdnUrl =
```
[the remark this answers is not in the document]
oluiss: I didn't find a way to access these endpoints

core/src/commonMain/kotlin/entity/Asset.kt
```diff
@@ -0,0 +1,77 @@
...
+    public val isAnimated: Boolean,
+    public val cdnUrl: CdnUrl,
+    override val kord: Kord,
+    private val forcedFormat: Image.Format? = null,
+) : KordObject {
+
+    public suspend fun getImage(format: Image.Format? = null, size: Image.Size? = null): Image = Image.fromUrl(
+        client = kord.resources.httpClient,
+        url = cdnUrl.toUrl {
+            this.format = forcedFormat ?: format ?: if (isAnimated) GIF else WEBP
```
Lukellmann, line 21: so webp is the right default format for non-animated assets? just curious because i myself don't know what it should be tbh

oluiss: I don't know either xd

Lukellmann: @HopeBaron, what do you think?

HopeBaron: Based on this: https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints
it seems that the default is PNG

Lukellmann: would also say png seems more reasonable, should be changed then

rest/src/main/kotlin/route/DiscordCdn.kt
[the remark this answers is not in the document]
Lukellmann: with 'Achievement Icon' and 'Store Page Asset' maybe, but 'Sticker' ([`id`](https://discord.com/developers/docs/resources/sticker#sticker-object-sticker-structure)) and 'Application Asset' ([`large_image` / `small_image`](https://discord.com/developers/docs/topics/gateway-events#activity-object-activity-assets)) should work, right?

core/src/commonMain/kotlin/entity/Asset.kt
Lukellmann: i changed the default to png in https://github.com/kordlib/kord/pull/703/commits/b61555d52fc15519b6aee56bdfb4b0f324aa04e1

rest/src/main/kotlin/route/DiscordCdn.kt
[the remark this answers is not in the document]
Lukellmann: was able to access 'Sticker', added in https://github.com/kordlib/kord/pull/703/commits/7a893c5348716c3594672e4aaacdd3b491e77fe8, will try 'Application Asset' next

tested with the following (also works for other `MessageStickerType`s):
```kt
val lottieStickerBytes = kord.nitroStickerPacks
    .flatMapConcat { it.stickers.asFlow() }
    .first { it.formatType == MessageStickerType.LOTTIE }
    .image
    .getImage()
    .data

channel.createMessage {
    addFile("lottie_sticker.json", ChannelProvider { ByteReadChannel(lottieStickerBytes) })
}
```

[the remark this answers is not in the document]
Lukellmann: ok, 'Application Asset' is strange (see [here](https://discord.com/developers/docs/topics/gateway-events#activity-object-activity-asset-image)), but it's such a niche feature that i think it's ok when we don't provide that, it's fine for me then

Lukellmann:
@oluiss i made quite some changes now but i think it should be ready now, could you take another look at what i did?

oluiss:
looks ok to me

Lukellmann (approved):

PR #704, opened 2022-10-24
Nitro Basic type

0spura:
see https://github.com/discord/discord-api-docs/commit/48d3c2d324796d8236355262ace3b22c1d29acb0

### DIFF
--- a/common/api/common.api
+++ b/common/api/common.api
@@ -7932,6 +7932,10 @@ public final class dev/kord/common/entity/UserPremium$Nitro : dev/kord/common/en
 	public static final field INSTANCE Ldev/kord/common/entity/UserPremium$Nitro;
 }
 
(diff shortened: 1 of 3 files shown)


Lukellmann:
Please update the annotation [here](https://github.com/kordlib/kord/blob/0.8.x/common/src/main/kotlin/entity/DiscordUser.kt) instead and run `./gradlew apiDump` to regenerate `UserPremium.kt` and update the API dump.

dookon:
> Please update the annotation [here](https://github.com/kordlib/kord/blob/0.8.x/common/src/main/kotlin/entity/DiscordUser.kt) instead and run `./gradlew apiDump` to regenerate `UserPremium.kt` and update the API dump.

Oh, sorry, I forgot

Lukellmann:
Btw the reason the build passed is that the generated file you editied got overwritten to the old version during the build.

Lukellmann (approved):
Thanks!

PR #706, opened 2022-10-26
Deletable ephemeral messages

dookon:
see https://github.com/discord/discord-api-docs/commit/8ebc79d47d554d0ba557e2a9103129ee83372fbc

### DIFF
--- a/core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
+++ b/core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
@@ -17,13 +17,17 @@ import kotlin.contracts.contract
  * [updatePublicMessage][ComponentInteractionBehavior.updatePublicMessage] or
  * [updateEphemeralMessage][ComponentInteractionBehavior.updateEphemeralMessage].
  *
- * This is the handle to a message, it supports [editing][MessageInteractionResponseBehavior.edit] and sending followup
- * messages to the interaction.
+ * This is the handle to a message, it supports
+ * [editing][MessageInteractionResponseBehavior.edit],
+ * [deleting][delete] and sending followup messages to the interaction.
  */
 public interface MessageInteractionResponseBehavior : FollowupPermittingInteractionResponseBehavior {
 
     /**
      * Requests to delete the message.
+     *
+     * This [MessageInteractionResponseBehavior] can still be used to send followup messages to the interaction.
+     *
      * @throws RestRequestException if something went wrong during the request.
      */
     public suspend fun delete() {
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
```diff
@@ -22,6 +22,14 @@ import kotlin.contracts.contract
  */
 public interface MessageInteractionResponseBehavior : FollowupPermittingInteractionResponseBehavior {
 
+    /**
+     * Requests to delete the message.
+     * @throws RestRequestException if something went wrong during the request.
+     */
```
Lukellmann, line 28:
```suggestion
    /**
     * Requests to delete the message.
     *
     * This [MessageInteractionResponseBehavior] can still be used to send followup messages to the interaction.
     *
     * @throws RestRequestException if something went wrong during the request.
     */
```
Or was there a reason that you left this part out?

Lukellmann (changes requested):
Looks good overall, just two little things, see comments.
Have you also tested that it works?

core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
dookon: I didn't understand this "This [MessageInteractionResponseBehavior] can still..." part being documented in this function, so I removed it, but I can put it back in

core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
```diff
@@ -22,6 +22,14 @@ import kotlin.contracts.contract
...
+    public suspend fun delete() {
+        kord.rest.interaction.deleteOriginalInteractionResponse(applicationId, token)
+    }
```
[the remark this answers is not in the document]
dookon: Sure!

dookon:
> Looks good overall, just two little things, see comments. Have you also tested that it works?

I tested it, it works

core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
Lukellmann: It means that after you run code like this:
```kotlin
val response: MessageInteractionResponseBehavior = interaction.respondPublic { /* ... */ }
response.delete()
```
you can still do this:
```kotlin
response.createPublicFollowup { /* ... */ }
```

If you can think of a way to better describe this, feel free to add it.

Lukellmann (approved):
lgtm and should be ready to merge except you want to clarify documentation for [this](https://github.com/kordlib/kord/pull/706#discussion_r1006360317).
Thanks!

core/src/main/kotlin/behavior/interaction/response/MessageInteractionResponseBehavior.kt
dookon: aaah, makes sense

dookon:
Able to merge.

PR #707, opened 2022-10-28
Select menu types

dookon:
Major API changes are:

 * renamed `ComponentType.SelectMenu` to `ComponentType.StringSelect`
 * new `ComponentType`s: `UserSelect`, `RoleSelect`, `MentionableSelect` and `ChannelSelect`
 * `SelectMenuComponent` subtype for each select menu type
 * DSL function for each select menu type in `ActionRowBuilder`

see https://github.com/discord/discord-api-docs/pull/5543 and https://github.com/discord/discord-api-docs/pull/5544


NoComment1105 (approved):
LGTM, very excited for some of these menu types 👀

core/src/main/kotlin/entity/component/SelectMenuComponent.kt
```diff
@@ -12,10 +13,7 @@ import dev.kord.core.entity.interaction.SelectMenuInteraction
  * An interactive dropdown menu rendered on a [Message] that consists of multiple [options].
  */
 
-public class SelectMenuComponent(override val data: ComponentData) : Component {
-
-    override val type: ComponentType.SelectMenu
-        get() = ComponentType.SelectMenu
+public open class SelectMenuComponent(override val data: ComponentData) : Component {
```
Lukellmann, line 16:
```suggestion
public open class SelectMenuComponent(override val data: ComponentData) : Component {

    @Suppress("DEPRECATION")
    @Deprecated("Binary compatibility", level = HIDDEN)
    @get:JvmName("getType")
    public val type0: ComponentType.SelectMenu get() = ComponentType.SelectMenu
```

core/src/main/kotlin/entity/component/SelectMenuComponent.kt
```diff
@@ -61,6 +61,18 @@ public class SelectMenuComponent(override val data: ComponentData) : Component {
...
+
+public class UserSelectComponent(data: ComponentData) : SelectMenuComponent(data)
+
+public class RoleSelectComponent(data: ComponentData) : SelectMenuComponent(data)
+
+public class MentionableSelectComponent(data: ComponentData) : SelectMenuComponent(data)
+
+public class ChannelSelectComponent(data: ComponentData) : SelectMenuComponent(data) {
+    public val channelTypes: List<ChannelType>? get() = data.channelTypes.value
+}
```
Lukellmann, line 74: I think these should override `type` like `SelectMenuComponent` did before.

core/src/main/kotlin/entity/component/SelectMenuComponent.kt
```diff
@@ -30,6 +28,8 @@ public class SelectMenuComponent(override val data: ComponentData) : Component {
 
     /**
      * The possible options to choose from.
+     *
+     * **Available only in [ComponentType.StringSelect]**
      */
     public val options: List<SelectOption> get() = data.options.orEmpty().map { SelectOption(it) }
```
Lukellmann, line 34: I think this should be deprecated instead and be added to the new `StringSelectComponent` subtype.

core/src/main/kotlin/entity/component/SelectMenuComponent.kt
Lukellmann, line 16:
```suggestion
public open class SelectMenuComponent
@Deprecated(
    "This will be made a sealed class in the future, please stop using this constructor. You can instead use the " +
            "constructor of one of the subtypes.",
    level = WARNING,
)
public constructor (override val data: ComponentData) : Component {
```
We will increase the deprecation level and then replace it with a sealed class.

rest/src/main/kotlin/builder/component/ActionRowBuilder.kt
```diff
@@ -44,12 +44,73 @@ public class ActionRowBuilder : MessageComponentBuilder {
      * Creates and adds a select menu with the [customId] and configured by the [builder].
      * An ActionRow with a select menu cannot have any other select menus or buttons.
      */
-    public inline fun selectMenu(customId: String, builder: SelectMenuBuilder.() -> Unit) {
+    @Deprecated("Renamed by discord to StringSelect", ReplaceWith("stringSelect"))
```
Lukellmann, line 47:
```suggestion
    @Deprecated("Renamed by discord to StringSelect", ReplaceWith("stringSelect(customId, builder)"))
```
will fill parameters when using automatic replacement in ide

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
```diff
@@ -16,9 +17,10 @@ import kotlin.contracts.contract
  * @param customId The identifier for the menu, max 100 characters.
  */
 @KordDsl
-public class SelectMenuBuilder(
+public open class SelectMenuBuilder(
     public var customId: String,
 ) : ActionRowComponentBuilder() {
```
Lukellmann, line 22: same thing as with `SelectMenuComponent` in core: deprecate the public constructor and replace with sealed class in the future

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
```diff
@@ -16,9 +17,10 @@ import kotlin.contracts.contract
...
+    internal open val type: ComponentType = ComponentType.StringSelect
```
Lukellmann, line 23:
```suggestion
    internal open val type: ComponentType get() = ComponentType.StringSelect
```
can be a getter, so no field is generated, same for subtypes

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
```diff
@@ -45,6 +47,8 @@ public class SelectMenuBuilder(
     /**
      * Adds a new option to the select menu with the given [label] and [value] and configured by the [builder].
      *
+     * **Available only in [ComponentType.StringSelect]**
```
Lukellmann, line 50: I think this should be deprecated instead and be added to the new `StringSelectBuilder` subtype.

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
```diff
@@ -69,3 +73,45 @@ public class SelectMenuBuilder(
...
+            type,
+            customId = Optional(customId),
+            disabled = _disabled,
+            placeholder = _placeholder,
+            minValues = OptionalInt.Value(allowedValues.start),
+            maxValues = OptionalInt.Value(allowedValues.endInclusive),
+            options = Optional(options.map { it.build() }),
+            channelTypes = _channelTypes
+        )
+    }
```
Lukellmann, line 116: I would instead add the following in the supertype instead
```kotlin
protected open fun buildChannelTypes(): Optional<List<ChannelType>> = Optional.Missing()
```
that is then used in the `build()` of the supertypes and override it here
```kotlin
override fun buildChannelTypes() = _channelTypes
```
This way we don't have to duplicate the whole logic of `build()` here.

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
```diff
@@ -69,3 +73,45 @@ public class SelectMenuBuilder(
...
+    override val type: ComponentType = ComponentType.MentionableSelect
+}
+
+@KordDsl
+public class ChannelSelectBuilder(customId: String) : SelectMenuBuilder(customId) {
+    override val type: ComponentType = ComponentType.ChannelSelect
+
+    private var _channelTypes: Optional<List<ChannelType>> = Optional.Missing()
+
+    public var channelTypes: List<ChannelType>? by ::_channelTypes.delegate()
```
Lukellmann, line 103: I suggest using `MutableList` here - it's more flexible and we use it in most of the other builders.

Lukellmann (changes requested):

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
Lukellmann: Take a look at the [automod builders](https://github.com/kordlib/kord/blob/828c924eccf89fe904593250ea216c0c8019a227/rest/src/main/kotlin/builder/automoderation/AutoModerationRuleCreateBuilder.kt#L31), they already use this pattern.

rest/src/main/kotlin/builder/component/SelectMenuBuilder.kt
dookon: How can I override an inline fun?

Lukellmann: I've got a solution that uses a hack with `kotlin.internal.LowPriorityInOverloadResolution`, I'll push it later if that's fine for you.

dookon: No problem, thank you!

Lukellmann: I've committed the solution here: https://github.com/kordlib/kord/pull/707/commits/b9a0682e4359387a3c4c4b15730a32285b279218

Lukellmann: You can try out that the automatic replacement works as intended with code like:
```kotlin
import dev.kord.core.behavior.channel.createMessage
import dev.kord.rest.builder.message.create.actionRow

channel.createMessage {
    actionRow {
        stringSelect("custom-id") {
            option("label", "value") { description = "description" }
        }
    }
}
```
It will then add `dev.kord.rest.builder.component.option` to the imports. The extension function is used instead of the member because of `@kotlin.internal.LowPriorityInOverloadResolution`.

Lukellmann:
Do you have unpushed changes for this? Since you already marked the comments as resolved.

dookon:
> Do you have unpushed changes for this? Since you already marked the comments as resolved.

Yes! I'll try to make commit later today

core/src/main/kotlin/entity/interaction/SelectMenuInteraction.kt
```diff
@@ -22,6 +24,12 @@ public sealed interface SelectMenuInteraction : ComponentInteraction {
     override val component: SelectMenuComponent
         get() = message.actionRows.firstNotNullOf { it.selectMenus[componentId] }
 
+    /**
+     * The resolved object is included in interaction payloads for user, role, mentionable, and channel select menu components
+     */
+    public val resolvedObjects: ResolvedObjects?
```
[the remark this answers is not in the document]
dookon: I was thinking about it, I think it's better to leave it like this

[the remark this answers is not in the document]
Lukellmann: Hm, could be nicer but I agree.
We can add something in the future if needed.

Lukellmann (approved):
Looks good, I just have a few last things I changed and just committed. Can you take a look at them? Then we should be good to go.

dookon:
Sure

dookon:
It's ok for me.

Lukellmann:
Alright, thanks!

PR #709, opened 2022-11-02
Forum posts

oluiss:
see https://github.com/discord/discord-api-docs/pull/4850


NoComment1105:
Wouldn't the `createPublicThread` make more sense to the user as `createPublicForum`? Seems an unnecessary confusion. If the thread element is necessary perhaps `createPublicForumThread`?

dookon:
do you mean `startPublicThread` in `ForumChannelBehavior`? I couldn't find anything by that name

NoComment1105:
Yes

dookon:
the name follows discord docs, I also agree it should be renamed to create

NoComment1105:
hm ok. The Discord docs don't really make much sense. It's creating a forum thread, not just a thread 🤔

dookon:
true lol

NoComment1105 (approved):
LGTM other than the naming for functions that create forums threads that I've voiced before 👍🏼

dookon:
Don't review yet, I want to add some things that were missing

dookon (commented):
Several problem: Distinguish a text thread from a forum thread

core/src/main/kotlin/entity/channel/thread/ForumChannelThread.kt
```diff
@@ -0,0 +1,15 @@
...
+import dev.kord.core.behavior.channel.threads.ForumChannelThreadBehavior
+import dev.kord.core.cache.data.ChannelData
+import dev.kord.core.supplier.EntitySupplier
+
+public class ForumChannelThread(
+    override val data: ChannelData,
+    override val kord: Kord,
+    override val supplier: EntitySupplier = kord.defaultSupplier
+) : ThreadChannel, ForumChannelThreadBehavior {
+    public val appliedTags: List<Snowflake>? get() = data.appliedTags.value
```
dookon, line 14: A way to make requests on tags has not yet been documented

core/src/main/kotlin/entity/channel/thread/TextChannelThread.kt
```diff
@@ -31,6 +33,11 @@ public class TextChannelThread(
      */
     public val isInvitable: Boolean get() = data.threadMetadata.value!!.invitable.discordBoolean
 
+    /**
+     * Only available when creating a thread in a forum channel
+     */
+    public val message: Message get() = data.message.unwrap { Message(it, kord) }!!
```
dookon, line 39: I will remove it from here

core/src/main/kotlin/entity/channel/Channel.kt
```diff
@@ -55,7 +56,9 @@ public interface Channel : ChannelBehavior {
             GuildNews -> NewsChannel(data, kord)
             GuildForum -> ForumChannel(data, kord)
             PublicNewsThread -> NewsChannelThread(data, kord)
-            PrivateThread, PublicGuildThread -> TextChannelThread(data, kord)
+            PrivateThread, PublicGuildThread -> {
```
dookon, line 59: Until discussion https://github.com/discord/discord-api-docs/discussions/5658 is answered or an efficient way to distinguish a forum thread from a text thread is found, this PR cannot be merge

core/src/main/kotlin/behavior/channel/threads/ForumChannelThreadBehavior.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * * Editing a thread to set [archived][ThreadChannel.isArchived] to false only requires the current user to be in the thread.
+ * * If [locked][ThreadChannel.isLocked] is true, then the user must have [Manage Threads][dev.kord.common.entity.Permission.ManageThreads]
+ * * Editing a thread to change the
+ * [name][ThreadModifyBuilder.name],
+ * [archived][ThreadModifyBuilder.archived],
+ * [autoArchiveDuration][ThreadModifyBuilder.autoArchiveDuration] fields
+ * requires [Manage Threads][dev.kord.common.entity.Permission.ManageThreads] or that the current user is the thread creator.
+ */
+public suspend inline fun ForumChannelThreadBehavior.edit(builder: ForumThreadModifyBuilder.() -> Unit): ForumChannelThread {
```
dookon, line 23: Overload resolution ambiguity, because `ThreadChannelBehavior.edit` function, I thought of creating an interface called `TextThreadChannelBehavior` and moving `ThreadChannelBehavior.edit` to it, while `ForumChannelThreadBehavior` would inherit `ThreadChannelBehavior`. Is it correct to do this?

HopeBaron: We should go with a generic builder approach to resolve these nasty sort of errors. But that's beyond this PR's scope

HopeBaron:
The PR does more than the changed documentation do, is this intentional?

HopeBaron:
The PR does more than the changed documentation do, is this intentional?

HopeBaron:
The PR does more than the changed documentation do, is this intentional?

core/src/main/kotlin/behavior/channel/threads/ForumChannelThreadBehavior.kt
Lukellmann: ok turns out this ambiguity only exists because the receiver of the passed lambda differs (`ForumThreadModifyBuilder` vs `ThreadModifyBuilder`), if it was the same there would be no ambiguity

should we just have one builder for both (meaning that you could some unsafe things when editing normal threads)?

NoComment1105: I believe, given hopes goal to change to generic builders, this would be a good solution for now.

HopeBaron: The solution for this is to move to Generic Builders as I added in Discord, I think it's better to add this to an issue.
But this will mean:
* Less code to maintain
* Builders resuability
* Solves these kind of issues

While we try to enforce types, the user still can make mistakes as we have seen with slash commands
So the whole point of having "Type enforced builders" doesn't stand a chance against the challenges and updates Discord have undergone

NoComment1105: A temporary solution needs to be found though, you said yourself that this is out of scope for this pr

HopeBaron: yeah it's ok to add them together.
Just stating that this is a good indicator we need to re-evaluate our abstractions

oluiss:
> The PR does more than the changed documentation do, is this intentional?

I wanted to take advantage of the pull request, as this is part of the forums, but if you want I can make a separate one

oluiss:
> The PR does more than the changed documentation do, is this intentional?

I wanted to take advantage of the pull request, as this is part of the forums, but if you want I can make a separate one

core/src/main/kotlin/entity/channel/Channel.kt
HopeBaron: You can tell by checking if there is an applied tags as non-optional

core/src/main/kotlin/cache/data/ChannelData.kt
```diff
@@ -78,6 +83,11 @@ public data class ChannelData(
                 defaultSortOrder,
                 totalMessageSent,
                 defaultForumLayout,
+                availableTags,
+                appliedTags,
+                defaultReactionEmoji,
+                defaultThreadRateLimitPerUser,
+                message.map { MessageData.from(it) }
```
HopeBaron, line 90: There is no message object as far as I'm looking at the channel object on the docs
you may be looking for the last_message_id which we already have available?

The only additional field will be the tags and the reaction emoji fields

core/src/main/kotlin/entity/channel/thread/ForumChannelThread.kt
HopeBaron: you can by using the `available_tags` field on the forum channel itself the set of tags that can be used in a GUILD_FORUM channel

HopeBaron (changes requested):

core/src/main/kotlin/entity/channel/ForumChannel.kt
```diff
@@ -31,26 +28,6 @@ public class ForumChannel(
 
     public val defaultThreadRateLimitPerUser: Duration? get() = data.defaultThreadRateLimitPerUser.value
 
-    public suspend fun createTag(name: String, builder: ForumTagBuilder.() -> Unit = {}): ForumChannel {
```
NoComment1105: @HopeBaron why tf remove????

core/src/main/kotlin/behavior/channel/ForumChannelBehavior.kt
```diff
@@ -1,20 +1,75 @@
...
 import kotlin.contracts.contract
 
 public interface ForumChannelBehavior : ThreadParentChannelBehavior {
+    override val activeThreads: Flow<TextChannelThread>
+        get() = super.activeThreads.filterIsInstance()
+
+    override fun getPublicArchivedThreads(before: Instant?, limit: Int?): Flow<TextChannelThread> {
+        return super.getPublicArchivedThreads(before, limit).filterIsInstance()
+    }
+    public suspend fun createTag(name: String, builder: ForumTagBuilder.() -> Unit = {}): ForumChannel {
```
NoComment1105, line 27: Formatting
```suggestion
    }
    
    public suspend fun createTag(name: String, builder: ForumTagBuilder.() -> Unit = {}): ForumChannel {
```

rest/src/main/kotlin/builder/channel/thread/ThreadModifyBuilder.kt
```diff
@@ -31,8 +31,8 @@ public class ThreadModifyBuilder : AuditRequestBuilder<ChannelModifyPatchRequest
     private var _invitable: OptionalBoolean = OptionalBoolean.Missing
     public var invitable: Boolean? by ::_invitable.delegate()
 
-    private var _flags: Optional<ChannelFlags> = Optional.Missing()
-    public var flags: ChannelFlags? by ::_flags.delegate()
+    private var _appliedTags: Optional<List<Snowflake>> = Optional.Missing()
+    public var appliedTags: List<Snowflake>? by ::_appliedTags.delegate()
```
NoComment1105, line 35: These should be `MutableList`s to make adding to them more simple for adding too.

NoComment1105 (changes requested):
1 issue and a minor nitpick. Other than that LGTM i cannot wait for merge :D

NoComment1105 (approved):

core/src/main/kotlin/cache/data/ChannelData.kt
oluiss: Message object retuns in DiscordChannel (??) https://discord.com/developers/docs/resources/channel#start-thread-in-forum-channel

PR #714, opened 2022-11-06
Add a getGuildOrNull function to the Kord class

NoComment1105:
Previously there was just a `getGuild` function that called the nullable variant below the hood, but the non-nullable variant also exists.
 I've created the `getGuildOrNull` function to clearly show it function returns a null, and the `getGuild` function now calls the non-nullable function, and as such will throw `EntityNotFoundException` when called and the guild is null.

This  could be quite a breaking change for bots, as the `getGuild` function no longer returns null, it throws. This should be announced somewhere, or a release created to document.

I also KDoc'd these functions :)

### DIFF
--- a/core/src/test/kotlin/rest/RestTest.kt
+++ b/core/src/test/kotlin/rest/RestTest.kt
@@ -86,7 +86,7 @@ class RestServiceTest {
 
         guildId = guild.id
 
-        this@RestServiceTest.guild = kord.getGuild(guildId)!!
+        this@RestServiceTest.guild = kord.getGuildOrNull(guildId)!!
 
         guild.edit {
             name = "Edited Guild Test"
@@ -555,7 +555,7 @@ class RestServiceTest {
     fun `create role with image icon`(): Unit = runBlocking {
         if (!boostEnabled)
             return@runBlocking
-        val guild = kord.getGuild(publicGuildId)!!
+        val guild = kord.getGuildOrNull(publicGuildId)!!
         guild.createRole {
             name = "Test Image Icon"
             hoist = true
@@ -567,7 +567,7 @@ class RestServiceTest {
     fun `create role with unicode icon`(): Unit = runBlocking {
         if (!boostEnabled)
             return@runBlocking
-        val guild = kord.getGuild(publicGuildId)!!
+        val guild = kord.getGuildOrNull(publicGuildId)!!
         guild.createRole {
             name = "Test Unicode Icon"
             hoist = true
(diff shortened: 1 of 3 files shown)


Lukellmann:
That's why the build fails:
```
e: warnings found and -Werror specified
w: /home/runner/work/kord/kord/core/src/test/kotlin/rest/RestTest.kt: (89, 60): Unnecessary non-null assertion (!!) on a non-null receiver of type Guild
w: /home/runner/work/kord/kord/core/src/test/kotlin/rest/RestTest.kt: (558, 49): Unnecessary non-null assertion (!!) on a non-null receiver of type Guild
w: /home/runner/work/kord/kord/core/src/test/kotlin/rest/RestTest.kt: (570, 49): Unnecessary non-null assertion (!!) on a non-null receiver of type Guild
```

Lukellmann:
> This  could be quite a breaking change for bots, as the `getGuild` function no longer returns null, it throws. This should be announced somewhere, or a release created to document.

The alternative would be to deprecate `getGuild` that can return `null`, replace it with `getGuildOrNull` and then later reintroduce `getGuild` that throws - this wouldn't be a breaking change but take longer until it's fully released.

NoComment1105:
That seems reasonable, personally I don't like the idea of going without the throwing function, maybe using JVM name annotations to allow both to exist could be used?

Lukellmann:
Problem with that is that it might be binary compatible but it would still silently remove `null` returns.

We could maybe add a temporaty function like `getGuildOrThrow` that will then be deprecated and removed again when `getGuild` is converted to throwing behavior.

NoComment1105:
Yeah that seems reasonable to me,if you're ok with that

core/src/test/kotlin/rest/RestTest.kt
```diff
@@ -86,7 +86,7 @@ class RestServiceTest {
 
         guildId = guild.id
 
-        this@RestServiceTest.guild = kord.getGuild(guildId)!!
+        this@RestServiceTest.guild = kord.getGuildOrNull(guildId)!!
```
Lukellmann, line 89: I think here we should just use `getGuildOrThrow`.

NoComment1105: Just there or in all applicable situations?

Lukellmann: in all three places, just was to lazy :D

NoComment1105: Haha ok

NoComment1105: All sorted

Lukellmann (approved):
Thanks! :)

Issue #711, opened 2022-11-02
Think about changing Snowflakes/Permissions/DiscordBitSets/UserFlags/etc to value classes


MrPowerGamerBR:
I was creating a entity cache server with Kord, and one of the things that I've noticed is the huge memory footprint.

(Using `-XX:+UseStringDeduplication`)

In these tests, I'm caching...
```
Users: 15031 users
Guilds: 47614 guilds
Channels: 1578752 channels
Guild Members: 47615 members (68290 total)
Guild Roles: 47614 guild roles (1237487 total)
Guild Emojis: 47614 guild emojis (414569 total)
Guild Voice States: 47615 guild voice states (4226 total)
```

```
No Snowflake value classes:
Used Memory: 1427MiB

Snowflake Value classes:
Used Memory: 1347MiB

Snowflake Value class + SnowflakeMap (SnowflakeMap = a class that wraps fastutil's Long2ObjectOpenHashMap)
Used Memory: 1215MiB

Snowflake `Permissions` and `DiscordBitSet` to be `value` and immutable:
Used Memory: 966MiB
```

Before:
![https://cdn.discordapp.com/attachments/587324906702766226/1037434893254279308/image.png](https://cdn.discordapp.com/attachments/587324906702766226/1037434893254279308/image.png)

After:
![https://cdn.discordapp.com/attachments/587324906702766226/1037434893690478752/image.png](https://cdn.discordapp.com/attachments/587324906702766226/1037434893690478752/image.png)

Even though this would cause a binary incompatibility, I think it would be a good change to reduce the memory footprint used by Kord.

The `DiscordBitSet` would be hard to change to `value` because it is mutable, and value classes cannot be mutable. I think that it would be nice to have `MutableDiscordBitSet` (or `DiscordBitSetBuilder`?) instead of allowing the current `DiscordBitSet` to be mutable.

The `Snowflake` would be hard to change to `value` due to its `ULong` constructor, which is used to coerce the value.

Changing the `UserFlags` and `Permissions` to value classes can be done in a straightforward manner.

zTrap:
Maybe add this in 1.0.0? These changes is hard breaking for api, but its ok if change major version too

DRSchlaubi:
haha I literally just brought this up internally

zTrap:
[...]

PR #717, opened 2022-11-09
Lightweight Snowflakes

MrPowerGamerBR:
Changes Snowflakes to be `value class` to reduce their memory footprint.

I know that this PR *probably* won't be merged because it is a huge binary change, so it would break everything. Anyhow, this PR is here so other users can implement this change in their own Kord fork, if they need it.

Related to #711

Sadly the constructor can't have `value.coerceIn(validValues)`, and we can't use a `interface` to "hide" the value class behind a `SnowflakeImpl` because, if you use interfaces, the class will be boxed. The failing tests are related to the missing `coerceIn` call.

### DIFF
--- a/common/api/common.api
+++ b/common/api/common.api
@@ -268,18 +268,18 @@ public final class dev/kord/common/entity/ActivityType$Watching : dev/kord/commo
 
 public final class dev/kord/common/entity/AllRemovedMessageReactions {
 	public static final field Companion Ldev/kord/common/entity/AllRemovedMessageReactions$Companion;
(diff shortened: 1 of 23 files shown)


Lukellmann:
> Sadly the constructor can't have `value.coerceIn(validValues)`, and we can't use a `interface` to "hide" the value class behind a `SnowflakeImpl` because, if you use interfaces, the class will be boxed. The failing tests are related to the missing `coerceIn` call.

I thought about using `require` in `init` before, this approach would still work for a `value class`. It would still be a change in behavior though.

ToxicMushroom:
Can't we use a typealias for snowflake ?

Lukellmann:
> Can't we use a typealias for snowflake ?
> It seems to just store a ULong and not do much else

That would mean that all extensions that we then would have to define on snowflakes are usable for all `Long`s...

HopeBaron:
This will be closed to be included in a up to date PR
Thanks @MrPowerGamerBR

PR #719, opened 2022-11-12
Use `libs.versions.toml` and clean gradle files

Lukellmann:
Declaring the version catalog in `libs.versions.toml` instead of `settings.gradle.kts` has the advantage that it can be reused in `buildSrc`. It also introduces a single place to define dependencies and their version.

General cleanup of gradle files was also done, including:

* using default output for dokka: relying on defaults means less config and maintenance burden

* removing [`DocsTask`](https://github.com/kordlib/kord/blob/904964d467f40a4e8bfd61636f6fa20a18e54719/buildSrc/src/main/kotlin/Publishing.kt) because it was unused, last usage of this custom task was [removed](https://github.com/kordlib/kord/commit/3cecc8c2d4a05499d902ce0fcd0ab4d2430615bd#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06L224) in #367

* getting rid of the [source set workaround](https://github.com/kordlib/kord/commit/c1e1bf5f4c78dd6e5f5ae0ddb7e6bdd2b5573e3f#diff-d10a70e0fc20d2762692662771da85dbb2cb2284d3b642259b06ba6ed2b264e8R17-R23) introduced in #386 by moving `TweetNaclFast` from `voice/src/main/kotlin` to `voice/src/main/java`

* removing `io.codearte.nexus-staging` plugin introduced in #154 since the tasks of the plugin were never used (Nexus GUI was used manually instead)


Lukellmann:
@DRSchlaubi [this](https://github.com/DRSchlaubi/kord-dokka/blob/3c74c5303a5637f80ae46470c98daaa64b8ee316/.github/workflows/docs.yaml#L27) will have to be updated to `build/dokka/htmlMultiModule` after this PR too (just like [here](https://github.com/kordlib/kord/pull/719/files#diff-8d6584d14a3deee2e4402ebe6adc45de21a0b862e3765bdae7ea4f54e1be7e2a))

build.gradle.kts
```diff
@@ -1,14 +1,12 @@
...
-    kotlin("jvm")
-    kotlin("plugin.serialization")
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.12.1"
-    id("org.jetbrains.dokka")
+    org.jetbrains.dokka
 
     signing
     `maven-publish`
-    id("io.codearte.nexus-staging") version "0.30.0"
+    alias(libs.plugins.nexusStaging)
```
dookon, line 9: Disable plugins for the root project (kord), like:
```kts
    org.jetbrains.dokka apply false
    signing apply false
    `maven-publish` apply false
    alias(libs.plugins.nexusStaging) apply false
```

Lukellmann: Dokka has to be applied to the root project so the `dokkaHtmlMultiModule` task that we use for generating docs at https://kordlib.github.io/kord is added.
I guess `signing` and `maven-publish` can be removed altogether from the root project.
I don't know what the nexus staging plugin does, so I'm not sure about that one.

Lukellmann: I removed the nexus staging plugin, we never used the tasks. The Nexus GUI is used instead. Right @HopeBaron?

.github/workflows/deployment-ci.yml
```diff
@@ -80,8 +80,3 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: -x test publish
-
-# Disabled for now
-#
-#      - name: Close and Release Repository
-#          run: ./gradlew closeAndReleaseRepository
```
DRSchlaubi, line 87: @HopeBaron why was this disabled?

build.gradle.kts
```diff
@@ -18,35 +11,8 @@ repositories {
...
-        outputDirectory.set(dokkaOutputDir)
-    }
-}
-
-apiValidation {
-    ignoredProjects += listOf("ksp-annotations", "ksp-processors")
+tasks.wrapper {
+    gradleVersion = libs.versions.gradle.get()
+    distributionType = ALL
+    distributionSha256Sum = libs.versions.gradleChecksum.get()
```
DRSchlaubi, line 17: Idk whether this is even needed, to update Gradle your should just call `./gradlew wrapper --gradle-version=<version>`

I don't think we need the checksum here, calling the task should save that to the `gradle-wrapper.properties` automatically

Lukellmann: This way you only have to update `gradle`, `gradleChecksum` and run `./gradlew wrapper`.
To achieve the same without you would have to run `./gradlew wrapper --gradle-version <version> --distribution-type all --gradle-distribution-sha256-sum <checksum>`.

But I think you are right, maybe we should only keep:
```kotlin
tasks.wrapper {
    distributionType = ALL
}
```
Then you would only have to run `./gradlew wrapper --gradle-version <version> --gradle-distribution-sha256-sum <checksum>`.

.github/workflows/deployment-ci.yml
Lukellmann: This was added (and already disabled) by @BartArys in #154, so you might have to ask them.

build.gradle.kts
DRSchlaubi: It doesn't update the checksum if you update to a new version, that sounds odd?

DRSchlaubi (approved):

dookon (approved):

.github/workflows/deployment-ci.yml
HopeBaron: Hello, it's because we wanted to remove the manual part of the work, so release on GH = release on nexus
currently we publish to nexus and I approve

HopeBaron: I asked Bart to disable it so we can fix the CI and I never turned it back on since then
We can enable it.

Lukellmann: Now I'm confused. [Here](https://discord.com/channels/556525343595298817/587324906702766226/1042753093781164062) you said we can remove the plugin but here on GH you said we should enable the task (it's from the very same plugin).

HopeBaron: Yes I didn't update you on this one here.
since we sue the maven-publish plugin now I'm looking through the docs for an alternative.
we can remove the plugin because we are not using it anymore

PR #730, opened 2022-11-21
Clean up

Lukellmann:
This PR is a collection of miscellaneous small cleanups.

* [**Remove useless `@Throws` annotation**](https://github.com/kordlib/kord/pull/730/commits/964ac7078271205f73ede1fe11f86c78ba143b6c)
It is only meaningful for Java interop, but since Kord is coroutine-based it's not usable from Java anyway.
see https://kotlinlang.org/docs/java-to-kotlin-interop.html#checked-exceptions
* [**Update stack trace recovery comments**](https://github.com/kordlib/kord/pull/730/commits/381f6c735fe70501dee653b9626bafae819a19e5)
* [**Make all `dev.kord.gateway.Command`s public**](https://github.com/kordlib/kord/pull/730/commits/3867fc0838c1c112d06facb3fb3a92cda92a79fe)
There is no reason that `Command.Heartbeat`, `Identify` and `Resume` were internal.
Changing the visibility to `public` makes them available for things like custom `Gateway` implementations.
* [**Add missing `@KordDsl` annotations**](https://github.com/kordlib/kord/pull/730/commits/cd477c8465c979a4baee875688c815894a5deec2)
* [**Remove `NothingSerializer`**](https://github.com/kordlib/kord/pull/730/commits/835f3e7881893d06a894d8907fae30bfb09bdfaa)
It was unused and kotlinx.serialization is going to add `NothingSerializer()` in the future: https://github.com/Kotlin/kotlinx.serialization/pull/1991
* [**Use sealed `when`s in more places**](https://github.com/kordlib/kord/pull/730/commits/a7f2c620605ee83b6f958f30b1b4731d753c018f)
* [**Deprecate `CommandArgument` convenience extensions**](https://github.com/kordlib/kord/pull/730/commits/7a0e106de5ceda4352b5fbd77f83221f47cbfec1)
This family of functions has inconsistent behavior (some use casts, other coerce the value) and doesn't have a variant for every `ApplicationCommandOptionType`.
@HopeBaron already tried to [remove this family of functions before](https://github.com/kordlib/kord/pull/531/commits/a5bb146c980bee10464fe802168cf5c9864b3e11), but [it was reintroduced](https://github.com/kordlib/kord/pull/531/commits/5f13f84cbfaf25ddc62a5ceb7d15f6dda21b7be0) because some variants were used in tests.
* [**Rename remaining symbols from int to integer**](https://github.com/kordlib/kord/pull/730/commits/e3089beaf04a41c3d744890e2cd8b11a63d87830)
It makes the distinction from `kotlin.Int` clearer.
This has already been done for other symbols in the past (e.g. [here](https://github.com/kordlib/kord/pull/572/commits/37b915544081927e0e49c07519bc65d69e03fb1e)).


dookon:
I think it would be better to define a formatting style for the `/.idea`, to maintain the consistency of the project

Lukellmann:
> I think it would be better to define a formatting style for the `/.idea`, to maintain the consistency of the project

For what aspect of this PR specifically?
Or do you mean for the code style in general?

dookon:
> > I think it would be better to define a formatting style for the `/.idea`, to maintain the consistency of the project
> 
> For what aspect of this PR specifically?
> Or do you mean for the code style in general?

In general, like: auto import, etc

Lukellmann:
I think this should be a separate PR since all the resulting formatting changes would clutter the diff here.

Also using looking into .editorconfig might be interesting for that.

dookon:
> I think this should be a separate PR since all the resulting formatting changes would clutter the diff here.
> 
> Also looking into .editorconfig might be interesting for that.

alright

dookon (approved):

PR #742, opened 2023-01-06
Add missing error codes

NoComment1105:
https://github.com/discord/discord-api-docs/commit/20c66eaaf3952522271fc598ae0f4d3f827d65f3
https://github.com/discord/discord-api-docs/commit/97fd99540d3500c413943cc237440fef025c8be6
https://github.com/discord/discord-api-docs/commit/bfffe3ba76bb7a0b18bcaf5b2a259dd7151f8282
https://github.com/discord/discord-api-docs/commit/e003051cba5dd1f60b6293b858ebd4b3b817879d

PR #743, opened 2023-01-06
Role subscriptions

NoComment1105:
https://github.com/discord/discord-api-docs/commit/e003051cba5dd1f60b6293b858ebd4b3b817879d

The JSON error codes are in #742


HopeBaron:
common.api is conflicting so it would be better to merge this after we add the other PR

NoComment1105:
I can run a quick apiDump if necessary

PR #746, opened 2023-01-06
Add git-hooks for apiCheck

NoComment1105:
This is entirely because I forget and others do too, so it just makes it easier to avoid random build failures. Not necessary, just useful imo

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,10 +2,13 @@ import org.gradle.api.tasks.wrapper.Wrapper.DistributionType.ALL
 
 plugins {
     org.jetbrains.dokka // for dokkaHtmlMultiModule task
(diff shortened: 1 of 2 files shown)


Lukellmann:
Hm, I'm not sure if I want to run `clean` on every commit. Couldn't you instead set up a git hook for yourself?

NoComment1105:
Maybe I'm not sure. I could drop the clean from apiCheck and just make it check the API on each commit?

DRSchlaubi:
I am unsure whether this is a great idea, because running Gradle in a git hook just takes forever and we do have CI for this

HopeBaron:
The point of having the check is compare against the api file on the repo; the defeats the whole point of using it?

PR #755, opened 2023-02-13
Upgrade to Gradle 8.0

Lukellmann:
The changes in `gradle-wrapper.jar`, `gradle-wrapper.properties` and `gradlew` are the result of running this command twice:
```
./gradlew wrapper --gradle-version 8.0 --gradle-distribution-sha256-sum f30b29580fe11719087d698da23f3b0f0d04031d8995f7dd8275a31f7674dc01
```

While touching the build files, the following was also done:

 * use Java toolchains, see https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support

 * use the new compilerOptions DSL introduced in Kotlin 1.8.0, see https://kotlinlang.org/docs/whatsnew18.html#exposing-kotlin-compiler-options-as-gradle-lazy-properties

~Running the `dokkaHtmlMultiModule` task is not possible with Dokka 1.7.20 and Gradle 8.0 due to https://github.com/Kotlin/dokka/issues/2796. This will prevent us from publishing Dokka docs until a fix is released.~ - Using dev build instead.


buildSrc/src/main/kotlin/kord-module.gradle.kts
```diff
@@ -34,16 +37,14 @@ atomicfu {
     dependenciesVersion = libs.findVersion("kotlinx-atomicfu").get().requiredVersion
 }
 
-tasks {
-    withType<JavaCompile> {
-        sourceCompatibility = Jvm.targetString
-        targetCompatibility = Jvm.targetString
-    }
+// the kspKotlin Task is created very late and isn't available via tasks.findByName("kspKotlin")
+val kspKotlin get() = tasks.withType<KspTask>().matching { it.name == "kspKotlin" }
```
DRSchlaubi, line 41: Do we know why this happens? It's very weird, that your method filtering by name works and findByName doesn't?

buildSrc/src/main/kotlin/kord-module.gradle.kts
```diff
@@ -63,14 +64,14 @@ tasks {
     withType<AbstractDokkaLeafTask> {
         // see https://kotlin.github.io/dokka/<dokka version>/user_guide/gradle/usage/#configuration-options
 
-        // make sure ksp generates files before building docs
-        dependsOn(compileKotlin)
+        // include documentation generated by ksp
+        dependsOn(kspKotlin)
```
DRSchlaubi, line 68: Does this have something to do with the new Gradle version?

buildSrc/src/main/kotlin/kord-module.gradle.kts
Lukellmann: I searched in the sources of the ksp gradle plugin and it seems to not register its tasks when the plugin is [applied](https://github.com/google/ksp/blob/1a1818bf6524665c163335ab6e6d70d8ded6167f/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L229) but instead [here](https://github.com/google/ksp/blob/1a1818bf6524665c163335ab6e6d70d8ded6167f/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L383) inside the [`applyToCompilation`](https://github.com/google/ksp/blob/1a1818bf6524665c163335ab6e6d70d8ded6167f/gradle-plugin/src/main/kotlin/com/google/devtools/ksp/gradle/KspSubplugin.kt#L260) function overriding the one in `KotlinCompilerPluginSupportPlugin`, whenever that is called.
The reason why `findByName` doesn't work is that it is not one of the [Task Configuration Avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) APIs that returns a `TaskProvider` but instead returns `Task?` directly. `tasks.withType<KspTask>().matching { it.name == "..." }` seems to be the only lazy way to query a task that might not exist (the `kspKotlin` tasks only exists in the subprojects that actually have a ksp processor in their dependencies).

buildSrc/src/main/kotlin/kord-module.gradle.kts
Lukellmann: Not directly, however the intention was always to only rely on the `kspKotlin` task, not the `compileKotlin` task. And because since Gradle 8.0 you are no longer allowed to have implicit task dependencies (like we had before between `sourcesJar` and `kspKotlin`) I had to have this reference to the `kspKotlin` task anyway and also used it here.

buildSrc/src/main/kotlin/kord-module.gradle.kts
Lukellmann: Ok I was able to solve this, see https://github.com/kordlib/kord/pull/755/commits/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16.

Lukellmann:
Do you have any idea why the build fails with a `ConcurrentModificationException` when [this line](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-module.gradle.kts#L41) is replaced with `withType<KotlinCompile>().configureEach {`?

Lukellmann:
> Do you have any idea why the build fails with a `ConcurrentModificationException` when [this line](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-module.gradle.kts#L41) is replaced with `withType<KotlinCompile>().configureEach {`?
> It doesn't fail [here](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-internal-module.gradle.kts#L15)...

I talked with @DRSchlaubi and it doesn't seem to be easily fixable.
I suggest to just leave it this way and maybe come back to it later.

However, before merging this I suggest to wait for the next Dokka release so we are able to publish docs.

DRSchlaubi:
> > Do you have any idea why the build fails with a `ConcurrentModificationException` when [this line](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-module.gradle.kts#L41) is replaced with `withType<KotlinCompile>().configureEach {`?
> > It doesn't fail [here](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-internal-module.gradle.kts#L15)...
> 
> I talked with @DRSchlaubi and it doesn't seem to be easily fixable. I suggest to just leave it this way and maybe come back to it later.
> 
> However, before merging this I suggest to wait for the next Dokka release so we are able to publish docs.

Are there nightly builds, as https://github.com/Kotlin/dokka/issues/2796 has been fixed for a while?

DRSchlaubi (approved):

Lukellmann:
> > > Do you have any idea why the build fails with a `ConcurrentModificationException` when [this line](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-module.gradle.kts#L41) is replaced with `withType<KotlinCompile>().configureEach {`?
> > > It doesn't fail [here](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-internal-module.gradle.kts#L15)...
> > 
> > I talked with @DRSchlaubi and it doesn't seem to be easily fixable. I suggest to just leave it this way and maybe come back to it later.
> > 
> > However, before merging this I suggest to wait for the next Dokka release so we are able to publish docs.
> 
> Are there nightly builds, as https://github.com/Kotlin/dokka/issues/2796 has been fixed for a while?

idk, would have to dig around in their build setup to be sure

DRSchlaubi:
> > > > Do you have any idea why the build fails with a `ConcurrentModificationException` when [this line](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-module.gradle.kts#L41) is replaced with `withType<KotlinCompile>().configureEach {`?
> > > > It doesn't fail [here](https://github.com/kordlib/kord/blob/00cbc9e73e3451cd3e49eec4e99a1ae1483e2c16/buildSrc/src/main/kotlin/kord-internal-module.gradle.kts#L15)...
> > > 
> > > 
> > > I talked with @DRSchlaubi and it doesn't seem to be easily fixable. I suggest to just leave it this way and maybe come back to it later.
> > > However, before merging this I suggest to wait for the next Dokka release so we are able to publish docs.
> > 
> > 
> > Are there nightly builds, as [Kotlin/dokka#2796](https://github.com/Kotlin/dokka/issues/2796) has been fixed for a while?
> 
> idk, would have to dig around in their build setup to be sure

haven't found anything in the documentation

Lukellmann:
Ok, there are some dev builds: https://github.com/Kotlin/dokka/issues/1425

I suggest we update to one of those, dokka is only used internally, so this won't affect users.

DRSchlaubi:
Yeah sounds good to me

PR #761, opened 2023-02-16
Guild Integration and fix audit log entry create

NoComment1105:
Adds support for integration create and delete, fix integration update and add missing stuff for Audit Log Entry Create

### DIFF
--- a/gateway/src/main/kotlin/Event.kt
+++ b/gateway/src/main/kotlin/Event.kt
@@ -182,6 +182,13 @@ public sealed class Event {
                         DiscordTyping.serializer()
                     ), sequence
                 )
+                "GUILD_AUDIT_LOG_ENTRY_CREATE" -> GuildAuditLogEntryCreate(
+                    decoder.decodeSerializableElement(
+                        descriptor,
+                        index,
+                        DiscordAuditLogEntry.serializer()
+                    ), sequence
+                )
                 "GUILD_CREATE" -> GuildCreate(
                     decoder.decodeSerializableElement(
                         descriptor,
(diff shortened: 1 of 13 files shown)


Lukellmann:
These renames are breaking changes, I think we should do something about that.

NoComment1105:
Ah - Didn't think about that. What do you think should be done?

gateway/src/main/kotlin/Event.kt
```diff
@@ -224,11 +231,25 @@ public sealed class Event {
                         DiscordUpdatedEmojis.serializer()
                     ), sequence
                 )
-                "GUILD_INTEGRATIONS_UPDATE" -> GuildIntegrationsUpdate(
+                "INTEGRATION_CREATE" -> GuildIntegrationCreate(
```
Lukellmann, line 234: oh wait, since this is here i wonder if the event was even received before

if not then the breaking change is fine i think, since it didn't work before anyway

NoComment1105: Ah yes, this confused me for a bit until i found this. The event certainly wasn't being received as my bots logs are filled with "unknown event INTEGRATION_..."

Lukellmann:
Wait there are these two events:
https://discord.com/developers/docs/topics/gateway-events#integration-update
https://discord.com/developers/docs/topics/gateway-events#guild-integrations-update

they are distinct and a different thing

NoComment1105:
Oh no

NoComment1105:
Right I'll make a new pr quickly

Lukellmann:
that's why i think a proper review is desriable before merging prs @HopeBaron

HopeBaron:
I did review but didn't pay attention to the difference between the two as well
apologizes

PR #762, opened 2023-02-16
Silent messages

DRSchlaubi:


rest/src/main/kotlin/builder/message/create/FollowupMessageCreateBuilder.kt
```diff
@@ -33,6 +32,10 @@ public class FollowupMessageCreateBuilder(public val ephemeral: Boolean) :
 
     override val files: MutableList<NamedFile> = mutableListOf()
 
+    override val flags: MessageFlags? = null
```
Lukellmann, line 35: this being a `val` seems useless - you will always get `null` when getting this

HopeBaron (approved):

rest/src/main/kotlin/builder/message/modify/MessageModifyBuilder.kt
```diff
@@ -35,6 +38,18 @@ public sealed interface MessageModifyBuilder {
 
     public var attachments: MutableList<DiscordAttachment>?
 
+    /**
+     * Optional custom [MessageFlags] to add to the message created.
+     *
+     * @see suppressEmbeds
+     */
+    public val flags: MessageFlags?
```
Lukellmann, line 46: is this intended to be a `val` too? (i.e. is there a reason why `flags` shouldn't be set to whatever a user wants?)

rest/src/main/kotlin/builder/message/modify/MessageModifyBuilder.kt
```diff
@@ -101,3 +116,15 @@ public inline fun MessageModifyBuilder.actionRow(builder: ActionRowBuilder.() ->
         it.add(ActionRowBuilder().apply(builder))
     }
 }
+
+/**
+ * Sets/Unsets the [MessageFlags] for this message.
+ *
+ * **Only supports [MessageFlag.SuppressEmbeds]**
+ */
+public inline fun MessageCreateBuilder.messageFlags(builder: MessageFlags.Builder.() -> Unit) {
```
Lukellmann, line 125:
```suggestion
public inline fun MessageModifyBuilder.messageFlags(builder: MessageFlags.Builder.() -> Unit) {
```
this will fail to compile because `flags` is not a `var` in `MessageModifyBuilder`

rest/src/main/kotlin/builder/message/create/MessageCreateBuilder.kt
```diff
@@ -86,6 +107,35 @@ public sealed interface MessageCreateBuilder {
      */
     public suspend fun addFile(path: Path): NamedFile =
         addFile(path.fileName.toString(), ChannelProvider { path.readChannel() })
+
+}
+
+internal fun buildMessageFlags(
+    base: MessageFlags?,
+    suppressEmbeds: Boolean? = null,
```
Lukellmann, line 115:
```suggestion
    suppressEmbeds: Boolean?,
```
this parameter is passed to every call of this function, i think it shouldn't have a default value

rest/src/main/kotlin/builder/message/modify/MessageModifyStateHolder.kt
```diff
@@ -30,4 +29,5 @@ internal class MessageModifyStateHolder {
 
     var components: Optional<MutableList<MessageComponentBuilder>> = Optional.Missing()
 
+    var suppressEmbeds: OptionalBoolean = OptionalBoolean.Missing
```
Lukellmann, line 32: this isn't needed, we don't use the optional value directly but instead work with the nullable boolean to calculate the flags to send, `override var suppressEmbeds: Boolean? = null` is easier to read than `override var suppressEmbeds: Boolean? by state::suppressEmbeds.delegate()`

the same goes for `flags` which could also be removed from `MessageModifyStateHolder`

rest/src/main/kotlin/builder/message/modify/InteractionResponseModifyBuilder.kt
```diff
@@ -40,9 +43,9 @@ public class InteractionResponseModifyBuilder :
                 embeds = state.embeds.mapList { it.toRequest() },
                 allowedMentions = state.allowedMentions.map { it.build() },
                 components = state.components.mapList { it.build() },
-                attachments = state.attachments
-            ),
-            state.files
+                attachments = state.attachments,
+                flags = buildMessageFlags(flags, suppressEmbeds = suppressEmbeds)
```
Lukellmann, line 47:
```suggestion
                flags = buildMessageFlags(flags, suppressEmbeds)
```

Lukellmann (commented):
looks good overall

rest/src/main/kotlin/builder/message/modify/MessageModifyBuilder.kt
Lukellmann: after reviewing the rest of this pr, i'm pretty sure this is supposed to be a var, please change it

rest/src/main/kotlin/builder/message/create/MessageCreateBuilder.kt
```diff
@@ -86,6 +107,35 @@ public sealed interface MessageCreateBuilder {
...
+        add(suppressNotifications, MessageFlag.SuppressNotifications)
+        add(ephemeral, MessageFlag.Ephemeral)
+    }.takeIf { it.code > 0 }
+
+    val allFlags = listOfNotNull(base, flags)
+    return if (allFlags.isEmpty()) {
+        Optional.Missing()
+    } else {
+        Optional(MessageFlags(allFlags))
+    }
```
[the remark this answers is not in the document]
Lukellmann: ah wait, there is a subtle semantic difference (your version allows empty flags when `base` is empty, mine would result in missing), nevermind this suggestion then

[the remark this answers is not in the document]
Lukellmann: ok, i've changed the suggestion to the following semantics:
flags are always sent if any of the parameters is non-null, even if the flags are empty

Lukellmann (approved):
thanks :)

PR #764, opened 2023-02-17
Integration event fixes 2

NoComment1105:
Hopefully I've done this correctly?


gateway/src/main/kotlin/Event.kt
```diff
@@ -686,11 +686,11 @@ public data class GuildBanRemove(val ban: DiscordGuildBan, override val sequence
...
 public data class GuildIntegrationsUpdate(val integrations: DiscordGuildIntegrations, override val sequence: Int?) :
     DispatchEvent()
-public data class IntegrationDelete(val integration: DiscordGuildIntegrationsDeleted, override val sequence: Int?) :
+public data class IntegrationDelete(val integration: DiscordIntegrationsDeleted, override val sequence: Int?) :
     DispatchEvent()
-public data class IntegrationCreate(val integration: DiscordGuildIntegrations, override val sequence: Int?) :
+public data class IntegrationCreate(val integration: DiscordIntegrations, override val sequence: Int?) :
     DispatchEvent()
-public data class IntegrationUpdate(val integration: DiscordGuildIntegrations, override val sequence: Int?) :
+public data class IntegrationUpdate(val integration: DiscordIntegrations, override val sequence: Int?) :
```
HopeBaron, line 693: Shoulldn't this be a DiscordIntegration?

NoComment1105: Its an extra class to wrap in the extra guild I'd field with this event. See lines 874 to EOF

Lukellmann: I think the json looks like this:
```json
{
  ... // normal integration fields
  "guild_id" : "1234"
}
```
but look at trace logs to be sure

NoComment1105: I can't find a way to trigger these events, since it requires an affiliated twitch/youtube account. or role subscriptions. How should I test this?

Lukellmann: doesn't changing an app's/bot's name trigger an integration update? or enabling/disabling an integration

NoComment1105: Maybe, let me try

PR #769, opened 2023-02-23
Add ability to receive Unknown events

DRSchlaubi:
/cc @gdude2002

### DIFF
--- a/core/src/main/kotlin/event/Event.kt
+++ b/core/src/main/kotlin/event/Event.kt
@@ -65,7 +65,6 @@ public class UnknownEvent(
     override val shard: Int,
     override val customContext: Any?,
 ) : Event {
-    override fun toString(): String {
-        return "UnknownEvent(kord=$kord, shard=$shard, customContext=$customContext, data=$data, name=$name)"
-    }
+    override fun toString(): String =
+        "UnknownEvent(name=$name, data=$data, kord=$kord, shard=$shard, customContext=$customContext)"
 }
--- a/core/src/main/kotlin/event/Event.kt
+++ b/core/src/main/kotlin/event/Event.kt
@@ -5,7 +5,13 @@ import dev.kord.core.Kord
 import dev.kord.core.KordObject
 import dev.kord.core.entity.Strategizable
 import dev.kord.gateway.Gateway
+import kotlinx.serialization.json.JsonElement
 
+/**
+ * Representation of an event received from the Discord gateway.
+ *
+ * @see UnknownEvent
+ */
 public interface Event : KordObject {
     /**
      * The Gateway that spawned this event.
@@ -45,3 +51,17 @@ public interface Event : KordObject {
     @KordPreview
     public val customContext: Any?
 }
+
+/**
+ * Representation of an event that has not been documented / added to Kord yet.
+ *
+ * @property data the raw [json body][JsonElement] of this event if provided
+ * @property name the name of the event if provided
+ */
+public data class UnknownEvent(
+    override val kord: Kord,
+    override val shard: Int,
+    override val customContext: Any?,
+    val data: JsonElement?,
+    val name: String?
+) : Event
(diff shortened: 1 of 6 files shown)


gdude2002 (approved):
LGTM. Appreciate this.

core/src/main/kotlin/event/Event.kt
```diff
@@ -45,3 +51,17 @@ public interface Event : KordObject {
...
     public val customContext: Any?
 }
+
+/**
+ * Representation of an event that has not been documented / added to Kord yet.
+ *
+ * @property data the raw [json body][JsonElement] of this event if provided
+ * @property name the name of the event if provided
+ */
+public data class UnknownEvent(
```
Lukellmann, line 61: other events aren't data classes, `equals`, `hashCode` and `copy` don't make much sense for events
i think we should instead just provide a custom `toString` implementation

core/src/main/kotlin/gateway/handler/DefaultGatewayEventInterceptor.kt
```diff
@@ -36,6 +36,7 @@ public class DefaultGatewayEventInterceptor @KordPreview public constructor(
     public constructor() : this(customContextCreator = null)
 
     private val listeners = listOf(
+        UnknownEventHandler(),
```
Lukellmann, line 39: i ordered the other handlers alphabetically to have the same order as the files in the ide, could you put it in the right place?

core/src/main/kotlin/event/Event.kt
```diff
@@ -45,3 +51,17 @@ public interface Event : KordObject {
...
+    override val kord: Kord,
+    override val shard: Int,
+    override val customContext: Any?,
+    val data: JsonElement?,
+    val name: String?
```
Lukellmann, line 66: i think `name` should be the first and `data` the second parameter an they should be put before `kord`, `shard` and `customContext` to match other events

gateway/src/main/kotlin/Event.kt
```diff
@@ -805,6 +805,12 @@ public data class GuildScheduledEventUserRemove(
     override val sequence: Int?,
 ) : DispatchEvent()
 
+public data class UnknownDispatchEvent(
+    val data: JsonElement?,
+    override val sequence: Int?,
+    val name: String?
```
Lukellmann, line 811: i think `name` should be the first parameter

Lukellmann (commented):
only some minor things that i think should be adressed but looks good overall

core/src/main/kotlin/gateway/handler/DefaultGatewayEventInterceptor.kt
DRSchlaubi: sure

core/src/main/kotlin/event/Event.kt
```diff
@@ -45,3 +51,21 @@ public interface Event : KordObject {
...
+public class UnknownEvent(
+    public val name: String?,
+    public val data: JsonElement?,
+    override val kord: Kord,
+    override val shard: Int,
+    override val customContext: Any?,
+) : Event {
+    override fun toString(): String {
+        return "UnknownEvent(kord=$kord, shard=$shard, customContext=$customContext, data=$data, name=$name)"
+    }
```
Lukellmann, line 70:
```suggestion
    override fun toString(): String =
        "UnknownEvent(name=$name, data=$data, kord=$kord, shard=$shard, customContext=$customContext)"
```
to match the order of the constructor parameters

gateway/src/main/kotlin/Event.kt
```diff
@@ -491,6 +535,7 @@ public sealed class Event {
                     ),
                     sequence
                 )
+
```
Lukellmann, line 538: could you remove the line breaks again? i think `)` having their own lines is enough spacing between branches (maybe adjust your formatter)

Lukellmann (commented):
almost approved :)

Lukellmann (approved):
looks good, just gonna write up a commit message and then merge
thanks :)

PR #770, opened 2023-02-23
Update CI

DRSchlaubi:


Lukellmann:
i think the build fails because of the toolchain specified - we expect to have a jdk 8 available but since you changed it to 19 in the ci config, it doesn't find a jdk 8
is there a reason why we shouldn't build with 8?

DRSchlaubi:
> i think the build fails because of the toolchain specified - we expect to have a jdk 8 available but since you changed it to 19 in the ci config, it doesn't find a jdk 8 is there a reason why we shouldn't build with 8?

No, that's not what happens, but now I know what happens thx

We install Oracle JDK 19, however we set 8 as a toolchain, Gradle supplies that using Temurin, which doesn't support -Xjdk-version.

Lukellmann:
i think the problem is that the `--release` flag only exists [since jdk 9](https://openjdk.org/jeps/247) and not the jdk vendor

NoComment1105 (approved):

.github/workflows/docs-ci.yml
```diff
@@ -16,23 +26,29 @@ jobs:
...
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v1
```
Lukellmann, line 54: could you explain what exactly this does? or just provide some reference where i could look it up myself?

DRSchlaubi: https://github.com/actions/starter-workflows/blob/main/pages/nuxtjs.yml

DRSchlaubi: It supposedly does some configuration, then uploads it to githubs cdn and then triggers a deployment, that way the gh.actions branch should be no longer needed

.github/workflows/docs-ci.yml
```diff
@@ -5,6 +5,16 @@ on:
   push:
     branches:
       - 0.8.x
+      -
```
Lukellmann, line 8:
```suggestion

```

.github/workflows/docs-ci.yml
```diff
@@ -5,6 +5,16 @@ on:
...
+permissions:
+  contents: read
+  pages: write
+  id-token: write
```
Lukellmann, line 12: why is the `id-token` permission needed?

.github/workflows/docs-ci.yml
```diff
@@ -16,23 +26,29 @@ jobs:
...
-          git-config-name: GitHub Actions
-          git-config-email: actions@github.com
-          commit-message: Update docs
+          path: build/dokka/htmlMultiModule
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
```
Lukellmann, line 50:
```suggestion
    needs: update_docs
```
`update_docs` could instead be renamed

.github/workflows/docs-ci.yml
Lukellmann: because of this: https://github.com/actions/deploy-pages#security-considerations

.github/workflows/docs-ci.yml
DRSchlaubi: The entire folder the file is in

.github/workflows/docs-ci.yml
DRSchlaubi: Yes

Lukellmann (approved):

Lukellmann:
do you know if there is an easy way to fix [this](https://github.com/kordlib/kord/actions/runs/4263982724/jobs/7421519902)? or should we just go back to `JamesIves/github-pages-deploy-action`?

DRSchlaubi:
I will look at it asap, just seems like we need to change one constant in the file

PR #772, opened 2023-02-23
Fix various bugs in `DiscordBitSet`

Lukellmann:
* `WIDTH` was defined to the width of a `Byte` not a `Long`, resulting in bugs in `size`, `get` and `set`

* `binary` just did some completely wrong things that I can't really describe

* `hashCode` could return different values for equal `DiscordBitSet`s (when they had different amounts of trailing zeros in the `data` array), breaking its general contract

* `equals` was comparing `this` with `this` and not `this` with `other`, which resulted in any two `DiscordBitSets` considered equal

* `contains` would erroneously return `false` if `other` did have more trailing zeros but would otherwise be contained

* `set` wasn't growing `data` correctly and also didn't unset bits properly

### DIFF
--- a/common/src/main/kotlin/DiscordBitSet.kt
+++ b/common/src/main/kotlin/DiscordBitSet.kt
@@ -13,19 +13,20 @@ import kotlin.math.max
 import kotlin.math.min
 
 private const val SAFE_LENGTH = 19
-private const val WIDTH = Byte.SIZE_BITS
+private const val WIDTH = Long.SIZE_BITS
 
 @Suppress("FunctionName")
-public fun EmptyBitSet(): DiscordBitSet = DiscordBitSet(0)
+public fun EmptyBitSet(): DiscordBitSet = DiscordBitSet()
 
 @Serializable(with = DiscordBitSetSerializer::class)
-public class DiscordBitSet(internal var data: LongArray) {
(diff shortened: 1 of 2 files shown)


Lukellmann:
please tell me if you happen to find any other bugs or if you think my fixes still include bugs

DRSchlaubi:
Don't know enough about this to review this properly

HopeBaron:
The re-write is perfectly fine much easier to read as well.
But the initial reason we used a byte size is to save ourselves from allocating non-necessary bits

Lukellmann:
> But the initial reason we used a byte size is to save ourselves from allocating non-necessary bits

yeah but since the implementation uses a `LongArray` rn, setting `WIDTH` to `Byte.SIZE_BITS` was simply wrong

HopeBaron:
You are right, let's just keep it long then

Lukellmann:
did some last small changes, see [here](https://github.com/kordlib/kord/compare/d6b501975bb0b4e328f081b9f1c253ef83893ad0..8cddb916424dec329db4bddcd2359927fa8059bc)

HopeBaron:
They look good, if there's nothing else you want to do on this PR we can merge

Lukellmann:
i'm done here 👍

PR #775, opened 2023-02-28
Support Kotlin/JS

DRSchlaubi:
This PR attempts to provide Kotlin/JS support for Kord (see #69)

# Breaking Changes
* The following member functions were replaced by extension functions. Importing the extension functions should fix this breaking change.
  * `dev.kord.common.Locale.asJavaLocale` -> `import dev.kord.common.asJavaLocale`
  * `dev.kord.core.cache.KordCacheBuilder.lruCache` -> `import dev.kord.core.cache.lruCache`
  * `dev.kord.rest.builder.message.create.MessageCreateBuilder.addFile` -> `import dev.kord.rest.builder.message.create.addFile`
  * `dev.kord.rest.builder.message.modify.MessageModifyBuilder.addFile` -> `import dev.kord.rest.builder.message.modify.addFile`
  * `dev.kord.rest.request.RequestBuilder.file` -> `import dev.kord.rest.request.file`
* The type of the following declarations was changed from `java.util.Locale` to `dev.kord.common.Locale` (if you really need to use the Java class, use the `asJavaLocale` extension):
  * `dev.kord.core.entity.Guild.preferredLocale`
  * `dev.kord.rest.builder.guild.GuildModifyBuilder.preferredLocale`
* All annotations in `dev.kord.common.annotation` and the `dev.kord.gateway.PrivilegedIntent` annotation now have `BINARY` instead of `RUNTIME` retention. This means they are no longer visible for reflection.
* All maven consumers must append `-jvm` to their artifact IDs for `kord-common`, `kord-rest`, `kord-gateway` and `kord-core`.
* The `core-voice` capability has been removed, please use the `kord-core-voice` artifact instead.


common/src/jsMain/kotlin/dev/kord/common/Platform.kt
```diff
@@ -0,0 +1,13 @@
...
+public actual object Platform {
+    public actual val IS_JVM: Boolean = false
+    public actual val IS_NODE: Boolean
+        get() = js(
+            "typeof process !== 'undefined' && process.versions != null && process.versions.node != null"
+        ) as Boolean
+    public actual val IS_BROWSER: Boolean
+        get() = js(
+            "typeof window !== 'undefined' && typeof window.document !== 'undefined' || typeof self !== 'undefined' && typeof self.location !== 'undefined'"
+        ) as Boolean
```
Lukellmann, line 12: i'm curious, where do you have this cursed logic from?

common/src/jvmMain/kotlin/Platform.kt
```diff
@@ -0,0 +1,7 @@
+package dev.kord.common
+
+public actual object Platform {
+    public actual val IS_JVM: Boolean = true
+    public actual val IS_NODE: Boolean = false
+    public actual val IS_BROWSER: Boolean = false
```
Lukellmann, line 6:
```suggestion
    public actual const val IS_JVM: Boolean = true
    public actual const val IS_NODE: Boolean = false
    public actual const val IS_BROWSER: Boolean = false
```
will then be inlined

common/src/nonJvm/kotlin/ConcurrentHashMap.kt
```diff
@@ -0,0 +1,7 @@
+package dev.kord.common
+
+import co.touchlab.stately.collections.ConcurrentMutableMap
+import dev.kord.common.annotation.KordInternal
+
+@KordInternal
+public actual typealias ConcurrentHashMap<K, V> = ConcurrentMutableMap<K, V>
```
Lukellmann, line 7: i just remembered that js is single-threaded, wouldn't `MutableMap` be also ok as the actual type?

Lukellmann (commented):
i've looked at common so far - looks good overall

can you also test if `dokkaHtmlMultiModule` can be run? i wasn't able to

buildSrc/src/main/kotlin/Compiler.kt
```diff
@@ -1,21 +1,54 @@
...
+private fun String.asOptIn() = "-opt-in=$this"
+
 object Jvm {
     const val target = 8
 }
 
-fun KotlinJvmCompilerOptions.applyKordCompilerOptions() {
-    allWarningsAsErrors.set(true)
+fun KotlinCommonCompilerOptions.applyKordCompilerOptions() {
+    allWarningsAsErrors.set(false)
```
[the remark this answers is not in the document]
DRSchlaubi: I think there was one bug in the js compiler that we cannot fix rn which causes a warning which then causes compilation to fail, will check this again later

buildSrc/src/main/kotlin/Documentation.kt
```diff
@@ -0,0 +1,54 @@
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.dokka.gradle.AbstractDokkaLeafTask
+import java.net.URL
+
+fun TaskContainer.configureDokka(additionalConfiguration: AbstractDokkaLeafTask.() -> Unit = {}) {
+    withType<AbstractDokkaLeafTask>().configureEach {
+        // see https://kotlin.github.io/dokka/<dokka version>/user_guide/gradle/usage/#configuration-options
```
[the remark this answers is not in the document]
DRSchlaubi: What is no longer valid?

common/src/jsMain/kotlin/dev/kord/common/Platform.kt
DRSchlaubi: https://github.com/ktorio/ktor/blob/23261519a5621a3401ab8b6db18740ad01f03b79/ktor-utils/js/src/io/ktor/util/PlatformUtilsJs.kt

common/src/nonJvm/kotlin/ConcurrentHashMap.kt
DRSchlaubi: Doesn't nodejs support multithreading?

buildSrc/src/main/kotlin/Compiler.kt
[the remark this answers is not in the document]
DRSchlaubi: Might be tracked as https://youtrack.jetbrains.com/issue/KT-51110/Pseudo-circular-dependency-causes-duplicate-library-name-warning-for-JS-target

DRSchlaubi:
Haha we have negative teamCity credits

For dokka: as core has a submodule a separate task `core:dokkaHtmlMultimodule` has been created which tried to generate docs for the `live-tests` module, which doesn't use dokka.

So using `:dokkaHtmlMultiModule` would have worked as it only runs for root. However this behavior has now been fixed and you should be able to run dokka again

common/src/jsMain/kotlin/dev/kord/common/Platform.kt
Lukellmann: and the browser part?

gateway/src/commonMain/kotlin/DefaultGateway.kt
```diff
@@ -179,21 +179,11 @@ public class DefaultGateway(private val data: DefaultGatewayData) : Gateway {
...
-            it.write(data)
-        }
-
-        return outputStream.use {
-            String(outputStream.toByteArray(), 0, outputStream.size(), Charsets.UTF_8)
-        }
-    }
-
     private suspend fun read(frame: Frame) {
+        defaultGatewayLogger.trace { "Received raw frame: $frame" }
```
Lukellmann, line 183: hm, should we add additional logging in the scope of this mpp pr?

Lukellmann: or rather: why did you add this?

rest/src/test/kotlin/request/StackTraceRecoveryTest.kt
```diff
@@ -1,55 +0,0 @@
...
-import io.ktor.client.*
-import io.ktor.client.engine.mock.*
-import io.ktor.http.*
-import io.ktor.util.*
-import io.ktor.utils.io.*
-import kotlinx.coroutines.test.runTest
-import kotlin.test.Test
-import kotlin.test.assertEquals
-
-class StackTraceRecoveryTest {
```
Lukellmann, line 14: why was this test deleted?

common/src/commonMain/kotlin/Platform.kt
```diff
@@ -0,0 +1,10 @@
+package dev.kord.common
+
+import dev.kord.common.annotation.KordInternal
+
+@KordInternal
+public expect object Platform {
```
Lukellmann, line 6: should we move this to `test-kit` while we don't need it in actual kord code?

this should eliminate these warnings (they seem to come up because of the dependency on `projects.common` in `test-kit` for access to this object):
```
> Task :common:compileTestKotlinJs                                                                                                                                                                                                                                                                                  
w: duplicate library name: kord:common 

> Task :common:compileTestDevelopmentExecutableKotlinJs                                                                                                                                                                                                                                                             
w: duplicate library name: kord:common
```

buildSrc/src/main/kotlin/Documentation.kt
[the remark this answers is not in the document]
Lukellmann: they moved the documentation to https://kotlinlang.org/docs/dokka-introduction.html instead

common/src/commonMain/kotlin/DiscordBitSet.kt
```diff
@@ -27,9 +32,10 @@ public class DiscordBitSet(internal var data: LongArray) { // data is in little-
     public val value: String
         get() {
             // need to convert from little-endian data to big-endian expected by BigInteger
-            val buffer = ByteBuffer.allocate(data.size * Long.SIZE_BYTES)
-            buffer.asLongBuffer().put(data.reversedArray())
-            return BigInteger(buffer.array()).toString()
+            return withBuffer(data.size * Long.SIZE_BYTES) {
+                writeFully(data.reversedArray())
+                formatIntegerFromByteArray(readBytes())
+            }
```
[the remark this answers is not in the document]
Lukellmann: i changed this to have the same code as before on jvm

core/src/commonTest/kotlin/performance/KordEventDropTest.kt
```diff
@@ -63,7 +68,9 @@ class KordEventDropTest {
     )
 
     @Test
-    fun `hammering the gateway does not drop core events`() = runBlocking {
+    @JsName("test1")
+    @IgnoreOnJs // currently this doesn't work on JS for some reason
+    fun `hammering the gateway does not drop core events`() = runTest {
```
Lukellmann, line 73: ok, it seems like the code is just too slow on js, is there a way to increase the timeout for kotlin js tests?

Lukellmann: yes you can, see https://github.com/kordlib/kord/pull/775/commits/7eb4a645503a49cdd2cdead337d02889147d2238

gateway/src/commonMain/kotlin/DefaultGateway.kt
DRSchlaubi: For additional debugging abilities of decompression. It might be smart to add this, as we know have to maintain two compression implementations

Lukellmann: sure 👍

rest/src/test/kotlin/request/StackTraceRecoveryTest.kt
Lukellmann: it was restored 👍

Lukellmann (approved):

Lukellmann:
about annotation retention, this is how kxser does it, should we do it like this too?
```kt
@MustBeDocumented
@Target(AnnotationTarget.PROPERTY, AnnotationTarget.CLASS, AnnotationTarget.TYPE)
//@Retention(AnnotationRetention.RUNTIME) // Runtime is the default retention, also see KT-41082
public annotation class Serializable(
    val with: KClass<out KSerializer<*>> = KSerializer::class // Default value indicates that auto-generated serializer is used
)
```
it would mean that it's still available at runtime on jvm at least, i.e. no breaking change

DRSchlaubi:
IMO these annotations shouldn't have been on runtime to begin with

Lukellmann:
> IMO these annotations shouldn't have been on runtime to begin with

but now they are, so why not keep it that way?

Lukellmann:
https://github.com/kordlib/kord/pull/659

DRSchlaubi:
I don't really see the benefit of keeping it

Lukellmann:
alright, it's fine for me

PR #776, opened 2023-03-02
fix(voice): use 74 byte ip discovery packet instead of the deprecated 70 byte discovery packet

lost-illusi0n:
This commit is in response to the following message from Discord Developers.
image

It was common use to use this deprecated 70 byte IP discovery packet and this commit updates to the new 74-byte version. In addition, this commit reuses a 66 zero-initialized byte array for the empty data field on the request rather than creating a new one on every IP discovery.

### DIFF
--- a/voice/src/main/kotlin/udp/GlobalVoiceUdpSocket.kt
+++ b/voice/src/main/kotlin/udp/GlobalVoiceUdpSocket.kt
@@ -14,6 +14,9 @@ import kotlin.text.String
 
 private val globalVoiceSocketLogger = KotlinLogging.logger { }
 
+private const val REQUEST: Short = 0x01
+private const val RESPONSE: Short = 0x02
+
 /**
  * A global [VoiceUdpSocket] for all [dev.kord.voice.VoiceConnection]s, unless specified otherwise.
  * Initiated once and kept open for the lifetime of this process.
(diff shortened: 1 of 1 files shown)


voice/src/main/kotlin/udp/GlobalVoiceUdpSocket.kt
```diff
@@ -40,12 +45,16 @@ public object GlobalVoiceUdpSocket : VoiceUdpSocket {
...
+            writeShort(70) // the length of the rest of the message
             writeInt(ssrc)
-            writeFully(ByteArray(66))
+            writeFully(EMPTY_DATA)
         })
 
         return with(receiveFrom(address).packet) {
-            discard(4)
+            require(readShort() == RESPONSE) { "did not receive a response." }
+            discardExact(6) // length (2) + ssrc (4)
```
Lukellmann, line 56: shouldn't this be checked too instead of being discarded?

lost-illusi0n: we could check the length to be 66, could give a better error message in the off chance discord messes up. ssrc would stay discarded though as in the previous behavior. theres nothing to do with it here.

Lukellmann: yeah, whatever, probably not worth the effort

PR #778, opened 2023-03-02
Fix docs-ci

DRSchlaubi:
Taken from: https://github.com/actions/starter-workflows/blob/main/pages/static.yml


Lukellmann (approved):
let's hope this works :D

PR #784, opened 2023-03-13
Add a nullable variant of `getAuthorAsMember`

NoComment1105:


Lukellmann (commented):
looks good, but shouldn't the `getAuthorAsMember` function be deprecated then?

NoComment1105:
Does it need to be? There are nullable and non null versions of a lot of other functions

Lukellmann:
yeah those usually do different things, these two have the same code

Lukellmann:
yeah, i think it's right to deprecate the other one, you usually want to have `OrNull` semantics (no exceptions as far as reasonable) either all the way or not at all, the existing function is a mix of both, your new one is clearly using `OrNull` semantics

Lukellmann (approved):
thanks :)

PR #793, opened 2023-03-24
Fix breaking changes and bugs before release

Lukellmann:
Some breaking changes were discovered since the last release (e.g. [here](https://discord.com/channels/556525343595298817/1014052126248423434/1087696229900750848)). This tries to revert those and includes some additional fixes.

* The overload resolution ambiguity reported [here](https://discord.com/channels/556525343595298817/1014052126248423434/1087696229900750848) was fixed.

* `NewsChannelBehavior` was missing `TopGuildMessageChannelBehavior` as a supertype since `ThreadParentChannelBehavior` is no longer a `TopGuildMessageChannelBehavior`.

* `TextChannelBehavior` was missing `TopGuildMessageChannelBehavior` as a supertype since `PrivateThreadParentChannelBehavior` is no longer a `TopGuildMessageChannelBehavior`.

* `ThreadParentChannelBehavior` was extended to be a `CategorizableChannelBehavior` (implicating that `ThreadParentChannel` was also extended to be a `CategorizableChannel`).

* `TextChannelBehavior`'s `startPublicThread` and `startPrivateThread` had an additional `reason` parameter added, breaking binary compatibility. The reason can be specified in the builder lambda.

* `TextChannelBehavior.startPublicThreadWithMessage` had the `archiveDuration` parameter removed, breaking binary compatibility.

* Binary incompatibility in `Integration` was fixed.

* Missing `Optional` default values were added in `IntegrationData`.

* The event type for `ForumChannel`s was changed from `UnknownChannel[Create|Update|Delete]Event` to `ForumChannel[Create|Update|Delete]Event`.

* Some optionalities and nullabilities were fixed.

* Integration events now use `Integration` entities from core instead of `DiscordIntegration` from common.

* Audit Log Entries from `GuildAuditLogEntryCreate` events are no longer cached. They aren't queried through cache so putting them there doesn't make sense.

* Wrong `SerialName` annotation for `components` was removed.

* Old constructor of `StartThreadWithMessageBuilder` was reintroduced to preserve binary compatibility.

* `StartThreadWithoutMessageBuilder` was renamed back to `StartThreadBuilder` to preserve binary compatibility.

* Forum Tag routes and related declarations were removed because they are not documented in Discord's docs and have odd behavior. See the discussion [here](https://discord.com/channels/556525343595298817/1014052126248423434/1088873517153529887) for details.

* The serializer of `GuildMemberFlags` was made internal, it is accessible through `GuildMemberFlags.serializer()`.

* Some minor things were refactored and simplified.

* The `asChannel`, `asChannelOrNull`, `fetchChannel`, `fetchChannelOrNull` and `withStrategy` methods were overridden in `ForumChannel` and `ForumChannelBehavior` so they have the correct return type.

* The `Intents.IntentsBuilder.enableEvent` function was updated to add the right intent for `ForumChannel` events.

* Support for webhooks was moved up from `TopGuildMessageChannelBehavior` to `CategorizableChannelBehavior`. This was done because `ForumChannel`s also support webhooks but are not `TopGuildMessageChannel`s. The only concrete `Channel` type that received webhook support by this change is `ForumChannel`.

* The `Discord` prefix was removed from `DiscordForumTag` and `DiscordDefaultReaction`. These types are used in core and do not contain any optional fields.

* `ForumTagBuilder` is no longer a `AuditRequestBuilder`, its `reason` was unused. The `name` property was also made `var` and `public`.

* Missing properties of `ThreadChannel`, `ForumChannel` and `TextChannel` were added.

* Message count caching was updated. Updating `lastMessageId` and message counts was combined into one update operation for `MessageCreate` and `totalMessageSent` is no longer decremented on `MessageDelete` and `MessageDeleteBulk`, the whole point of it is to not decrement.

* The default value for the `builder` paremeter of `ForumChannelBehavior.startPublicThread` was removed. Not having a default empty builder lambda in will nudge users into thinking about what belongs in there. This is needed because you have to provide a message in that lambda to successfully create a forum post.

* Removed `ForumThreadModifyBuilder`, it was almost the same as `ThreadModifyBuilder`, just with an additional `flags` property, which was just copied over to `ThreadModifyBuilder` instead.
  Having a single builder for both cases is more maintainable and keeps the API simpler.

* `StartForumThreadBuilder.message` was made non-nullable, as it is always required. `StartForumThreadBuilder.createMessage` was also renamed to `message` because a thread is started that just contains a message.

* The `builder` parameter of `ChannelService.startThread` was accidentally unused.

The full diff between the last release and this PR can be seen here to verify that the breaking changes were in fact reverted: https://github.com/kordlib/kord/compare/0.8.0-M17...0.8.0-fixes


Lukellmann:
I want to make sure that the breaking changes were reverted by looking at the full diff in detail. I will tell you if I'm done with that and this PR is ready to merge (however, it is already ready to review).

PR #794, opened 2023-03-25
Bump `DeprecationLevel`s

Lukellmann:
`WARNING` -> `ERROR`
`ERROR` -> `HIDDEN`
`HIDDEN` -> remove

### DIFF
--- a/common/api/common.api
+++ b/common/api/common.api
@@ -222,8 +222,8 @@ public abstract class dev/kord/common/entity/ActivityType {
 	public final fun getCode ()I
 	public final fun hashCode ()I
 	public final fun toString ()Ljava/lang/String;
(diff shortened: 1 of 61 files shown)


Lukellmann:
`0.9.x` is the right target for this pr, right?

HopeBaron:
Yes

Lukellmann:
> Yes - Let's just wait for the MPP merge in a few first

on the cache project or here?

HopeBaron:
Cache MPP > Kord MPP > Deprecation

Lukellmann:
tbh i don't see why this should wait for mpp, imho it's better to have this as soon as possible after the release, so that we can base the further development on it

HopeBaron:
Hmmm I think conflicts would be minimal if any

Lukellmann:
> Hmmm I think conflicts would be minimal if any

that's what i thought too

PR #795, opened 2023-03-25
Add support for soundboard permissions

NoComment1105:
https://github.com/discord/discord-api-docs/pull/6017


common/src/main/kotlin/entity/Permission.kt
```diff
@@ -327,6 +338,7 @@ public sealed class Permission(public val code: DiscordBitSet) {
                 ManageRoles,
                 ManageWebhooks,
                 ManageEmojisAndStickers,
```
Lukellmann, line 340: this should be replaced by the renamed one

NoComment1105: fixed

Lukellmann (approved):
thanks :)
will hold of the merge until the referenced pr is merged

PR #796, opened 2023-03-25
Fix `ClassCastException` with `getParent` for forums

Lukellmann:
The bug can be reproduced with this code:

```kt
val thread = forumChannel.startPublicThread("forum thread name") {
    message("message content")
}
val parent = thread.getParent() // CCE here
println(parent)
```

The cause was that `ForumChannelBehavior.startPublicThread` returns a `TextChannelThread` that was originally designed to be a thread whose parent can only be a `TextChannel`. It therefore overwrote the `getParent` and `getParentOrNull` methods to return `TextChannel`s via casting.

This is fixed by removing the overwrites. It can be a small breaking change for code that assumed to receive a `TextChannel` when calling those methods.


NoComment1105 (approved):
Sweet lgtm

PR #798, opened 2023-03-29
BaseGateway abstraction

viktorhrekh:
Abstract out some code to base abstract class

### DIFF
--- a/gateway/src/main/kotlin/ConnectionManagedGateway.kt
+++ b/gateway/src/main/kotlin/ConnectionManagedGateway.kt
@@ -19,8 +19,10 @@ import java.net.URI
 import kotlin.coroutines.CoroutineContext
 import kotlin.time.Duration
 
+internal typealias ConnectionProvider = suspend () -> GatewayConnection
+
 internal class ConnectionManagedGateway(
-    private val connectionProvider: suspend () -> GatewayConnection,
+    private val connectionProvider: ConnectionProvider,
     private val data: DefaultGatewayData
 ) : BaseGateway() {
 


NoComment1105 (approved):

Lukellmann:
I like the idea of revisiting the `Gateway` implementation, however I thought of a somewhat different seperation.

I think of the `Gateway` interface as something that manages [connections](https://discord.com/developers/docs/topics/gateway#connections) and decides when to stop or restart them. It therefore seems obvious to seperate the logic into a connection class/interface that has a clearly defined lifecycle (open ws connection with either identify or resume payload -> receive and send events -> when requested to stop (by discord or the user) or some error occurrs close the connection and end) after which is has fulfilled its purpose. `Gateway` would then simply manage the connection (keep track of the active one, stop it, start a new one when appropiate).

I already started to implement the idea ([here](https://discord.com/developers/docs/topics/gateway#connections)), however it's just in its early stages, so I think it makes sense to coordinate efforts here.

viktorhrekh:
@Lukellmann 

Can we have a post on forum in Discord for that? Would be easier to discuss I guess :)

Lukellmann:
Sure.

viktorhrekh:
@Lukellmann @HopeBaron This is done I guess?

Lukellmann:
i'd still like to make an in depth review.
also some tests would be nice

gateway/src/main/kotlin/ConnectionManagedGateway.kt
```diff
@@ -0,0 +1,128 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import kotlinx.serialization.json.Json
+import java.net.URI
+import kotlin.coroutines.CoroutineContext
+import kotlin.time.Duration
+
+internal class ConnectionManagedGateway(
+    private val connectionProvider: suspend () -> GatewayConnection,
```
HopeBaron, line 23: I don't like to see such declarations in code as it doesn't really give a hint on what it is used for outside of this context.
Provide a typealias for it please.

gateway/src/main/kotlin/DefaultGateway.kt
```diff
@@ -61,21 +48,16 @@ public data class DefaultGatewayData(
...
+public class DefaultGateway(private val data: DefaultGatewayData) : BaseGateway() {
 
     private val compression: Boolean
-
     private val _ping = MutableStateFlow<Duration?>(null)
-    override val ping: StateFlow<Duration?> get() = _ping
 
-    override val events: SharedFlow<Event> = data.eventFlow
+    override val ping: StateFlow<Duration?> get() = _ping
+    override val events: MutableSharedFlow<Event> = data.eventFlow
```
HopeBaron, line 57: We should be aware that this now exposes a MutableSharedFlow. which means it can be interred with.
While I'm not against it since it would mean the user of Kord may emit their own custom Events through.

Unless I'm missing something.

HopeBaron (changes requested):
It looks good overall just a few comments. Will help me expand further

gateway/src/main/kotlin/connection/DefaultGatewayConnection.kt
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+    override suspend fun open(data: Data): CloseReason {
+        if (!atomicState.compareAndSet(State.Uninitialized, State.Opening)) errorInvalidState()
+        this.data = data
+        val url = Url(data.uri)
+        val isZLibCompressed = url.parameters.contains("compress", "zlib-stream")
+        inflater = if (isZLibCompressed) FrameInflater.ZLib() else FrameInflater.None
+
+        coroutineScope {
```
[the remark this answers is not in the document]
viktorhrekh: > what's the point of this coroutineScope here?

Look at 133, 163, 170. We are launching some jobs in the scope, and then, if the connection is closed, it will await until the jobs are done, before returning the close result

gateway/src/main/kotlin/DefaultGateway.kt
viktorhrekh: I guess if someone would relay on implementation rather than the interface they should be aware of what they are doing and the flexibility should not harm?

HopeBaron: I didn't say it's a bad change I'm just pointing out the change :D

HopeBaron (approved):

PR #799, opened 2023-03-29
Make use of Ktor websocket extensions and serialization

DRSchlaubi:


DRSchlaubi:
Only glanced at https://github.com/kordlib/kord/pull/798 and didn't see anything but yes we should be aware of that

PR #800, opened 2023-03-30
Add inspection for @DslMarker

DRSchlaubi:
- Add inspection for annotating builders with @KordDsl
- Recursivly resolve super types - ignore internal builders

### DIFF
--- a/ksp-processors/src/main/kotlin/KSPUtils.kt
+++ b/ksp-processors/src/main/kotlin/KSPUtils.kt
@@ -31,3 +31,17 @@ internal val KSReferenceElement.isClassifierReference: Boolean
         is KSParenthesizedReference -> element.isClassifierReference
         else -> error("Unexpected KSReferenceElement: $this")
     }
+
+internal val KSClassDeclaration.allSupertypes: Sequence<KSClassDeclaration>
+    get() = superTypes.flatMap {
+        val declaration = it.classDeclaration
+        declaration.allSupertypes + declaration
+    }
+
+@Suppress("RecursivePropertyAccessor")
+private val KSTypeReference.classDeclaration: KSClassDeclaration
+    get() = when(val declaration = resolve().declaration) {
+        is KSClassDeclaration -> declaration
+        is KSTypeAlias -> declaration.type.classDeclaration
+        else -> error("Unsupported type: $declaration")
+    }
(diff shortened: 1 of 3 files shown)


ksp-processors/src/main/kotlin/KSPUtils.kt
```diff
@@ -27,3 +31,17 @@ internal val KSReferenceElement.isClassifierReference: Boolean
         is KSParenthesizedReference -> element.isClassifierReference
         else -> error("Unexpected KSReferenceElement: $this")
     }
+
+internal val KSClassDeclaration.allSupertypes: Sequence<KSClassDeclaration>
```
Lukellmann, line 35: there is already a `getAllSupertypes` function in ksp

ksp-processors/src/main/kotlin/inspection/BuilderDslMarkerInspectionProcessor.kt
```diff
@@ -0,0 +1,37 @@
...
+    override fun create(environment: SymbolProcessorEnvironment): SymbolProcessor
+        = BuilderDslMarkerInspectionProcessor(environment.logger)
+}
+
+class BuilderDslMarkerInspectionProcessor(private val logger: KSPLogger) : SymbolProcessor {
+    override fun process(resolver: Resolver): List<KSAnnotated> {
+        resolver.resolveAllNewClasses()
+            // some internal implementations are not annotated with
+            // @KordDsl on purpose
+            .filter { Modifier.INTERNAL !in it.modifiers }
```
Lukellmann, line 19: `filterNot` might be clearer

ksp-processors/src/main/kotlin/inspection/BuilderDslMarkerInspectionProcessor.kt
```diff
@@ -0,0 +1,37 @@
...
+            .filter {
+                it.allSupertypes.any { declaration ->
+                    declaration.qualifiedName?.asString() == "dev.kord.rest.builder.RequestBuilder"
+                }
+            }
+            .filter {
+                it.annotations.none { annotation ->
```
Lukellmann, line 26: we could already filter out the classes which have the annotation before checking the supertypes, that way we wouldn't have to resolve their supertypes

ksp-processors/src/main/kotlin/inspection/BuilderDslMarkerInspectionProcessor.kt
```diff
@@ -0,0 +1,37 @@
...
+                    annotation.annotationType.resolve()
```
Lukellmann, line 27: checking the short name before resolution would save resolution cost in some cases (see `isOfType`)

Lukellmann (commented):
should we also check if all of the annotated classes have an appropiate supertype? rn it checks for missing annotations, we could also check for wrong annotations.

Lukellmann (approved):
thanks :)

PR #810, opened 2023-04-07
Add interaction channel field

NoComment1105:
Implements a partial channel object for this. As there was not one before

https://github.com/discord/discord-api-docs/commit/88a761816b26d6891f8a276683bf57a4133aa233

### DIFF
--- a/core/src/commonMain/kotlin/entity/interaction/Interaction.kt
+++ b/core/src/commonMain/kotlin/entity/interaction/Interaction.kt
@@ -8,6 +8,7 @@ import dev.kord.core.Kord
 import dev.kord.core.behavior.interaction.InteractionBehavior
 import dev.kord.core.cache.data.InteractionData
 import dev.kord.core.entity.User
+import dev.kord.core.entity.channel.PartialChannel
 import dev.kord.core.supplier.EntitySupplyStrategy
 
 /**
@@ -24,9 +25,12 @@ public sealed interface Interaction : InteractionBehavior {
 
     override val applicationId: Snowflake get() = data.applicationId
 
-    override val channelId: Snowflake get() = data.channelId
+    override val channelId: Snowflake get() = data.channelId.value!!
 
     override val token: String get() = data.token
+
+    /** A [PartialChannel] object of the channel for the interaction. */
+    public val partialChannel: PartialChannel? get() = data.channel.value?.let { PartialChannel(it, kord, supplier) }
 
     /**
      * The type of the interaction.
--- a/common/api/common.api
+++ b/common/api/common.api
@@ -4162,29 +4162,31 @@ public final class dev/kord/common/entity/DiscordIntegrationsAccount$Companion {
 
 public final class dev/kord/common/entity/DiscordInteraction {
 	public static final field Companion Ldev/kord/common/entity/DiscordInteraction$Companion;
-	public synthetic fun <init> (ILdev/kord/common/entity/Snowflake;Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/InteractionType;Ldev/kord/common/entity/InteractionCallbackData;Ldev/kord/common/entity/optional/OptionalSnowflake;Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ljava/lang/String;ILdev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Lkotlinx/serialization/internal/SerializationConstructorMarker;)V
-	public fun <init> (Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/InteractionType;Ldev/kord/common/entity/InteractionCallbackData;Ldev/kord/common/entity/optional/OptionalSnowflake;Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ljava/lang/String;ILdev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;)V
-	public synthetic fun <init> (Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/InteractionType;Ldev/kord/common/entity/InteractionCallbackData;Ldev/kord/common/entity/optional/OptionalSnowflake;Ldev/kord/common/entity/Snowflake;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ljava/lang/String;ILdev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;Ldev/kord/common/entity/optional/Optional;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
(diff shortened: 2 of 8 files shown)


Lukellmann:
i think there is no need for a new partial channel type, the exisiting `DiscordChannel` has everything optional except for `id` and `type`

core/src/commonMain/kotlin/entity/interaction/Interaction.kt
```diff
@@ -24,10 +25,13 @@ public sealed interface Interaction : InteractionBehavior {
 
     override val applicationId: Snowflake get() = data.applicationId
 
-    override val channelId: Snowflake get() = data.channelId
+    override val channelId: Snowflake get() = data.channelId.value!!
```
Lukellmann, line 28: hm, i think this should be made nullable instead. i also think it should use the id from the `channel` field (["We recommend that you begin using this channel field to identify the source of channel of the interaction [...]"](https://discord.com/developers/docs/change-log#interaction-channel-data))

core/src/commonMain/kotlin/entity/interaction/Interaction.kt
```diff
@@ -24,10 +25,13 @@ public sealed interface Interaction : InteractionBehavior {
...
 
     override val token: String get() = data.token
 
+    /** A Partial [Channel] object for the interaction. */
+    public val partialChannel: Channel? get() = data.channel.value?.let { Channel(it, kord, supplier) }
```
Lukellmann, line 33: should this maybe use `Channel.from()`? or should we only expose the channel type as a property (as it's the only field guaranteed to be present)?

Lukellmann: the behavior of the channel is already accessible through `InteractionBehavior.channel`, so that would also give a cleaner api

Lukellmann:
any updates on this @NoComment1105 ?

NoComment1105:
I saw this, I read this, I forgot about this. I'll sort it now

core/src/commonMain/kotlin/behavior/interaction/InteractionBehavior.kt
```diff
@@ -20,14 +20,14 @@ public interface InteractionBehavior : KordEntity, Strategizable {
     public val token: String
 
     /** The id of the channel the interaction was sent from. */
-    public val channelId: Snowflake
+    public val channelId: Snowflake?
 
     /** The behavior of the channel the interaction was sent from. */
-    public val channel: MessageChannelBehavior get() = MessageChannelBehavior(channelId, kord)
+    public val channel: MessageChannelBehavior? get() = channelId?.let { MessageChannelBehavior(it, kord) }
```
Lukellmann, line 26: i think these should be overridden with non-nullable in subtypes that need to have a channel id (i.e. `ComponentInteractionBehavior`, `ChatInputCommandInteraction` and `MessageCommandInteraction`)

core/src/commonMain/kotlin/entity/interaction/Interaction.kt
```diff
@@ -24,7 +24,7 @@ public sealed interface Interaction : InteractionBehavior {
 
     override val applicationId: Snowflake get() = data.applicationId
 
-    override val channelId: Snowflake get() = data.channelId
+    public override val channelId: Snowflake? get() = data.channel.value?.id
```
Lukellmann, line 27:
```suggestion
    override val channelId: Snowflake? get() = data.channel.value?.id ?: data.channelId.value
```
use the old `channel_id` field as a fallback

Lukellmann (commented):
we should also add a `Interaction.channelType` property taken from this new `channel` field

core/src/commonMain/kotlin/behavior/interaction/InteractionBehavior.kt
Lukellmann: maybe a new interface like `ChannelInteractionBehavior` would be nice for this

core/src/commonMain/kotlin/behavior/interaction/ChannelInteractionBehavior.kt
```diff
@@ -0,0 +1,23 @@
...
+    public val channel: MessageChannelBehavior? get() = channelId?.let { MessageChannelBehavior(it, kord) }
+
+    public suspend fun getChannelOrNull(): MessageChannel? = channelId?.let { supplier.getChannelOfOrNull(it) }
+
+    @Deprecated(
+        "Discord no longer provides a non-null variant channel",
+        ReplaceWith("getChannelOrNull()"),
+        DeprecationLevel.WARNING
+    )
+    public suspend fun getChannel(): MessageChannel = channelId?.let { supplier.getChannelOf(it) }!!
```
lukellmann, line 20: what i was referring to was making `channel` and `channelId` nullable in `InteractionBehavior` and overriding them non-nullable in this new interface

lukellmann (changes requested):

NoComment1105:
Can this be reviewed please @lukellmann ?

lukellmann:
ah yes, forgot about it, will try to get to it today

core/src/commonMain/kotlin/behavior/interaction/ChannelInteractionBehavior.kt
HopeBaron, line 1: Is this required?
it has the same attributes as a `MessageChannelBehavior`

core/src/commonMain/kotlin/entity/interaction/Interaction.kt
```diff
@@ -24,10 +26,17 @@ public sealed interface Interaction : InteractionBehavior {
...
     override val applicationId: Snowflake get() = data.applicationId
 
-    override val channelId: Snowflake get() = data.channelId
+    override val channelId: Snowflake? get() = data.channel.value?.id ?: data.channelId.value
+
+    override val channel: MessageChannelBehavior? get() = channelId?.let { MessageChannelBehavior(it, kord) }
 
     override val token: String get() = data.token
 
+    /**
```
HopeBaron, line 35: No need for this, they can access it using Interaction#channel#type

lukellmann:
after thinking about this for a bit, i think it might actually be better (and would result in way less intrusive changes) to just change the logic for getting the `channelId` to `data.channel.value?.id ?: data.channelId.value!!` in all cases, without adding `ChannelInteractionBehavior`

what do you think?

NoComment1105:
> after thinking about this for a bit, i think it might actually be better (and would result in way less intrusive changes) to just change the logic for getting the `channelId` to `data.channel.value?.id ?: data.channelId.value!!` (i.e. a non-null `channelId`) in all cases, without adding `ChannelInteractionBehavior`
> 
> what do you think?

That sounds better actually, I'll look into doing that ASAP. Sorry I've missed this for so long 😄

core/src/commonMain/kotlin/entity/interaction/GuildInteraction.kt
```diff
@@ -12,6 +13,10 @@ public sealed interface GuildInteraction : Interaction, GuildInteractionBehavior
 
     override val guildId: Snowflake get() = data.guildId.value!!
 
+    override val channelId: Snowflake get() = data.channel.value?.id ?: data.channelId.value!!
+
+    override val channel: GuildMessageChannelBehavior get() = GuildMessageChannelBehavior(channelId, guildId, kord)
```
lukellmann, line 18: these overrides aren't needed

core/src/commonMain/kotlin/entity/interaction/Interaction.kt
```diff
@@ -24,7 +25,9 @@ public sealed interface Interaction : InteractionBehavior {
 
     override val applicationId: Snowflake get() = data.applicationId
 
-    override val channelId: Snowflake get() = data.channelId
+    override val channelId: Snowflake get() = data.channel.value?.id ?: data.channelId.value!!
+
+    override val channel: MessageChannelBehavior get() = MessageChannelBehavior(channelId, kord)
```
lukellmann, line 30: not needed

lukellmann (commented):
except for all the redundant overrides (only `Interaction.channelId` should need to be changed), this looks good 👍

lukellmann (approved):
thanks for all the effort 😅

PR #814, opened 2023-04-15
Add support for voice messages

DRSchlaubi:
see https://github.com/discord/discord-api-docs/pull/6082

### DIFF
--- a/common/src/commonMain/kotlin/entity/DiscordMessage.kt
+++ b/common/src/commonMain/kotlin/entity/DiscordMessage.kt
@@ -91,6 +91,7 @@ import dev.kord.common.entity.optional.Optional
 import dev.kord.common.entity.optional.OptionalBoolean
 import dev.kord.common.entity.optional.OptionalInt
 import dev.kord.common.entity.optional.OptionalSnowflake
+import dev.kord.common.serialization.DurationInFloatingPointSeconds
 import dev.kord.common.serialization.LongOrStringSerializer
 import dev.kord.ksp.GenerateKordEnum
 import dev.kord.ksp.GenerateKordEnum.Entry
@@ -544,7 +545,7 @@ public data class DiscordAttachment(
 
     val ephemeral: OptionalBoolean = OptionalBoolean.Missing,
     @SerialName("duration_secs")
-    val durationSecs: Optional<Float> = Optional.Missing(),
+    val durationSecs: Optional<DurationInFloatingPointSeconds> = Optional.Missing(),
     val waveform: Optional<String> = Optional.Missing()
 )
 
--- a/common/src/commonMain/kotlin/entity/DiscordMessage.kt
+++ b/common/src/commonMain/kotlin/entity/DiscordMessage.kt
@@ -543,6 +543,7 @@ public data class DiscordAttachment(
     val width: OptionalInt? = OptionalInt.Missing,
 
     val ephemeral: OptionalBoolean = OptionalBoolean.Missing,
+    @SerialName("duration_secs")
     val durationSecs: Optional<Float> = Optional.Missing(),
     val waveform: Optional<String> = Optional.Missing()
 )
--- a/common/src/commonMain/kotlin/entity/DiscordMessage.kt
+++ b/common/src/commonMain/kotlin/entity/DiscordMessage.kt
@@ -406,7 +406,12 @@ public enum class MessageFlag(public val code: Int) {
     FailedToMentionSomeRolesInThread(1 shl 8),
 
     /** This message will not trigger push and desktop notifications. */
-    SuppressNotifications(1 shl 12)
+    SuppressNotifications(1 shl 12),
+
+    /**
+     * This message is a voice message.
+     */
+    IsVoiceMessage(1 shl 13)
 }
 
 @Serializable(with = MessageFlags.Serializer::class)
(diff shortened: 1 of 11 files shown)


common/src/commonMain/kotlin/entity/DiscordMessage.kt
```diff
@@ -534,7 +542,10 @@ public data class DiscordAttachment(
     */
     val width: OptionalInt? = OptionalInt.Missing,
 
-    val ephemeral: OptionalBoolean = OptionalBoolean.Missing
+    val ephemeral: OptionalBoolean = OptionalBoolean.Missing,
+    @SerialName("duration_secs")
+    val durationSecs: Optional<Float> = Optional.Missing(),
```
Lukellmann, line 547: i think we should introduce a new `Duration` serializer for this, also `Double` is probably better

common/src/commonMain/kotlin/entity/Permission.kt
```diff
@@ -295,6 +295,10 @@ public sealed class Permission(public val code: DiscordBitSet) {
     /** Allows for using soundboard in a voice channel. */
     public object UseSoundboard : Permission(1L shl 42)
 
+    /**
+     * Allows sending voice messages.
+     */
+    public object SendVoiceMessages : Permission(1L shl 46)
```
Lukellmann, line 301: should be included in `values`

core/src/commonMain/kotlin/entity/Attachment.kt
```diff
@@ -58,12 +63,30 @@ public data class Attachment(val data: AttachmentData, override val kord: Kord)
...
+
+    /**
+     * A sampled waveform (currently for voice messages).
+     */
+    val waveForm: ByteArray? by lazy {
+        if (data.waveform is Optional.Value<String>) {
+            data.waveform.value.decodeBase64Bytes()
+        } else {
+            null
+        }
```
Lukellmann, line 80:
```suggestion
        data.waveform.value?.let { it.decodeBase64Bytes() }
```

Lukellmann (changes requested):
api dump is missing + could you use trailing commas? that way the diff will be clearer when these places are changed again in the future.

Lukellmann:
converted it to draft for now since https://github.com/discord/discord-api-docs/pull/6082 is not merged yet and may change

DRSchlaubi:
> api dump is missing + could you use trailing commas? that way, the diff will be clearer when these places are changed again in the future.

I hate them, but the style guide enforces them, so ig we need them, However, ktlint supports them AFAIK, so once we add that we can add all of them at once

common/src/commonMain/kotlin/serialization/DurationSerializers.kt
```diff
@@ -45,6 +45,33 @@ public sealed class DurationAsLongSerializer(
...
+/** A [Duration] that is [serializable][Serializable] with [DurationInFloatingPointSecondsSerializer]. */
+public typealias DurationInFloatingPointSeconds = @Serializable(with = DurationInFloatingPointSecondsSerializer::class) Duration
+
+/** Serializer that encodes and decodes [Duration]s as a [Double] number of seconds */
+public object DurationInFloatingPointSecondsSerializer : KSerializer<Duration> {
+    override val descriptor: SerialDescriptor =
+        PrimitiveSerialDescriptor("dev.kord.common.serialization.DurationInFloatingPointSeconds", PrimitiveKind.DOUBLE)
+
+    override fun serialize(encoder: Encoder, value: Duration) {
+        when (val valueAsDouble = value.toDouble(MILLISECONDS)) {
```
Lukellmann, line 57:
```suggestion
        when (val valueAsDouble = value.toDouble(SECONDS)) {
```

common/src/commonMain/kotlin/serialization/DurationSerializers.kt
```diff
@@ -45,6 +45,33 @@ public sealed class DurationAsLongSerializer(
...
+            Double.MIN_VALUE, Double.MAX_VALUE -> throw SerializationException(
```
Lukellmann, line 58: this check is not appropriate for `Duration.toDouble()` (it doesn't coerce but round instead), only for `Duration.toLong()`

common/src/commonMain/kotlin/serialization/DurationSerializers.kt
```diff
@@ -45,6 +45,33 @@ public sealed class DurationAsLongSerializer(
...
+                }
+            )
+
+            else -> encoder.encodeDouble(valueAsDouble)
+        }
+    }
+
+    override fun deserialize(decoder: Decoder): Duration {
+        return decoder.decodeDouble().div(1000).toDuration(MILLISECONDS)
+    }
```
Lukellmann, line 72:
```suggestion
    override fun deserialize(decoder: Decoder): Duration = decoder.decodeDouble().toDuration(SECONDS)
```
no need for this division

Lukellmann (changes requested):
also tests would be nice for this serializer

common/src/commonMain/kotlin/serialization/DurationSerializers.kt
DRSchlaubi: does toDuration() already do that for us?

Lukellmann: if you specify `SECONDS` instead of `MILLISECONDS` as the unit, yes

also you would have to multiply with 1000 if you wanted to convert the seconds to milliseconds and then to `Duration`, so the original code contained a bug

Oliver-makes-code:
Bumping as I'm currently waiting on this PR for my bot

Lukellmann:
the discord pr was merged, but there was also a new commit, adding error codes and clarifying what `waveform` is

core/src/commonMain/kotlin/entity/Attachment.kt
```diff
@@ -58,12 +60,25 @@ public data class Attachment(val data: AttachmentData, override val kord: Kord)
...
+     * The duration of the audio file (currently for voice messages).
+     */
+    val duration: Duration? get() = data.durationSecs.value
+
+    /**
+     * A sampled waveform (currently for voice messages).
+     */
+    val waveForm: ByteArray? by lazy {
+        data.waveform.value?.decodeBase64Bytes()
+    }
```
Lukellmann, line 73:
```suggestion
    val waveForm: ByteArray? get() = data.waveform.value?.decodeBase64Bytes()
```
probably not worth caching, it's rather short + mutable

Lukellmann:
i can fix the serializer and add tests if you don't want to

DRSchlaubi:
Yeah sure that's probably faster

Lukellmann (approved):
i've changed `DurationInDoubleSecondsSerializer` and added some tests, can you check if it looks ok to you?

Lukellmann:
> Bumping as I'm currently waiting on this PR for my bot

should be available with `0.9.x-SNAPSHOT` and will be included in the next release

PR #815, opened 2023-04-15
Get rid of gh-pages branch

DRSchlaubi:
Since it works for the wiki, let's try this again

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -2,19 +2,32 @@ name: Build docs
 
 on:
   push:
-    branches: [ "main" ]
+    branches: [ "0.9.x" ]
+  tag:
+  workflow_dispatch:
 
 # Gives the workflow permissions to clone the repo and create a page deployment
 permissions:
   id-token: write
   pages: write
 
+# This workflow updates the versioned docs on dokka.kord.dev using the following strategy
+# 1) pull cache to docs-cache
+# 2) Call dokkaHtmlMultiModule for current version
+#   - if this is a release build move current docs into docs-cache/$version
+#   - otherwise move it into docs-cache/current
+# 3) Build GitHub Pages artifact
+#   - Move all folders not named "current" into docs/<folder>
+#   - Move current into docs/
+# 4) Trigger deploy to GitHub Actions
 jobs:
   build-job:
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
+    GITHUB_TAG_NAME: ${{ github.event.release.tag_name }}
+    GITHUB_BRANCH_NAME: ${{ github.ref }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -23,16 +36,20 @@ jobs:
         with:
           distribution: temurin
           java-version: 8
+      - uses: actions/cache@v3
+        with:
+          path: docs-cache
+          key: docs-cache
       - name: Build docs with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: dokkaHtmlMultiModule
+          arguments: :produceDocs
       - name: Setup Pages
         uses: actions/configure-pages@v2
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
         with:
-          path: build/dokka/htmlMultiModule
+          path: docs
       - name: Deploy to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@v1
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -1,25 +1,23 @@
-# This workflow will build and deploy docs to GitHub Pages whenever something is pushed to the default branch
+name: Build docs
 
-name: Docs
+on:
+  push:
+    branches: [ "main" ]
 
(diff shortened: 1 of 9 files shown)


Lukellmann (changes requested):

.github/workflows/docs-ci.yml
```diff
@@ -1,39 +1,55 @@
...
+        with:
+          path: docs-cache
+          key: docs-cache
       - name: Build docs with Gradle
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: --stacktrace --info dokkaHtmlMultiModule
-      - name: Deploy docs to GitHub Pages
-        uses: JamesIves/github-pages-deploy-action@v4
+          arguments: :produceDocs
```
Lukellmann, line 46: where does this task come from?

gradle/libs.versions.toml
```diff
@@ -98,6 +99,7 @@ pluginsForBuildSrc = [
     "kotlin-jvm-plugin",
     "kotlin-serialization-plugin",
     "dokka-plugin",
+    "dokka-versioning-plugin",
```
Lukellmann, line 102:
```suggestion
```
should not be part of this bundle, it's a library

.github/workflows/docs-ci.yml
```diff
@@ -1,39 +1,55 @@
...
+  pages: write
 
+# This workflow updates the versioned docs on dokka.kord.dev using the following strategy
+# 1) pull cache to docs-cache
+# 2) Call dokkaHtmlMultiModule for current version
+#   - if this is a release build move current docs into docs-cache/$version
+#   - otherwise move it into docs-cache/current
+# 3) Build GitHub Pages artifact
+#   - Move all folders not named "current" into docs/<folder>
+#   - Move current into docs/
```
Lukellmann, line 21: i don't get where this is done

Lukellmann (changes requested):
the versioning plugin seems to be quite complex to configure (see [here](https://github.com/Kotlin/dokka/tree/master/plugins/versioning#usage-example)), should we maybe separate this into two prs? (one getting rid of the gh-pages branch, another one configuring the versioning plugin)

.github/workflows/docs-ci.yml
```diff
@@ -1,39 +1,55 @@
...
+name: Build docs
 
-name: Docs
-
-on: push # but only to default branch, see if below
+on:
+  push:
+    branches: [ "0.9.x" ]
+  tag:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
DRSchlaubi: Yes to trigger the build for the current release

.github/workflows/docs-ci.yml
```diff
@@ -1,39 +1,55 @@
-# This workflow will build and deploy docs to GitHub Pages whenever something is pushed to the default branch
+name: Build docs
 
-name: Docs
-
-on: push # but only to default branch, see if below
+on:
+  push:
+    branches: [ "0.9.x" ]
+  tag:
```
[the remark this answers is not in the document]
DRSchlaubi: yes `tag` doesn't exist anyways

.github/workflows/docs-ci.yml
DRSchlaubi: the produceDocs tasks

.github/workflows/docs-ci.yml
DRSchlaubi: it's called `:produceDocs` the `:` means it's from the root project, so it's in the root `build.gradle.kts`

buildSrc/src/main/kotlin/Documentation.kt
```diff
@@ -34,4 +36,16 @@ fun AbstractDokkaLeafTask.applyKordDokkaOptions() {
             suppress = true
         }
     }
+    project.dependencies {
+        "dokkaHtmlPartialPlugin"(project.libs.findLibrary("dokka-versioning-plugin").get())
+    }
+    applyVersioningPlugin()
+}
+
+fun AbstractDokkaTask.applyVersioningPlugin() {
```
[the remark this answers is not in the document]
DRSchlaubi: We need to apply this configuration for the root project and the sub projects so putting it just in the root project makes no sense

.github/workflows/docs-ci.yml
Lukellmann: is it custom? cause if so, you didn't commit it

.github/workflows/docs-ci.yml
Lukellmann: did you commit it?

.github/workflows/docs-ci.yml
```diff
@@ -1,39 +1,56 @@
...
+        uses: actions/checkout@v3
       - name: Set up JDK
         uses: actions/setup-java@v3
         with:
           distribution: temurin
           java-version: 8
+      - uses: actions/cache@v3
+        with:
+          path: docs-cache
+          key: docs-cache
```
Lukellmann, line 43: how long will this cache be persisted? because if it's cleared, we no longer have the docs of older versions

DRSchlaubi: After 7 days of not being accessed, we might need to change this to something else

DRSchlaubi:
Superseded by #855

PR #817, opened 2023-04-16
Improve `GenerateKordEnum` and `AnnotationArguments`

Lukellmann:
The `GenerateKordEnum` annotation was simplified by using a single array for both normal and deprecated entries.

`AnnotationArguments` is now more type-safe and allows checking if a default is used for a parameter.


ksp-annotations/src/commonMain/kotlin/GenerateKordEnum.kt
```diff
@@ -46,26 +43,12 @@ annotation class GenerateKordEnum(
         /** Name of the entry. */
         val name: String,
         /** [Int] value of the entry. Only set this if [GenerateKordEnum.valueType] is [INT]. */
-        val intValue: Int = DEFAULT_INT_VALUE,
+        val intValue: Int = 0,
```
DRSchlaubi, line 46: Why was this changed?

DRSchlaubi: Okay I see

DRSchlaubi (approved):

PR #818, opened 2023-04-17
Fix `InstantInEpochMillisecondsSerializerTest` on JS

Lukellmann:
The tests failed because the boundaries for `Instant` on JS are narrower than on JVM.


DRSchlaubi:
This should also fix native right?

Lukellmann:
> This should also fix native right?

yes

DRSchlaubi (approved):

PR #823, opened 2023-04-27
Simplify CI

Lukellmann:
* call Git CLI from Gradle instead of depending on CI specifics
* use normal environment variables for signing
* remove JitPack support, we publish snapshots for all branches anyway

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -15,8 +15,6 @@ jobs:
     concurrency: # Allow one concurrent deployment
       group: pages
       cancel-in-progress: true
-    env:
-      GITHUB_TAG_NAME: ${{ github.event.release.tag_name }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


Lukellmann:
#822 should be based on this

NoComment1105 (approved):

.github/workflows/docs-ci.yml
```diff
@@ -17,7 +17,6 @@ jobs:
       cancel-in-progress: true
     env:
       GITHUB_TAG_NAME: ${{ github.event.release.tag_name }}
```
DRSchlaubi, line 19: Should we replace this too?

DRSchlaubi (changes requested):

.github/workflows/docs-ci.yml
Lukellmann: sure

DRSchlaubi (approved):

PR #825, opened 2023-05-05
Support new username system

lukellmann:
see https://github.com/discord/discord-api-docs/pull/6218


core/src/commonMain/kotlin/entity/User.kt
```diff
@@ -67,7 +96,8 @@ public open class User(
     /**
      * The complete user tag.
      */
-    public val tag: String get() = "$username#$discriminator"
+    @Suppress("DEPRECATION")
+    public val tag: String get() = if (migratedToNewUsernameSystem) "@$username" else "$username#$discriminator"
```
lukellmann, line 100: i'm not sure about this one, should we do it like this or just deprecate it?

lukellmann: [JDA](https://github.com/DV8FromTheWorld/JDA/pull/2462/files#diff-4c39d04c5713b45f7fed072d33c065313d40af373e8b6513516ede58a94da8b8) seems to deprecate it, [discord.js](https://github.com/discordjs/discord.js/pull/9512/files#diff-b7b221539f62dcbc8c29a6cad99d732320243e3bcc5fab3e4fec4383ac4d6bd1) seems to do it like i did here

NoComment1105: I'd go with a straight deprecation and add another Val that gets the @username because the at username has a different name now if I remember correctly. I just don't remember said name

matytyma: > Val that gets the @username because the at username has a different name now if I remember correctly. I just don't remember said name

It's still called a [username](https://support.discord.com/hc/en-us/articles/12620128861463#h_01GXPQAGG6W477HSC5SR053QG1) :smile:

HopeBaron: Hello
I'd go with deprecation
It's possible that Discord will give us a window to migrate given that the username change is rolling out incrementally.

core/src/commonMain/kotlin/entity/User.kt
```diff
@@ -48,8 +54,31 @@ public open class User(
...
+    // "0" when data.discriminator is missing: if the field is missing, all users were migrated,
+    // see https://discord.com/developers/docs/change-log#identifying-migrated-users:
+    // "After migration of all users is complete, the discriminator field may be removed."
+    @Suppress("DEPRECATION", "DeprecatedCallableAddReplaceWith")
+    @Deprecated(
+        "Discord's username system is changing and discriminators are being removed, see " +
+            "https://discord.com/developers/docs/change-log#unique-usernames-on-discord for details.",
+        level = WARNING,
+    )
+    public val discriminator: String get() = data.discriminator.value ?: "0"
```
lukellmann, line 71: should we even deprecate this? bots will continue to have discriminators...

core/src/commonMain/kotlin/entity/Member.kt
```diff
@@ -30,9 +30,9 @@ public class Member(
         get() = memberData.guildId
 
     /**
-     * The name as shown in the discord client, prioritizing the [nickname] over the [username].
+     * The name as shown in the discord client, prioritizing [nickname] over [globalName] and [username].
      */
-    public val displayName: String get() = nickname ?: username
+    override val displayName: String get() = nickname ?: super.displayName
```
lukellmann, line 35: should this really be overridden? or should `User` have an extension property for this?

i.e. what would you expect in the following situation:
```kt
fun doSomethingWithUser(user: User) {
    println(user.displayName)
}

val member: Member = TODO()
// should this print the member nickname or the user globalName?
doSomethingWithUser(member)
```

core/src/commonMain/kotlin/entity/User.kt
lukellmann: update: discord.js also seems to deprecate it: https://github.com/discordjs/discord.js/pull/9512/commits/22b743f861855d64a0672d038de6c9fc98fea965

core/src/commonMain/kotlin/entity/Member.kt
HopeBaron: This is not alright as the user will lose their ability to pick the original display name.
make a effectiveDisplayName property on the member 
otherwise you can always pick the displayName property

core/src/commonMain/kotlin/entity/User.kt
HopeBaron: It's better to push toward the new system so it's good to have that. [DeprecatedSinceKord](https://github.com/kordlib/kord/blob/0.9.x/common/src/commonMain/kotlin/annotation/Annotations.kt#LL82C25-L82C44) could be used.

HopeBaron (approved):

core/src/commonMain/kotlin/entity/Member.kt
lukellmann: so you suggest changing the behavior of the already existing `Member.displayName` like this?
```kt
// new property in the User class:
val displayName = globalName ?: username

// previous behavior in the Member class:
val displayName = nickname ?: username
// new behavior in the Member class:
val displayName = globalName ?: username // same as for User

// new property in the Member class:
val effectiveDisplayName = nickname ?: globalName ?: username
```

HopeBaron: @lukellmann Yes, that's the point
Since the global name is prioritized over the username I think it would serve as a better approach in the new username system.

lukellmann: i actually came up with a way that wouldn't change the behavior for members but instead deprecate it

lukellmann:
if [this](https://github.com/kordlib/kord/pull/825#discussion_r1224663102) is ok, this should be good to go

core/src/commonMain/kotlin/entity/Member.kt
```diff
@@ -42,7 +42,7 @@ public class Member(
     /**
      * The member's effective name, prioritizing [nickname] over [globalName] and [username].
      */
-    public val effectiveName: String get() = nickname ?: (this as User).displayName
+    public val effectiveName: String get() = nickname ?: (this as User).effectiveName
```
HopeBaron, line 47: Can't we replace that with a super call?

lukellmann: no, super only works for members, it isn't a normal expression

DRSchlaubi (approved):

HopeBaron (approved):

PR #831, opened 2023-06-11
Use main as the permanent default branch

lukellmann:


lukellmann:
we can also change the name of this branch if you think that'd fit better, other candidates would e.g. be `dev` or `develop`

HopeBaron:
So remove the version branches?

PR #832, opened 2023-06-11
Add support for join raid protection

lukellmann:
see https://github.com/discord/discord-api-docs/pull/5778 and https://github.com/discord/discord-api-docs/pull/6133


core/src/commonMain/kotlin/cache/data/GuildData.kt
```diff
@@ -65,7 +65,8 @@ public data class GuildData(
     val stageInstances: Optional<List<StageInstanceData>> = Optional.Missing(),
     val stickers: Optional<List<StickerData>> = Optional.Missing(),
     val guildScheduledEvents: Optional<List<GuildScheduledEventData>> = Optional.Missing(),
-    val premiumProgressBarEnabled: Boolean
+    val premiumProgressBarEnabled: Boolean,
+    val safetyAlertsChannelId: Snowflake?,
```
HopeBaron, line 69: It's weird this is required to be filled, according to how the docs describe it this should be optional

rest/src/commonMain/kotlin/json/request/GuildRequests.kt
```diff
@@ -244,6 +244,8 @@ public data class GuildModifyRequest(
     @SerialName("preferred_locale")
     val preferredLocale: Optional<String?> = Optional.Missing(),
     val features: Optional<Set<GuildFeature>> = Optional.Missing(),
+    @SerialName("safety_alerts_channel_id")
+    val safetyAlertsChannelId: OptionalSnowflake? = OptionalSnowflake.Missing,
```
HopeBaron, line 248: You set it as optional here, but as a required nullable in the guild data object?

Please refer to my comment above

core/src/commonMain/kotlin/cache/data/GuildData.kt
lukellmann: in the guild object it's described to only be nullable: https://discord.com/developers/docs/resources/guild#guild-object-guild-structure (`?` before the type means nullable)

rest/src/commonMain/kotlin/json/request/GuildRequests.kt
lukellmann: yeah, this is the modify request:
missing means leave it as is, null means reset it

https://discord.com/developers/docs/resources/guild#modify-guild-json-params says it's optional and nullable (all parameters are optional as described in the info box above)

lukellmann:
could you take another look @HopeBaron?

DRSchlaubi (approved):

PR #841, opened 2023-07-03
Prepare flag generation

lukellmann:
This is a preparation for #766. Already applying these changes to the generated kord enums will make the diff of #766 less noisy.


lukellmann:
included in #842

PR #845, opened 2023-07-10
Publish KSP processors

DRSchlaubi:
I want to use this in my OAuth lib, so would be useful to have it

### DIFF
--- a/ksp-processors/build.gradle.kts
+++ b/ksp-processors/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     `kord-internal-module`
+    org.jetbrains.dokka
     `kord-publishing`
 }
 


lukellmann:
also wouldn't a pure snapshot branch with this change be enough?

DRSchlaubi:
Ig, but if we update the processor, we also need to update this

lukellmann:
true, it's ok for me then.

however are you sure that `ksp-processors` is really published (check with the snapshot of the branch), `kord-module.kts` also registers a publication, while this `build.gradle.kts` does not.
a comment about why this is published would also be good.

and remember, we might change from using annotations/ksp for generating to a custom dsl and gradle task / simple application with `main()` instead. reason being that there is almost no benifit but rather limitations from using annotations for the data about what should be generated (we discussed this before).

lukellmann:
i thought about this again, can't we simply add the types you need to kord?

DRSchlaubi:
No, it's also used in lavalink.kt

lukellmann:
so for non-discord related generation?

PR #850, opened 2023-07-26
Add icon to IntelliJ

DRSchlaubi:
@HopeBaron would be great if you could replace the svg with the actual original


lukellmann:
> @HopeBaron would be great if you could replace the svg with the actual original

where did you get this icon from then?

.gitignore
```diff
@@ -1,9 +1,10 @@
...
 .gradle/
-.idea/
+.idea/*
 out/
 
 **/build/*
 !**/build/generated/
 # re-exclude BuildConfigGenerated.kt
 common/build/generated/sources/buildConfig/main/dev/kord/common/BuildConfigGenerated.kt
+!.idea/icon.svg
```
lukellmann, line 10: i think the include should be put right under the `.idea/*` exclude

DRSchlaubi:
> > @HopeBaron would be great if you could replace the svg with the actual original
> 
> where did you get this icon from then?

Used a png to svg converter

lukellmann:
ok, that explains this:
kord

DRSchlaubi:
> ok, that explains this: kord

yeah exactly

lukellmann:
i now simply replaced the svg with a png, is that also ok @DRSchlaubi?

DRSchlaubi:
If IntelliJ can read it yeah ig

lukellmann:
[kotlinx.coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/master/.idea/icon.png) and [Kotlin](https://github.com/JetBrains/kotlin/blob/master/.idea/icon.png) also use png, so it seems to work

DRSchlaubi:
alright

PR #852, opened 2023-08-02
Add `avatar_decoration` to user

NoComment1105:
https://github.com/discord/discord-api-docs/pull/5723


lukellmann:
this should also have an `Asset` in core like [`User.avatar`](https://github.com/kordlib/kord/blob/main/core/src/commonMain/kotlin/entity/User.kt#L53)

lukellmann (changes requested):

core/src/commonMain/kotlin/cache/data/UserData.kt
```diff
@@ -28,7 +28,8 @@ public data class UserData(
     val bot: OptionalBoolean = OptionalBoolean.Missing,
     val publicFlags: Optional<UserFlags> = Optional.Missing(),
     val banner: String? = null,
-    val accentColor: Int? = null
+    val accentColor: Int? = null,
+    val avatarDecoration: String? = null
```
lukellmann, line 32: this isn't passed in from `DiscordUser`

lukellmann (changes requested):

core/src/commonMain/kotlin/cache/data/UserData.kt
NoComment1105: Oops, I'll fix that when I get home. Is that removing this or making it an optional

lukellmann: you need to pass `avatarDecoration` to this constructor in the `from` function inside the companion object. to avoid conversion, it should therefore be also an optional.

lukellmann (approved):
thanks :)

PR #853, opened 2023-08-03
Add `MediaChannel`s

lukellmann:
Common functionality between `ForumChannel` and `MediaChannel` was extracted to a new `ThreadOnlyChannel` supertype.

see https://github.com/discord/discord-api-docs/pull/6232

### DIFF
--- a/common/api/common.api
+++ b/common/api/common.api
@@ -1711,6 +1711,10 @@ public final class dev/kord/common/entity/ChannelFlag$Companion {
 	public fun values ()[Ldev/kord/common/entity/ChannelFlag;
 }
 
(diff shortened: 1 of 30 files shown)


lukellmann:
i think one thing that could still be improved is a common supertype for the create / edit builders, they are mostly copy-pasted from the forum ones

DRSchlaubi:
Yeey a new channel interface

lukellmann:
> i think one thing that could still be improved is to have a common supertype for the create / edit builders, they are mostly copy-pasted from the forum ones

i've decided that this can be done in a separate pr, so this one doesn't take longer than necessary

PR #854, opened 2023-08-05
Improve publishing

DRSchlaubi:
This PR aims to prepare publishing for K/N and other advantages:
- We can now automatically publish to maven central even in a KMP setup where artifacts come from different CI runs
- We have the same ci configuration for all modules (since the plugin supports all sources)
- We no longer have to register artifacts on our own


DRSchlaubi:
I think the reson it failed is because the key is wrong, see https://vanniktech.github.io/gradle-maven-publish-plugin/central/#in-memory-gpg-key

lukellmann:
the secret holds the plain ascii-armored key but base64 encoded

DRSchlaubi:
> the secret holds the plain ascii-armored key but base64 encoded

Yeah I think that's different to what is expected here, it wants the base64 of the key, as in the parts between the ----BEGIN--- ---END--- parts

DRSchlaubi:
Superseded by #855

PR #860, opened 2023-08-27
Add webhook client

DRSchlaubi:


Galarzaa90:
I was just looking for this! I see this will be almost a year. Is there something missing or something to help with?

DRSchlaubi:
> I was just looking for this! I see this will be almost a year. Is there something missing or something to help with?
> 
> I'm really interested in a webhook client.

Time flies by whoohoo

Yeah the review from @lukellmann is missing so let's nudge him again

lukellmann:
tbh i'm not sure about adding a way to get a half baked `Kord` instance - but doing it in some other way would require a LOT of work (extra entities for anything that has a reference to a webhook client)

DRSchlaubi:
This is simmilar to Kord.restOnly

Galarzaa90:
I'm already using a half baked `Kord` instance to manage webhooks:

```kt
Kord.restOnly("") {
    applicationId = Snowflake(-1)
}
```

At the moment, it is only using the execute:

```kt
kord.rest.webhook.executeWebhook(..., ...) {
}
```

So no problem there. But also the fact that I have everything in the API available to call even though it will fail, it is something you have to be careful with. 🤷🏼‍♂️

lukellmann:
> This is simmilar to Kord.restOnly

hm, that's right - maybe we could just add another factory function and don't even have the `WebhookClient` interface. we could eventually design some proper kinds of different clients with their own entities and stuff (autogeneration? 👀)

PR #861, opened 2023-08-27
Bring kord enum API closer to bit flags API

lukellmann:
This was done by only printing the raw value for `Unknown` entries in `toString()` and replacing the public `Unknown` constructor with a `from()` function in the companion object.


lukellmann:
The function is named `fromShift` in the generated bit flag classes, but i think `fromValue` / `fromCode` etc. is a bit redundant, the parameter is already named like this.

DRSchlaubi (approved):

PR #855, opened 2023-08-05
Add support for Kotlin/Native

DRSchlaubi:
This PR aims at implementing Kotlin/Native for the following platforms:
- linuxX64
- mingwX64
- macosArm64
- macosX64
- iosArm64
- iosX64
- iosSimulatorArm64
- watchosX64
- watchosArm64
- watchosSimulatorArm64
- tvosX64
- tvosArm64
- tvosSimulatorArm64

# Experimental Snapshot

There currently is an experimental snapshot on [OSSRH](https://oss.sonatype.org/content/repositories/snapshots/) under the `feature-native-SNAPSHOT` for you to try

# Current limitations

Also see #69 
- WebSockets related APIs (gateway, voice, core (except for rest only) don't work on Linux
- The required CI infrastructure is setup, but not fully configured yet, so the experimental snapshot is pushed by hand

# Remaining tasks
- [ ] Port voice to native (cc @lost-illusi0n)
- [ ] Configure CI 

# Blocked by
- #822 
- https://github.com/kordlib/cache/pull/12

PR #863, opened 2023-08-31
Don't use value class `Reset` with `AtomicRef`

DRSchlaubi:
The use of value classes for the `Reset` class has already caused issues implementing #855. However, it was isolated to native platforms. Unfortunately, this behavior will become a compiler error soon, so we need to change it. This can be achieved by simply using the backing `Instant` instead of `Reset` with `AtomicRef`.

# Related Issues
#855 
#69 
Kotlin/kotlinx-atomicfu#291
[KT-61584](https://youtrack.jetbrains.com/issue/KT-61584)


lukellmann:
if we have a breaking change here anyway, do we really need this wrapper around `Instant` at all?

DRSchlaubi:
Probably not

lukellmann:
let me see what i can do then

lukellmann:
ok, so i kept `Reset` as an `@JvmInline value class`, deprecated it and replaced the usages with plain `Instant`. with some annotation trickery, i think i managed to preserve binary compatibility, so only source compatibility is broken now (which is easier to fix).

lukellmann:
can you take a look @DRSchlaubi?

lukellmann:
wait a second, i just noticed that we don't even have to change any public api, we can just box/unbox the `Reset.value` internally without api changes

lukellmann (approved):
simply using the backing `Instant` should do the job, what do you think @DRSchlaubi?

DRSchlaubi:
Looks good

PR #864, opened 2023-09-02
Fix usage of `BigInteger` in `DiscordBitSet.value`

lukellmann:
`DiscordBitSet` can only represent positive numbers, so the `BigInteger` used for getting a decimal representation can be constructed using the sign-magnitude representation instead of the two's-complement binary representation.

This fixes two bugs:

 * `DiscordBitSet`s with an empty `data` array couldn't be converted to a `BigInteger` because the constructor taking the two's-complement binary representation throws a `NumberFormatException` if the given array is empty.
 This bug was reported by @Tmpod: https://discord.com/channels/556525343595298817/1147254164469063773

 * `DiscordBitSets` with a negative `Long` at the last position in their `data` array would be misinterpreted as negative numbers in the conversion to `BigInteger`s because the two's-complement binary representation was used.
 Example where this bug could be observed:
  ```kotlin
  // printed -9223372036854775808, should be 9223372036854775808
  println(DiscordBitSet(1L shl 63).value)
  ```

### DIFF
--- a/common/src/commonTest/kotlin/BitSetTests.kt
+++ b/common/src/commonTest/kotlin/BitSetTests.kt
@@ -1,6 +1,8 @@
 package dev.kord.common
 
 import kotlin.js.JsName
+import kotlin.random.Random
+import kotlin.random.nextLong
 import kotlin.test.*
 
 class BitSetTests {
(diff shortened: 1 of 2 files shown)


Tmpod:
Tested this on my affected project and it's all good now, thank you!

PR #884, opened 2023-10-13
Add interaction permissions to MemberData and Member

### LINKED ISSUE #664: Missing interactions `permissions` field on Kord Core's `Member` object
When you receive a interaction, the member object has a `permissions` field, which allows you to check what permissions the member has, however this is "thrown away" when creating the `MemberData`.

It would be nice if there was a `Member` entity that does expose the `permissions` field, so you can do permissions checks without needing to do REST calls.

While @Lukellmann pointed out that there is `GuildInteraction.permissions`, having the `Member` object have the `permissions` field is also useful for slash command options, because Members retrieved from the `resolved` objects field in slash commands also have the `permissions` field.

Example: Let's suppose you have a `/user info @LustigerLurch` command, and you want to show the permissions' of the provided member in the command, the permissions of the user are provided in the partial member within the `resolved.members` map. This avoids a REST request to get the roles' permissions.

image

Tmpod:
Fixes #664

I'm unsure if the `Optional` vs `null` usage is in line with the project's preferences, please correct me if it isn't!

### DIFF
--- a/core/src/commonMain/kotlin/entity/Member.kt
+++ b/core/src/commonMain/kotlin/entity/Member.kt
@@ -11,6 +11,7 @@ import dev.kord.core.behavior.RoleBehavior
 import dev.kord.core.behavior.UserBehavior
 import dev.kord.core.cache.data.MemberData
 import dev.kord.core.cache.data.UserData
+import dev.kord.core.entity.interaction.GuildInteraction
 import dev.kord.core.supplier.EntitySupplier
 import dev.kord.core.supplier.EntitySupplyStrategy
 import kotlinx.coroutines.flow.*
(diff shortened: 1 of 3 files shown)


core/src/commonMain/kotlin/entity/Member.kt
```diff
@@ -93,6 +93,12 @@ public class Member(
     /** The [GuildMemberFlags] of this member. */
     public val flags: GuildMemberFlags get() = memberData.flags
 
+    /**
+     * The [Permissions] of this member.
+     * Non-null in interaction contexts.
+     */
```
lukellmann, line 99:
```suggestion
    /**
     * The total [Permissions] of this member in the channel an interaction was sent from.
     *
     * This is only non-null when obtained from a [GuildInteraction][GuildInteraction.user].
     */
```
slightly adjusted documentation, see [here](https://discord.com/developers/docs/resources/guild#guild-member-object-guild-member-structure)

this would also need an import for `dev.kord.core.entity.interaction.GuildInteraction` to resolve the reference correctly

lukellmann (commented):
looks good, just a small documentation nit

i think using `Optional` is the right choice here

core/src/commonMain/kotlin/entity/Member.kt
Tmpod: awesome, thank you! will make the changes now

PR #886, opened 2023-10-24
Migrate Snowflake and flags to inline value classes

zTrap:
Related to https://github.com/kordlib/kord/issues/711


common/src/commonMain/kotlin/entity/Snowflake.kt
```diff
@@ -16,23 +17,15 @@ import kotlin.time.TimeMark
...
  */
+@JvmInline
 @Serializable(with = Snowflake.Serializer::class)
-public class Snowflake : Comparable<Snowflake> {
+public value class Snowflake(public val value: ULong) : Comparable<Snowflake> {
 
-    /**
-     * The raw value of this Snowflake as specified by the
-     * [Discord Developer Documentation](https://discord.com/developers/docs/reference#snowflakes).
-     */
```
lukellmann, line 26: i think this documentation should be kept

common/src/commonMain/kotlin/entity/optional/OptionalSnowflake.kt
```diff
@@ -96,7 +94,7 @@ public sealed class OptionalSnowflake {
         override val descriptor: SerialDescriptor = ULong.serializer().descriptor
 
         override fun deserialize(decoder: Decoder): OptionalSnowflake =
-            Value(decoder.decodeInline(descriptor).decodeLong().toULong())
+            Value(Snowflake(decoder.decodeInline(descriptor).decodeLong().toULong()))
```
lukellmann, line 97: can be delegated to `Snowflake`'s serializer instead

lukellmann (commented):
can you run `./gradlew apiDump` and commit the result? i want to know how bad it's going to be...

lukellmann:
Is there an issue already for the problems with mockk?

zTrap:
I've got wrong. This problem isn't related to inline classes, but to unsigned numbers. https://github.com/mockk/mockk/issues/544

lukellmann:
> I've got wrong. This problem isn't related to inline classes, but to unsigned numbers. [mockk/mockk#544](https://github.com/mockk/mockk/issues/544)

the issue you pointed to is now closed, is there any update on this?

zTrap:
Yes, issue is closed, but it not released yet

zTrap:
@lukellmann seems like all works fine with `mockk` v1.13.11

lukellmann:
I'd like to discuss this some more before deciding what to do with this, see https://discord.com/channels/556525343595298817/1245773750486438060

lukellmann:
i've updated the target branch btw

lukellmann (changes requested):

lukellmann (approved):

lukellmann:
these changes can be tested using `experiments-inline-value-classes-SNAPSHOT`

PR #887, opened 2023-10-26
feat: snowflake extension

TozyDev:
It would be nice to have an extension to create Snowflake instances.

Without extension:
```kotlin
kord.createGuildApplicationCommands(Snowflake(123456789)) {
        TODO()
}
```

With extension:
```kotlin
kord.createGuildApplicationCommands(123456789.snowflake()) {
        TODO()
}
```


lukellmann:
Hm, is there really that much value in having an extension function for this?

TozyDev:
> Hm, is there really that much value in having an extension function for this?

It will be very convenient in some cases. For example, to get raw data from config or database, just call the extension instead of calling the constructor, the code will be more readable or something like that.

lukellmann:
if it's about the postfix notation, you could also do `someExpression.let(::Snowflake)`

PR #890, opened 2023-11-09
Use debug level for unknown gateway event name 

viztea:
Similar to the voice package, use debug logs on unknown event. 

Using `warn` isn't beneficial to the developer and may end up filling logs with something useless.


DRSchlaubi (approved):
Sounds reasonable enough for me

lukellmann (approved):

PR #897, opened 2023-12-12
Implement AES Encryption for Voice

viztea:
AES for voice receive is currently unsupported. 

I tested voice send/receive for all 3 xsalsa20 poly1305 nonce strategies, all work as expected.

### DIFF
--- a/voice/src/main/kotlin/VoiceConnectionBuilder.kt
+++ b/voice/src/main/kotlin/VoiceConnectionBuilder.kt
@@ -7,8 +7,8 @@ import dev.kord.gateway.Gateway
 import dev.kord.gateway.UpdateVoiceStatus
 import dev.kord.gateway.VoiceServerUpdate
 import dev.kord.gateway.VoiceStateUpdate
+import dev.kord.voice.encryption.VoiceEncryption
 import dev.kord.voice.encryption.strategies.LiteNonceStrategy
-import dev.kord.voice.encryption.strategies.NonceStrategy
 import dev.kord.voice.exception.VoiceConnectionInitializationException
 import dev.kord.voice.gateway.DefaultVoiceGatewayBuilder
 import dev.kord.voice.gateway.VoiceGateway
@@ -31,7 +31,7 @@ public class VoiceConnectionBuilder(
     public var gateway: Gateway,
     public var selfId: Snowflake,
     public var channelId: Snowflake,
-    public var guildId: Snowflake
+    public var guildId: Snowflake,
 ) {
     /**
      * The amount in milliseconds to wait for the events required to create a [VoiceConnection]. Default is 5000, or 5 seconds.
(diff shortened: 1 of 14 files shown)


DRSchlaubi (changes requested):
please run `./gradlew apiDump`

lukellmann:
i'm not very familiar with voice connections, can you explain a bit why this is needed?

viztea:
> i'm not very familiar with voice connections, can you explain a bit why this is needed?

It kinda is and kinda isn't. I initially wanted to implement AES because it was faster than xsalsa but after doing a couple benchmarks it looks like xsalsa is a little bit more consistent and a couple microseconds faster.

This isn't really that surprising considering Kord Voice isn't the most efficient voice library, it doesn't help that Java relies on `ByteBuffer`s to be performant.

voice/src/main/kotlin/VoiceConnectionBuilder.kt
```diff
@@ -166,19 +167,23 @@ public class VoiceConnectionBuilder(
             .build()
         val udpSocket = udpSocket ?: GlobalVoiceUdpSocket
         val audioProvider = audioProvider ?: EmptyAudioPlayerProvider
-        val nonceStrategy = nonceStrategy ?: LiteNonceStrategy()
+        val encryption = if ((receiveVoice || streams != null) && encryption?.supportsDecryption != true) {
+            VoiceEncryption.XSalsaPoly1305()
+        } else {
+            encryption ?: VoiceEncryption.AeadAes256Gcm
+        }
```
lost-illusi0n, line 174: It might be useful to log if the user's chosen encryption method ends up not being used.

lost-illusi0n:
Functionality-wise LGTM. It likely will make adding/working with encryption a bit easier in the future too, if (or rather when) Discord decides to change things up.

voice/src/main/kotlin/VoiceConnectionBuilder.kt
lukellmann: or even throw an exception? but logging a warning is probably best.

voice/src/main/kotlin/encryption/VoiceEncryption.kt
```diff
@@ -0,0 +1,163 @@
...
+import javax.crypto.Cipher
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.SecretKeySpec
+
+public sealed interface VoiceEncryption {
+    public val mode: EncryptionMode
+
+    public val nonceLength: Int
+
+    public val supportsDecryption: Boolean get() = true
```
lukellmann, line 21: not having a default implementation here will avoid mistakes if another `VoiceEncryption` is added at some point

voice/src/main/kotlin/encryption/strategies/NonceStrategy.kt
```diff
@@ -27,4 +23,17 @@ public sealed interface NonceStrategy {
      * Writes the [nonce] to [cursor] in the correct relative position.
      */
     public fun append(nonce: ByteArrayView, cursor: MutableByteArrayCursor)
+
+    public interface Factory {
```
lukellmann, line 27: why do we need this factory? i'd rather not add it if not needed.

voice/src/main/kotlin/encryption/VoiceEncryption.kt
```diff
@@ -0,0 +1,163 @@
...
+            get() = EncryptionMode.AeadAes256Gcm
+
+        override val nonceLength: Int
+            get() = 4
+
+        override val supportsDecryption: Boolean
+            get() = false
+
+        override fun createBox(key: ByteArray): Box = object : Box {
+            //
```
lukellmann, line 86: what are these empty comments for?

viztea:
- Implemented voice receive for AES, had to refactor a bit of the code to support passing of AEAD data.
- Removed `VoiceEncryption#supportsDecryption`
- Removed the sealed modifier for `VoiceEncryption` to allow external implementations (??)
- Added kdoc to the `VoiceEncryption` interface.

Not sure if removing supportsDecryption is a good idea with it being an open interface 🤷🏼

PR #898, opened 2023-12-12
Implement Media Sink Wants in Voice

viztea:
Adds a new voice gateway command (op code `15`) that controls whether the voice server sends voice packets to the bot. 

Still a bit unsure on the API since Kord voice allows setting the deafen value when joining and the streams implementation to use.

### DIFF
--- a/voice/src/main/kotlin/gateway/Command.kt
+++ b/voice/src/main/kotlin/gateway/Command.kt
@@ -31,6 +31,10 @@ public sealed class Command {
                     composite.encodeSerializableElement(descriptor, 0, OpCode.Serializer, OpCode.Heartbeat)
                     composite.encodeLongElement(descriptor, 1, value.nonce)
                 }
+                is MediaSinkWants -> {
+                    composite.encodeSerializableElement(descriptor, 0, OpCode.Serializer, OpCode.MediaSinkWants)
+                    composite.encodeSerializableElement(descriptor, 1, MediaSinkWants.serializer(), value)
+                }
                 is SendSpeaking -> {
                     composite.encodeSerializableElement(descriptor, 0, OpCode.Serializer, OpCode.Speaking)
                     composite.encodeSerializableElement(descriptor, 1, SendSpeaking.serializer(), value)
@@ -70,6 +74,12 @@ public data class SendSpeaking(
     val speaking: SpeakingFlags,
     val delay: Int,
     val ssrc: UInt
+) : Command()
+
+@KordVoice
+@Serializable
+public data class MediaSinkWants(
+    val any: Int
 ) : Command()
 
 @KordVoice
--- a/voice/src/main/kotlin/gateway/DefaultVoiceGateway.kt
+++ b/voice/src/main/kotlin/gateway/DefaultVoiceGateway.kt
@@ -37,6 +37,7 @@ public data class DefaultVoiceGatewayData(
     val sessionId: String,
     val client: HttpClient,
     val reconnectRetry: Retry,
+    val isDeaf: Boolean,
     val eventFlow: MutableSharedFlow<VoiceEvent>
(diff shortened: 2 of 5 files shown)


DRSchlaubi (changes requested):
please run `./gradlew apiDump`

lukellmann:
i'm curious, are there any resources to read up on this feature? it isn't documented in the discord developer portal yet

viztea:
> i'm curious, are there any resources to read up on this feature? it isn't documented in the discord developer portal yet

It isn't documented, just like voice receive (which is supported by kord voice), this was discovered through reverse engineering the official Discord Client.

voice/src/main/kotlin/gateway/DefaultVoiceGateway.kt
```diff
@@ -196,7 +197,7 @@ public class DefaultVoiceGateway(
                     val copy = command.copy(data = command.data.copy(address = "ip"))
                     "Voice Gateway >>> ${Json.encodeToString(Command.SerializationStrategy, copy)}"
                 }
-                is Heartbeat, is Resume, is SendSpeaking -> "Voice Gateway >>> $json"
+                else -> "Voice Gateway >>> $json"
```
lukellmann, line 200: Please don't use `else` here to prevent a future command with sensitive data being logged.

voice/src/main/kotlin/gateway/Command.kt
```diff
@@ -72,6 +76,12 @@ public data class SendSpeaking(
     val ssrc: UInt
 ) : Command()
 
+@KordVoice
+@Serializable
+public data class MediaSinkWants(
+    val any: Int
```
lukellmann, line 82: can you add a comment to describe what this does? `any` isn't a very descriptive name...

voice/src/main/kotlin/gateway/DefaultVoiceGatewayBuilder.kt
```diff
@@ -21,6 +21,7 @@ public class DefaultVoiceGatewayBuilder(
     public var client: HttpClient? = null
     public var reconnectRetry: Retry? = null
     public var eventFlow: MutableSharedFlow<VoiceEvent> = MutableSharedFlow(extraBufferCapacity = Int.MAX_VALUE)
+    public var isDeaf: Boolean = false
```
lukellmann, line 24: `VoiceConnectionBuilder` already has the `selfDeaf` option, can't this be used instead?

lukellmann (changes requested):

voice/src/main/kotlin/gateway/DefaultVoiceGatewayBuilder.kt
viztea: I suppose. I think the Voice Connection Builder could get cleaned up quite a bit though. 

Like I state in the PR description, you can set `selfDeaf` and a `streams` implementation.

Since `selfDeaf` is just a UI update right now it's not that much of an issue but if it gets updated to actually prevent voice receive it might be a good idea to set it depending on the streams implementation.

And maybe add some more functionality to it as well:

```kt
interface Streams {
	suspend fun setDeafen(value: Boolean)
	
	suspend fun setMute(ssrc: UInt, value: Boolean)
}
```

lost-illusi0n: I think this is a fair solution. Makes the deafening much behavior much more consistent, especially with this addition. Streams#setDeafen might not be needed, as it isn't provided for the UI deafen either. I could see the use for a VoiceConnection#(set)Deafen which would UI deafen and media sink deafen (if streams are open).

lost-illusi0n:
Functionality-wise LGTM.

PR #914, opened 2024-01-31
`DiscordVoiceState#member` uses incorrect serial name

viztea:


lukellmann (approved):
thanks for catching this :)

PR #923, opened 2024-03-05
Rewrite `Event.DeserializationStrategy`

lukellmann:
When deserializing `Event`s, `Event.DeserializationStrategy` assumed that the [`d` field](https://discord.com/developers/docs/topics/gateway-events#payload-structure) was the last field to be observed. If it wasn't the last field, the deserialization could fail in two ways:
* the `t` and `s` fields were ignored if they came after the `d` field
* an exception was thrown if the `op` field came after the `d` field

To fix these possible failure cases, the deserialization logic has been changed to work in two steps:
1. decode all fields regardless of order, treating the `d` field as a plain `JsonElement`
2. after all fields have been decoded, construct an `Event` from the `JsonElement`, depending on the values of the `op`, `t` and `s` fields

The supertype of `Event.DeserializationStrategy` has also been changed from `DeserializationStrategy<Event?>` to `DeserializationStrategy<Event>` - `deserialize` no longer returns `null` in some cases of illegal event payloads, but throws exceptions instead.

Fixes #922

PR #925, opened 2024-03-16
Rewrite `VoiceEvent.DeserializationStrategy`

lukellmann:
Similar to #923 but for the voice gateway.

### DIFF
--- a/voice/src/main/kotlin/gateway/VoiceEvent.kt
+++ b/voice/src/main/kotlin/gateway/VoiceEvent.kt
@@ -3,70 +3,67 @@ package dev.kord.voice.gateway
 import dev.kord.common.entity.Snowflake
 import dev.kord.voice.EncryptionMode
 import dev.kord.voice.SpeakingFlags
-import io.github.oshai.kotlinlogging.KotlinLogging
-import kotlinx.serialization.ExperimentalSerializationApi
-import kotlinx.serialization.SerialName
-import kotlinx.serialization.Serializable
-import kotlinx.serialization.builtins.nullable
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
 import kotlinx.serialization.descriptors.buildClassSerialDescriptor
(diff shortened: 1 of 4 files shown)


voice/src/main/kotlin/gateway/VoiceEvent.kt
```diff
@@ -3,71 +3,68 @@ package dev.kord.voice.gateway
...
+                    // https://discord.com/developers/docs/topics/voice-connections#resuming-voice-connection-example-resumed-payload
+                    Resumed
                 }
-                endStructure(descriptor)
-                return data
+                OpCode.ClientDisconnect -> null // TODO what should we do with this event?
+                // OpCodes for Commands, they shouldn't be received
+                OpCode.Identify, OpCode.SelectProtocol, OpCode.Heartbeat, OpCode.Resume ->
+                    throw IllegalArgumentException("Illegal opcode for voice gateway event: $op")
+                OpCode.Unknown -> null // TODO should we throw an exception instead?
```
lukellmann, line 54: @DRSchlaubi @lost-illusi0n do you know what we should do for the [Client Disconnect](https://discord.com/developers/docs/topics/opcodes-and-status-codes#voice-voice-opcodes) opcode? It's not mentioned in the [voice documentation](https://discord.com/developers/docs/topics/voice-connections).

Also how likely is it to encounter unknown opcodes? If it isn't expected to happen, should we throw an exception for unknown opcodes like we do for the [non-voice gateway](https://github.com/kordlib/kord/blob/d1651a756f8db73909b0aa489040d63f32943619/gateway/src/commonMain/kotlin/Event.kt#L73)?

viztea: A lot of the voice protocol isn't documented (see my prs lol)

The Client Disconnect op is sent every time a user disconnects so it's probably best to ignore it to prevent spam

lukellmann: alright, i should probably add back the debug logging for this and then it's ok for now

DRSchlaubi: > A lot of the voice protocol isn't documented (see my prs lol)

So ig that means the answer to how likely it is to encounter unknown op codes is VERY LIKELY

DRSchlaubi (approved):

PR #929, opened 2024-03-20
Enable binary compatibility validation for KLibs

lukellmann:
See https://github.com/Kotlin/binary-compatibility-validator/blob/0.15.0-Beta.1/README.md#experimental-klib-abi-validation-support


DRSchlaubi:
I looked at this yesterday (before I sent you the news ^-^) however, there is an issue regarding #855, because K/N can only produce API dumps for targets it can compile for, this isn't a big deal for the check task, as we run on all hosts anyways, however now every PR needs to produce a dump for Linux, Windows and MAC, one way to fix this, would be to add a GitHub bot (via GitHub Actions or TeamCity) that will do that for you, if you comment something like `/dump-api`, that will produce the dump for you, however there is still an issue regarding that approach, as it looks like the dump is a single file, so 3 different CI runs would need to produce a single combined file

lukellmann:
i haven't experimented with native, but won't it just work for common code?

DRSchlaubi:
No as it dumps the platform API not the common api

lukellmann:
running this on the `feature/native` branch on my windows machine, i get this:
```
> Task :common:macosArm64ApiInfer
An ABI dump for target macosArm64 was inferred from the ABI generated for the following targets as the former target is not supported by the host compiler: [linuxArm64,linuxX64,mingwX64]. Inferred dump may not reflect an actual ABI for the target macosArm64. It is recommended to regenerate the dump on the host supporting all required compilation target.
```
so we only need an apple machine if we would want to add apple only api

DRSchlaubi:
When I did it it failed

lukellmann:
because the linux targets are disabled,  there is no target left for `voice` and `core-voice` except apple. using wsl or enabling the targets on windows for dumping fixes this.

DRSchlaubi:
> because the linux targets are disabled, there is no target left for `voice` and `core-voice` except apple. using wsl or enabling the targets on windows for dumping fixes this.

Neither of which we can do, but that does indeed explain it

lukellmann:
> Neither of which we can do, but that does indeed explain it

it's at least a workaround so we don't have to build a bot for this. at some point the issue with linux targets may be fixed (?) or we will add js/windows targets to voice. then the dump can be inferred from there.

DRSchlaubi:
The issue is with curl on windows, which will most definitely never get fixed

DRSchlaubi (approved):

PR #938, opened 2024-04-18
Create sourceLink URL with non-deprecated URI constructor

SebastianAigner:
Fixes failing Gradle import of the project since `-Werror` is set and the URL constructor is deprecated.


lukellmann (approved):
Thanks, haven't run into this myself yet as I'm still using JDK 17 to run Gradle (this deprecation is since Java 20).

DRSchlaubi:
I think I fixed this in #855 too ;)

PR #941, opened 2024-05-16
Switch to FlexInspect

DRSchlaubi:


common/src/commonMain/kotlin/entity/Interactions.kt
```diff
@@ -156,19 +156,19 @@ public sealed class Choice {
 
     public data class IntegerChoice(
         override val name: String,
-        override val nameLocalizations: Optional<Map<Locale, String>?>,
+        override val nameLocalizations: Optional<Map<Locale, String>?> = Optional.Missing(),
```
lukellmann, line 159: i think it should be fine to not have a default if it's not a class marked with `@Serializable`. that's what the KSP processor did too.

rest/src/commonMain/kotlin/builder/interaction/OptionsBuilder.kt
```diff
@@ -96,7 +96,6 @@ public sealed class BaseChoiceBuilder<T>(
  *
  * @see LocalizedNameCreateBuilder
  */
-@KordDsl
```
lukellmann, line 99: why was this removed?

DRSchlaubi: Probably because it's not a request builder, it shouldn't really mater as it's just a superclass for request builders, but I'll add it again

common/src/commonMain/kotlin/entity/Interactions.kt
DRSchlaubi: These classes should be serializable as `Choice` is serializable?

lukellmann: they are, but the default for optional is only needed with generated serializers (that will only be generated if the annotation is present), in this case there exists a [manual serializer implementation](https://github.com/kordlib/kord/blob/569b113a6ae0dc6c352be5cd2491634af65e16a1/common/src/commonMain/kotlin/entity/Interactions.kt#L175) that handles the optionals correctly

common/src/commonMain/kotlin/entity/Interactions.kt
```diff
@@ -527,7 +527,7 @@ public sealed class CommandArgument<out T> : Option() {
         override val name: String,
         override val type: ApplicationCommandOptionType,
         override val value: String,
-        override val focused: OptionalBoolean
+        override val focused: OptionalBoolean = OptionalBoolean.Missing
```
lukellmann, line 530: i would leave this as it was too

lukellmann (approved):

PR #942, opened 2024-05-21
Kotlin 2.0

DRSchlaubi:
Currently still waiting for KSP


lukellmann:
i would update other dependencies in a separate pr and do a release before we merge this, so that there is a up to date version for both k2 and non k2

voice/src/main/kotlin/gateway/VoiceEvent.kt
```diff
@@ -118,15 +112,17 @@ public object Resumed : VoiceEvent() {
...
+
+    private object Serializer : KSerializer<Resumed> {
+        @OptIn(InternalSerializationApi::class, ExperimentalSerializationApi::class)
+        override val descriptor = buildSerialDescriptor("dev.kord.voice.gateway.Resumed", StructureKind.OBJECT)
+        override fun serialize(encoder: Encoder, value: Resumed) = encoder.encodeStructure(descriptor) {}
+        override fun deserialize(decoder: Decoder) = decoder.decodeStructure(descriptor) {
+            val index = decodeElementIndex(descriptor)
+            if (index != CompositeDecoder.DECODE_DONE) throw SerializationException("Unexpected index $index")
+            Resumed
+        }
```
lukellmann, line 125: i think i can do a bit better here, will try later

PR #946, opened 2024-05-30
Optional.Missing and Optional.Null migrated to kotlin objects

zTrap:


lukellmann:
we might do this eventually, but i don't see the benefit of this binary incompatible change for now

PR #947, opened 2024-05-30
All flags and Color migrated to value classes

zTrap:
Related to https://github.com/kordlib/kord/issues/711


common/src/commonMain/kotlin/Color.kt
```diff
@@ -1,62 +1,55 @@
...
-
-        override fun deserialize(decoder: Decoder): Color = Color(decoder.decodeInt())
-
-        override fun serialize(encoder: Encoder, value: Color) {
-            encoder.encodeInt(value.rgb)
-        }
+        private val delegate = Int.serializer()
+        override val descriptor: SerialDescriptor = delegate.descriptor
+        override fun deserialize(decoder: Decoder): Color = Color(delegate.deserialize(decoder))
+        override fun serialize(encoder: Encoder, value: Color) = delegate.serialize(encoder, value.rgb)
```
DRSchlaubi, line 53: I think we no longer need a custom serializer now, right?

ksp-processors/src/main/kotlin/generation/bitflags/FactoryFunctions.kt
```diff
@@ -45,7 +45,6 @@ internal fun FileSpec.Builder.addFactoryFunctions() {
         }
     }
     addFactoryFunctionForIterable(baseParameterType = entityCN, IterableType.VARARG)
-    addFactoryFunctionForIterable(baseParameterType = collectionCN, IterableType.VARARG)
```
DRSchlaubi, line 48: Why was this removed?

ksp-processors/src/main/kotlin/generation/bitflags/BitFlagsGeneration.kt
```diff
@@ -79,9 +79,11 @@ internal fun BitFlags.generateFileSpec(originatingFile: KSFile) = fileSpecForGen
     }
     addClass(collectionCN) {
         addCollectionKDoc()
+        addAnnotation<JvmInline> {}
         addAnnotation<Serializable> {
             addMember("with·=·%T.Serializer::class", collectionCN)
         }
+        addModifiers(PUBLIC, VALUE)
```
DRSchlaubi, line 86: Doesn't kotlinpoet add the public modifier automatically?

ksp-processors/src/main/kotlin/generation/bitflags/BitFlagsGeneration.kt
```diff
@@ -79,9 +79,11 @@ internal fun BitFlags.generateFileSpec(originatingFile: KSFile) = fileSpecForGen
     }
     addClass(collectionCN) {
         addCollectionKDoc()
+        addAnnotation<JvmInline> {}
         addAnnotation<Serializable> {
             addMember("with·=·%T.Serializer::class", collectionCN)
```
DRSchlaubi, line 84: I think we don't need a custom serializer here either?

DRSchlaubi (changes requested):

ksp-processors/src/main/kotlin/generation/bitflags/FactoryFunctions.kt
zTrap: Value classes can't be used as vararg 😞

common/src/commonMain/kotlin/Color.kt
zTrap: Unfortunately, we need it here. We need to (de)serialize color without alpha, but backed value contains it (`packedRGB` prop). Alpha removes only by triggering `rgb` prop getter.

lukellmann:
I'd like to discuss this some more before deciding what to do with this, see https://discord.com/channels/556525343595298817/1245773750486438060

zTrap:
All changes moved to #886

PR #950, opened 2024-06-08
Added a method to check if a user is the same as the bot.

Taubsie:


core/src/commonMain/kotlin/entity/User.kt
```diff
@@ -37,6 +37,13 @@ public open class User(
     override val id: Snowflake
         get() = data.id
 
+    /**
+     * Returns true if the user is the same as the bot.
+     */
+    public val isSelf: Boolean by lazy {
+        id == kord.selfId
+    }
+
```
DRSchlaubi, line 46:
```suggestion
public val isSelf: Boolean get() = id == kord.selfId
```

I think the lazy actually causes worse performance in this case because it needs to allocate a bunch of objects

Taubsie: Changed it - makes sense

core/src/commonMain/kotlin/entity/User.kt
```diff
@@ -37,6 +37,10 @@ public open class User(
     override val id: Snowflake
         get() = data.id
 
+    /**
+     * Returns true if the user is the same as the bot.
+     */
+public val isSelf: Boolean get() = id == kord.selfId
```
NoComment1105, line 43: There is some funny indentation going on here
```suggestion
   public val isSelf: Boolean get() = id == kord.selfId
   
```

Taubsie: mb, only checked the code on github web

Taubsie: still looks weird tho

core/src/commonMain/kotlin/entity/User.kt
```diff
@@ -37,6 +37,11 @@ public open class User(
     override val id: Snowflake
         get() = data.id
 
+    /**
+     * Returns true if the user is the same as the bot.
+     */
+   public val isSelf: Boolean get() = id == kord.selfId
```
NoComment1105, line 43:
```suggestion
    public val isSelf: Boolean get() = id == kord.selfId
```

NoComment1105: Might fix it

NoComment1105 (approved):
Looks good to me now :D

core/src/commonMain/kotlin/entity/User.kt
Taubsie: We commited the same thing at the same time - mind if I just resolve that?

DRSchlaubi: Did you break it again?

NoComment1105: yeah just resolve it, It looks fine in the changes window

NoComment1105:
Just needs an API dump to get checks to pass

DRSchlaubi:
Please run `gradlew apiDump`

Taubsie:
\o/
Should be working now

DRSchlaubi (approved):

PR #959, opened 2024-07-22
Add MessageBuilder.Limits object

Galarzaa90:
Similar to `EmbedBuilder.Limits`,  added one for `MessageBuilder`, altough this one is an interface, so I'm not sure if there are any side effects to this?

Also, I'm missing one for files/attachments, but then when I saw that they are two separate fields, I wasn't sure how to do it.

### DIFF
--- a/rest/api/rest.api
+++ b/rest/api/rest.api
@@ -2143,6 +2143,12 @@ public final class dev/kord/rest/builder/message/MessageBuilder$DefaultImpls {
 	public static fun addFile (Ldev/kord/rest/builder/message/MessageBuilder;Ljava/lang/String;Lio/ktor/client/request/forms/ChannelProvider;)Ldev/kord/rest/NamedFile;
 }
 
+public final class dev/kord/rest/builder/message/MessageBuilder$Limits {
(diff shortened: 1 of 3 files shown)


lukellmann:
> Similar to `EmbedBuilder.Limits`,  added one for `MessageBuilder`, altough this one is an interface, so I'm not sure if there are any side effects to this?

No, that shouldn't be an issue.

lukellmann:
I'm curious, what do you use these values for? Hardcoding them into Kord might be an issue if Discord increases them at some point.

Galarzaa90:
> I'm curious, what do you use these values for? Hardcoding them into Kord might be an issue if Discord increases them at some point.

These ones in particular, I only have one instance where I use them, because for the most part I use embeds. But same as `EmbedBuilder.Limits`, I use them to limit information to not exceed the limits.

I know that hardcoding might be an issue, but either you hardcode them here or you hardcode them youself in your implementation, and since we already have `EmbedBuilder.Limits`, thought it would make sense to have these as well for completeness/consistency. Might come handy for average users.

If the value is updated, I would have to update my hardcoded values anyway. If I'm using the constants from here, I would temporarily replace them with my own constants, while the value here is updated.

PR #961, opened 2024-07-29
Publish snapshots to https://repo.kord.dev/snapshots

lukellmann:

PR #962, opened 2024-07-29
Update module READMEs

lukellmann:
This reduces duplication in the README files of the individual modules and fixes some outdated links and code.

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -1,6 +1,6 @@
 # Kord gateway
 
-A low-level implementation of discord's [gateway](https://discordapp.com/developers/docs/topics/gateway).
+A low-level implementation of discord's [gateway](https://discord.com/developers/docs/topics/gateway).
 
 ## Example usage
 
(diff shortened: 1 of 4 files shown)


lukellmann:
also so i don't have to update all these readmes in #961

gateway/README.md
```diff
@@ -11,7 +11,9 @@ suspend fun main(args: Array<String>) {
     val gateway = DefaultGateway { // optional builder for custom configuration
         client = HttpClient {
             install(WebSockets)
-            install(JsonFeature)
+            install(ContentNegotiation) {
+                json()
+            }
```
DRSchlaubi, line 16: I am quite sure this is unnecessary

DRSchlaubi (changes requested):

gateway/README.md
lukellmann: what part specifically? all of it?

lukellmann: or do you mean because it is gateway and we don't do http requests?

DRSchlaubi: we neither do http requests nor do we use the content negotiation websocket feature

DRSchlaubi (approved):

PR #964, opened 2024-08-11
Fix getMemberOrNull and getGuildMembers caching user data instead of member data

Galarzaa90:
[As reported in Discord](https://canary.discord.com/channels/556525343595298817/1271484633610981493), these calls are storing the results to the user cache, so subsequent calls would not find them from cache.

When using `cacheWithCachingRestFallback`, this results in REST calls being made every single time to get members, unless they got cached from a gateway event.

### DIFF
--- a/core/src/commonMain/kotlin/supplier/StoreEntitySupplier.kt
+++ b/core/src/commonMain/kotlin/supplier/StoreEntitySupplier.kt
@@ -61,7 +61,7 @@ public class StoreEntitySupplier(
     }
 
     override suspend fun getMemberOrNull(guildId: Snowflake, userId: Snowflake): Member? {
-        return storeAndReturn(supplier.getMemberOrNull(guildId, userId)) { it.data }
+        return storeAndReturn(supplier.getMemberOrNull(guildId, userId)) { it.memberData }
     }
 
     override suspend fun getMessageOrNull(channelId: Snowflake, messageId: Snowflake): Message? {
@@ -105,7 +105,7 @@ public class StoreEntitySupplier(
     }
 
     override fun getGuildMembers(guildId: Snowflake, limit: Int?): Flow<Member> {
-        return storeOnEach(supplier.getGuildMembers(guildId, limit)) { it.data }
+        return storeOnEach(supplier.getGuildMembers(guildId, limit)) { it.memberData }
     }
 
     override fun getGuildVoiceRegions(guildId: Snowflake): Flow<Region> {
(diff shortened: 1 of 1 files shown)


lukellmann:
You want to store both `UserData` and `MemberData`, see [GuildEventHandler](https://github.com/kordlib/kord/blob/5bcb0e3d927de41359cf195c0737e0e16be27dca/core/src/commonMain/kotlin/gateway/handler/GuildEventHandler.kt#L270). I think it needs some extra logic (`storeAndReturn` and `storeOnEach` won't suffice if I'm not wrong).

Galarzaa90:
> You want to store both `UserData` and `MemberData`, see [GuildEventHandler](https://github.com/kordlib/kord/blob/5bcb0e3d927de41359cf195c0737e0e16be27dca/core/src/commonMain/kotlin/gateway/handler/GuildEventHandler.kt#L270). I think it needs some extra logic (`storeAndReturn` and `storeOnEach` won't suffice if I'm not wrong).

For `getMemberOrNull` this could work:

```kt
override suspend fun getMemberOrNull(guildId: Snowflake, userId: Snowflake): Member? {
    return storeAndReturn(supplier.getMemberOrNull(guildId, userId)) {
        cache.put(it.data)
        it.memberData
    }
}
```

For `getGuildMembers`, that won't work because `cache.put` is a suspend function and the transform on `storeOnEach` is `crossinline`.

Should additional utility functions be added for these single uses?

Alternative for `getGuildMembers`:

```kt
override fun getGuildMembers(guildId: Snowflake, limit: Int?): Flow<Member> {
    return supplier.getGuildMembers(guildId, limit).onEach { 
        cache.put(it.data)
        cache.put(it.memberData)
    }
}
```

lukellmann:
> or without `storeAndReturns`, so it doesn't look "hacky":
> 
> ```kotlin
> override suspend fun getMemberOrNull(guildId: Snowflake, userId: Snowflake): Member? {
>     return supplier.getMemberOrNull(guildId, userId)?.also { 
>         cache.put(it.data)
>         cache.put(it.memberData)
>     }
> }
> ```
Yes, I'd prefer something like this.

> For `getGuildMembers`, that won't work because `cache.put` is a suspend function and the transform on `storeOnEach` is `crossinline`.
> 
> Should additional utility functions be added for these single uses?
> 
> Alternative for `getGuildMembers`:
> 
> ```kotlin
> override fun getGuildMembers(guildId: Snowflake, limit: Int?): Flow<Member> {
>     return supplier.getGuildMembers(guildId, limit).onEach { 
>         cache.put(it.data)
>         cache.put(it.memberData)
>     }
> }
> ```
I think we don't need an extra function and make it complicated, keeping it simple with something like your alternative should be fine.

Note that in both cases using an expression body function and a named lambda parameter instead of `it` might be slightly more readable IMHO.

Galarzaa90:
Changes applied

lukellmann (approved):
thanks :)

PR #965, opened 2024-08-11
Switch to codegen.kt

DRSchlaubi:


DRSchlaubi:
I think codegen kt doesn't run on jdk 8

buildSrc/src/main/kotlin/Projects.kt
```diff
@@ -29,5 +29,5 @@ val Project.isRelease get() = tag != null
 
 object Repo {
     const val releasesUrl = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
-    const val snapshotsUrl = "https://oss.sonatype.org/content/repositories/snapshots/"
```
lukellmann, line 32: shouldn't we do it like this in https://github.com/kordlib/kord/pull/961?

ksp-processors/src/main/kotlin/generation/shared/Names.kt
```diff
@@ -1,16 +1,24 @@
+@file:OptIn(ExperimentalContracts::class, DelicateKotlinPoetApi::class)
```
lukellmann, line 1: i don't like the opt-ins here, why not keep the old approach of using strings?

ksp-processors/src/main/kotlin/KSPUtils.kt
lukellmann, line 1: is there a reason you didn't just delete the file?

ksp-processors/src/main/kotlin/KotlinPoetDsl.kt
lukellmann, line 1: ditto

lukellmann:
> I think codegen kt doesn't run on jdk 8

the build failure looks more like a dependency resolution issue

DRSchlaubi:
Yeah because codegen.kt defines a jdk dependency in the Gradle meta

ksp-processors/src/main/kotlin/generation/shared/Names.kt
DRSchlaubi: Because this one is actually type-safe and I like it

DRSchlaubi:
Since the CI now uses Java 21, this should work again

PR #968, opened 2024-08-15
fix: Implementation for Equals & Hashcode in Builder

Distractic:
# Context

When we compare the builder, like `EmbedBuilder`, the comparison between two instances will always return false no matter the field set. For example:

```kt
println(EmbedBuilder() == EmbedBuilder()) // false
println(EmbedBuilder().apply { title = "title" } == EmbedBuilder().apply { title = "title" }) // false

println(EmbedBuilder().hashCode()) // 2096057945
println(EmbedBuilder().hashCode()) // 1689843956
```

This behavior seems to be a bug and needs to be fixed because according to the result, equals & hashcode methods don't work.

# Usage

In my bot, when I render my components, I would like to compare the last `InteractionResponseModifyBuilder` and the new from the rendering to send only the changed components (embed / actionRow / text) to the discord API and consequently reduce the network load of the bot

### DIFF
--- a/rest/src/commonMain/kotlin/builder/application/ApplicationRoleConnectionMetadata.kt
+++ b/rest/src/commonMain/kotlin/builder/application/ApplicationRoleConnectionMetadata.kt
@@ -39,6 +39,20 @@ public class ApplicationRoleConnectionMetadataRecordsBuilder :
     }
 
     override fun toRequest(): List<DiscordApplicationRoleConnectionMetadata> = records.map { it.toRequest() }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
+
+        other as ApplicationRoleConnectionMetadataRecordsBuilder
+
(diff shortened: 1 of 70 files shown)


lukellmann:
As already discussed [on Discord](https://discord.com/channels/556525343595298817/1273031671343743089) I don't really feel like this is the right solution. Kord's builder classes are intended to be short-lived, mutable (!) objects that help to create immutable requests and closely model fields of the Discord API (e.g. optional vs. nullable).

Having a data model specific to your domain logic seems to be a way more robust solution to your usecase.

I'm also not aware that implementing `equals` and `hashCode` is frequently done for similar builder classes in other projects.

Distractic:
Understood. I thought that could be useful for several projects (like Kordex, I think they keep the builders to compare in a next time) but for my need, I can create my own builders, that allows me to have more flexibility than using Kord builder

lukellmann:
Nevertheless, thank you for your work :)

PR #971, opened 2024-08-25
Add support for enforceNonce

DRSchlaubi:


rest/src/commonMain/kotlin/builder/message/create/UserMessageCreateBuilder.kt
```diff
@@ -25,6 +24,15 @@ public class UserMessageCreateBuilder : AbstractMessageCreateBuilder(), RequestB
     /** A value that can be used to verify a message was sent (up to 25 characters). */
     public var nonce: String? by ::_nonce.delegate()
 
+    private var _enforeceNonce: OptionalBoolean = OptionalBoolean.Missing
```
lukellmann, line 27: typo, should be `_enforceNonce`

DRSchlaubi: I think it's Spanish

lukellmann (approved):

PR #973, opened 2024-08-25
Add support for application emojis

DRSchlaubi:


rest/src/commonMain/kotlin/service/ApplicationService.kt
```diff
@@ -1,10 +1,81 @@
...
+        builder: EmojiModifyBuilder.() -> Unit
+    ): DiscordEmoji {
+        contract { callsInPlace(builder, InvocationKind.EXACTLY_ONCE) }
+        return modifyApplicationEmoji(appId, emojiId, EmojiModifyBuilder().apply(builder).toRequest())
+    }
+
+    public suspend inline fun deleteApplicationEmoji(
+        appId: Snowflake,
+        emojiId: Snowflake,
+        reason: String? = null,
```
psmarko, line 73: Application emojis don't support the audit log header.

```suggestion
```

psmarko (changes requested):
Few little things need fixing like route keys and Discord not supporting audit log reasons.

rest/src/commonMain/kotlin/service/ApplicationService.kt
```diff
@@ -1,10 +1,81 @@
...
+        keys[Route.ApplicationId] = appId
+    }
+
+    public suspend fun getApplicationEmoji(appId: Snowflake, emojiId: Snowflake): DiscordEmoji =
+        call(Route.GetApplicationEmoji) {
+            keys[Route.ApplicationId] = appId
+            keys[Route.EmojiId] = emojiId
+        }
+
+    public suspend fun createApplicationEmoji(appId: Snowflake, request: EmojiCreateRequest): DiscordEmoji =
```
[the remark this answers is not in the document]
DRSchlaubi: I think breaking the API for this is not worth it, so let's just leave this as is

common/src/commonMain/kotlin/entity/DiscordEmoji.kt
```diff
@@ -51,3 +51,11 @@ public data class DiscordPartialEmoji(
...
     val animated: OptionalBoolean = OptionalBoolean.Missing,
 )
+
+/**
+ * List of application [emojis][DiscordEmoji].
+ *
+ * @property items the list of emojis
+ */
+@Serializable
+public data class ApplicationEmojis(val items: List<DiscordEmoji>) : List<DiscordEmoji> by items
```
lukellmann, line 61: `equals` and `hashCode` produced by data classes don't conform to the requirements of `List`, I suggest to simply not implement `List`.

core/src/commonMain/kotlin/entity/Emoji.kt
```diff
@@ -118,33 +152,28 @@ public class GuildEmoji(
...
+    public val member: MemberBehavior? get() = userId?.let { MemberBehavior(guildId, it, kord) }
 
     /**
      * Requests to delete this emoji, with the given [reason].
      *
      * @param reason the reason showing up in the audit log
      * @throws RequestException if anything went wrong during the request.
      */
-    public suspend fun delete(reason: String? = null) {
+    public suspend fun delete(reason: String?) {
```
lukellmann, line 171: This is a binary incompatible change, please keep the default argument.

core/src/commonMain/kotlin/Kord.kt
```diff
@@ -106,6 +107,16 @@ public class Kord(
...
 
+    /**
+     * Flow of [application emojis][ApplicationEmoji].
+     */
+    public val emojis: Flow<ApplicationEmoji> = flow {
+        rest.application.getApplicationEmojis(selfId).forEach {
+            val data = EmojiData.from(selfId, it.id!!, it)
+            ApplicationEmoji(data, this@Kord)
+        }
+    }
```
lukellmann, line 118: `emit` is missing, this flow will always be empty.

core/src/commonMain/kotlin/entity/Emoji.kt
```diff
@@ -48,21 +48,17 @@ public class StandardEmoji(override val name: String) : Emoji {
 }
 
 /**
- * An instance of an [emoji](https://discord.com/developers/docs/resources/emoji#emoji-object) belonging to a specific
- * [Guild].
+ * Superclass for all non standard emojis.
```
lukellmann, line 51: The comment on `Emoji` should be updated to say what the subtypes are.

core/src/commonMain/kotlin/entity/Emoji.kt
```diff
@@ -48,21 +48,17 @@ public class StandardEmoji(override val name: String) : Emoji {
...
+ *
+ * @see GuildEmoji
+ * @see ApplicationEmoji
  */
-public class GuildEmoji(
-    public val data: EmojiData,
-    override val kord: Kord,
-    override val supplier: EntitySupplier = kord.defaultSupplier
-) : Emoji, KordEntity, Strategizable {
+public sealed class CustomEmoji : Emoji, KordEntity, Strategizable {
```
lukellmann, line 56: sealed interface would be nicer

lukellmann, line 56: Should abstract override `withStrategy` to return `CustomEmoji`.

rest/src/commonMain/kotlin/route/Route.kt
```diff
@@ -1118,4 +1121,34 @@ public sealed class Route<T>(
...
+    public object PatchApplicationEmoji : Route<DiscordEmoji>(
+        HttpMethod.Patch,
+        "/applications/$ApplicationId/emojis/$EmojiId",
+        DiscordEmoji.serializer()
+    )
+
+    public object DeleteApplicationEmoji : Route<Unit>(
+        HttpMethod.Delete,
+        "/applications/$ApplicationId/emojis/$EmojiId",
+        Unit.serializer()
```
lukellmann, line 1152: Please use `NoStrategy` instead.

rest/src/commonMain/kotlin/route/Route.kt
```diff
@@ -1118,4 +1121,34 @@ public sealed class Route<T>(
...
+
+    public object GetApplicationEmoji : Route<DiscordEmoji>(
+        HttpMethod.Get,
+        "/applications/$ApplicationId/emojis/$EmojiId",
+        DiscordEmoji.serializer()
+    )
+
+    public object PostApplicationEmoji : Route<DiscordEmoji>(
+        HttpMethod.Post,
+        "/applications/$ApplicationId/emojis/",
```
lukellmann, line 1139: Please remove trailing /

rest/src/commonMain/kotlin/service/ApplicationService.kt
```diff
@@ -1,10 +1,72 @@
...
+            body(EmojiCreateRequest.serializer(), request)
+        }
+
+    public suspend fun createApplicationEmoji(
+        appId: Snowflake,
+        name: String,
+        image: Image,
+    ): DiscordEmoji = createApplicationEmoji(appId, EmojiCreateBuilder(name, image).toRequest())
+
+    public suspend inline fun modifyApplicationEmoji(
```
lukellmann, line 44: Please remove `inline`.

rest/src/commonMain/kotlin/service/ApplicationService.kt
```diff
@@ -1,10 +1,72 @@
...
+        appId: Snowflake,
+        emojiId: Snowflake,
+        builder: EmojiModifyBuilder.() -> Unit
+    ): DiscordEmoji {
+        contract { callsInPlace(builder, InvocationKind.EXACTLY_ONCE) }
+        return modifyApplicationEmoji(appId, emojiId, EmojiModifyBuilder().apply(builder).toRequest())
+    }
+
+    public suspend inline fun deleteApplicationEmoji(
```
lukellmann, line 65: Please remove `inline`.

lukellmann (changes requested):
Sorry for the late review after merge, can you address my comments in a followup PR @DRSchlaubi?

common/src/commonMain/kotlin/entity/DiscordEmoji.kt
lukellmann: Or better remove this class as a whole.

lukellmann: Nope, don't remove, discord returns a JSON object with `items` in it.

DRSchlaubi: We could just make it a value class instead of a data class

core/src/commonMain/kotlin/entity/Emoji.kt
DRSchlaubi: It's already mentioned using @see, so why mention it twice

common/src/commonMain/kotlin/entity/DiscordEmoji.kt
lukellmann: that doesn't fix it either

core/src/commonMain/kotlin/entity/Emoji.kt
lukellmann: I was referring to this line: https://github.com/kordlib/kord/blob/d23b8c5a2658ddde7e4e367d5ac8397f62ac1928/core/src/commonMain/kotlin/entity/Emoji.kt#L21

lukellmann:
See https://github.com/kordlib/kord/compare/2002d0c0ec1a0336f589b68a0d2ca2af1ce73177...eaf5dbe1755a4b2005098c6a931465266fed0b65 for a cleaner diff of ABI changes.

PR #974, opened 2024-08-30
Add `MemberProfileAutoModerationRule`

lukellmann:
The DSL for creating/editing `MemberProfileAutoModerationRule`s looks like this:

```kotlin
val rule = guild.createMemberProfileAutoModerationRule("name") {
    anywhereKeyword("cat")
    regexPattern("^b(a|@)d\$w(o|0)rd(s|$)")
    allowPrefixKeyword("black cat")
    blockMemberInteraction()
    sendAlertMessage(channelId)
    enabled = true
}
println(rule)
```

See https://github.com/discord/discord-api-docs/pull/6040


DRSchlaubi:
I think an `@Language("regexp")` should be used, which should be doable using @OptionalExpectation

lukellmann:
> I think an `@Language("regexp")` should be used, which should be doable using @OptionalExpectation

does this work for rust regex? auto mod [supposedly uses rust regex syntax](https://discord.com/developers/docs/resources/auto-moderation#auto-moderation-rule-object-trigger-metadata).

DRSchlaubi:
> > I think an `@Language("regexp")` should be used, which should be doable using @OptionalExpectation
> 
> does this work for rust regex? auto mod [supposedly uses rust regex syntax](https://discord.com/developers/docs/resources/auto-moderation#auto-moderation-rule-object-trigger-metadata).

regexp is simply the name the regex plugin uses, it stands for **RegularExp**ression, not regexp

lukellmann:
> > > I think an `@Language("regexp")` should be used, which should be doable using @OptionalExpectation
> > 
> > does this work for rust regex? auto mod [supposedly uses rust regex syntax](https://discord.com/developers/docs/resources/auto-moderation#auto-moderation-rule-object-trigger-metadata).
> 
> regexp is simply the name the regex plugin uses, it stands for **RegularExp**ression, not regexp

yes, i know. the concern that i'm having is that [`@Language("RegExp")`](https://github.com/JetBrains/java-annotations/blob/master/java-annotations%2Fsrc%2Fmain%2Fjava%2Forg%2Fintellij%2Flang%2Fannotations%2FLanguage.java) is for [`java.util.regex.Pattern`](https://docs.oracle.com/en%2Fjava%2Fjavase%2F22%2Fdocs%2Fapi%2F%2F/java.base/java/util/regex/Pattern.html) and the regex syntax might be different from [rust's `regex::Regex`](https://docs.rs/regex/latest/regex/struct.Regex.html) (or whatever discord uses).

DRSchlaubi:
Nah I think that's fine

lukellmann:
hm, i'm not sure if it's really worth the effort of an expect/actual annotation. even less if it might even provide slightly wrong results.

PR #978, opened 2024-09-03
Replace Gradle toolchain with `-Xjdk-release`

lukellmann:
See https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/, https://github.com/kordlib/gradle-tools/pull/3 and https://kotlinlang.org/docs/compiler-reference.html#xjdk-release-version


lukellmann:
i've manually verified that the new artifact is identical to the old one

DRSchlaubi (approved):

PR #981, opened 2024-09-19
Fix misuse of reified type parameters

lukellmann:
[KT-52469](https://youtrack.jetbrains.com/issue/KT-52469) issues warnings for `Kord.getGlobalApplicationCommandOf`, `Kord.getGlobalApplicationCommandOfOrNull` and `VoiceState.getChannelOrNull` since Kotlin 2.1.0-dev-7621.


DRSchlaubi:
I do not understand this

lukellmann:
This was a weird interaction between type inference and reified type parameters that will cause a warning in Kotlin 2.1.0. These functions were in fact buggy:

---

`Kord.getGlobalApplicationCommandOf` and `Kord.getGlobalApplicationCommandOfOrNull` were basically ignoring the type argument for `T` (it wasn't `reified`) and calling `EntitySupplier.getGlobalApplicationCommandOf`/`EntitySupplier.getGlobalApplicationCommandOfOrNull` with `GlobalApplicationCommand` for the `reified T` of the `EntitySupplier` functions unconditionally. Meaning you would get a `ClassCastException` instead of `null` for this code:
```kotlin
val command = kord.createGlobalUserCommand("user")
val shouldBeNull: GlobalChatInputCommand? = kord.getGlobalApplicationCommandOfOrNull<GlobalChatInputCommand>(command.id)
println(shouldBeNull)
```
The type parameter also had no bound so this was possible but also threw a `ClassCastException`:
```kotlin
val command = kord.createGlobalUserCommand("user")
val howShouldThisBeAString: String = kord.getGlobalApplicationCommandOf<String>(command.id)
println(howShouldThisBeAString)
```
With this PR, the first example works as expected and the second one no longer compiles.

---

`VoiceState.getChannelOrNull` also wasn't properly checking for and casting to `BaseVoiceChannelBehavior` but only for/to `ChannelBehavior`.

DRSchlaubi (approved):

PR #983, opened 2024-09-21
fix: `Json` configuration not applying when using custom `HttpClient`

JellyBrick:


JellyBrick:
Any updates?

lukellmann:
i will take a look later

lukellmann:
We actually don't use the `ContentNegotiation` plugin, right @DRSchlaubi? The only thing it seems to do for us is adding an `Accept: application/json` header to our requests, which is probably useless. So we might instead just remove this configuration of the `ContentNegotiation` plugin.

DRSchlaubi:
That's correct we don't use it as of now

PR #986, opened 2024-10-12
Update dependencies

lukellmann:
* [Gradle](https://github.com/gradle/gradle) 8.10 -> 8.10.2
* [Kotlin](https://github.com/JetBrains/kotlin) 2.0.20 -> 2.0.21
* [Ktor](https://github.com/ktorio/ktor) 2.3.12 -> 3.0.0
* [kotlinx.coroutines](https://github.com/Kotlin/kotlinx.coroutines) 1.8.1 -> 1.9.0
* [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization) 1.7.2 -> 1.7.3
* [kotlin-node](https://github.com/JetBrains/kotlin-wrappers) 20.14.10-pre.800 -> 22.5.4-pre.818
* [KSP](https://github.com/google/ksp) 2.0.20-1.0.24 -> 2.0.21-1.0.25
* [JUnit Jupiter](https://github.com/junit-team/junit5) 5.11.0 -> 5.11.2
* [JUnit Platform](https://github.com/junit-team/junit5) 1.11.0 -> 1.11.2
* [MockK](https://github.com/mockk/mockk) 1.13.12 -> 1.13.13
* [gradle-buildconfig-plugin](https://github.com/gmazzo/gradle-buildconfig-plugin) 5.4.0 -> 5.5.0

`io.ktor:ktor-client-websockets` is an empty artifact, support for WebSockets is included in `io.ktor:ktor-client-core`, so the dependency was removed from `kord-gateway`.

### DIFF
--- a/voice/src/main/kotlin/udp/RTPPacket.kt
+++ b/voice/src/main/kotlin/udp/RTPPacket.kt
@@ -4,7 +4,10 @@ import dev.kord.voice.io.ByteArrayView
 import dev.kord.voice.io.MutableByteArrayCursor
 import dev.kord.voice.io.mutableCursor
 import dev.kord.voice.io.view
-import io.ktor.utils.io.core.*
+import kotlinx.io.Source
+import kotlinx.io.readByteArray
+import kotlinx.io.readUInt
+import kotlinx.io.readUShort
 import kotlin.experimental.and
 
 internal const val RTP_HEADER_LENGTH = 12
@@ -38,8 +41,8 @@ public data class RTPPacket(
     public companion object {
         internal const val VERSION = 2
 
-        public fun fromPacket(packet: ByteReadPacket): RTPPacket? = with(packet) base@{
-            if (remaining <= 13) return@base null
+        public fun fromPacket(packet: Source): RTPPacket? = with(packet) base@{
+            if (!request(byteCount = 14)) return@base null
 
             /*
              * first byte | bit table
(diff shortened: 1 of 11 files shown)


voice/src/main/kotlin/udp/RTPPacket.kt
```diff
@@ -38,8 +41,8 @@ public data class RTPPacket(
     public companion object {
         internal const val VERSION = 2
 
-        public fun fromPacket(packet: ByteReadPacket): RTPPacket? = with(packet) base@{
-            if (remaining <= 13) return@base null
+        public fun fromPacket(packet: Source): RTPPacket? = with(packet) base@{
+            if (!request(byteCount = 14)) return@base null
```
lukellmann, line 45: @lost-illusi0n was the `<=` here and below intended? The RTP header is 12 bytes and then you required to have 2 more bytes.

lost-illusi0n: I believe I intentionally required it to have at least one more byte (12 byte header + a one byte payload), but wrote it like that. Woops. This could probably be changed, but a one byte payload is unlikely in reality (luckily). To be truly correct, an RTP packet could just be a header, with no payload. Maybe supporting that wouldn't be a bad idea. I just didn't bother (or forgot to) three years ago.

For [the second `<=` case](https://github.com/kordlib/kord/blame/3ef564ccdaa1659ff04305d76ee33756e47cd1e0/voice/src/main/kotlin/udp/RTPPacket.kt#L80), that was to keep the intentional requirement of a payload in place. Made the same mistake of requiring a two byte payload again though. At least it's consistent)) If the requirement of a payload is dropped, then these extra data requirements can also be dropped. If not, then changing the upper bound of the data checks to one less should be fine too.

lukellmann: Alright, thanks for taking a look at this after 3 years :D

I think I'm going to leave the accidental +2 byte requirement in place for now. This PR is supposed to be a dependency upgrade and as you said it's unlikely that this causes problems in practice.

I'm [working](https://github.com/lukellmann/kord/tree/new-voice-encryption) on the [new encryption modes](https://discord.com/developers/docs/change-log#voice-encryption-modes) at the moment and I'm considering to touch the RTP logic with that too, so I'd rather bring in the change there.

DRSchlaubi (approved):

PR #995, opened 2025-01-03
Add platform field to ActivityData and DiscordActivity

giginet:
## Summary

This PR adds `platform` property to `Activity`.

When users play games from consoles. The Discord Activity API returns the below response. 

```json
{
  "type": 0,
  "timestamps": {
    "start": 1702191977428
  },
  "platform": "ps5",
  "name": "Sonic Superstars",
  "id": "20ac8563beebf2ec",
  "flags": 0,
  "created_at": 1702191977427,
  "assets": {
    "small_image": "mp:external/at5adOeXGcCj4jN9_zB0SiIL3FR8wOxq_d39CDh9Pqo/https/cdn.discordapp.com/assets/playstation-logo-rich-presence.png"
  }
}
```

According to [the documentation](https://discord.com/developers/docs/events/gateway-events#activity-object), this field is undocumented, but I confirmed this value was returned. 

See also the [this PR](https://github.com/Rapptz/discord.py/pull/9677) for Discord.py.


DRSchlaubi:
Are these documented, if so could we use a "kordenum"

DRSchlaubi:
Something like this: https://github.com/kordlib/kord/blob/main/common/src/commonMain/kotlin/entity/DiscordMessage.kt#L1-L38

giginet:
I found an implementation like this.
https://github.com/kyranet/discord-api-types/blob/0aa384ca9a8fe5a35cc4d13f3e2c0b4e5a55c0ca/deno/payloads/v10/gateway.ts#L229

But these platforms are inaccurate, so it's safer to be just a string.

DRSchlaubi:
> I found an implementation like this. https://github.com/kyranet/discord-api-types/blob/0aa384ca9a8fe5a35cc4d13f3e2c0b4e5a55c0ca/deno/payloads/v10/gateway.ts#L229
> 
> But these platforms are inaccurate, so it's safer to be just a string.

Kord enums do have an `Unknown` type which will be used in the case of an invalid type, however I can't find these type to the be documented anywhere

giginet:
OK, I’ll define the enum values based on this.

giginet:
@DRSchlaubi I defined the known platforms as the Kord Enum.

lukellmann (approved):

lukellmann:
thanks!

giginet:
@lukellmann I’d like to use this feature. Could you release a new version.

DRSchlaubi:
You can use the snapshot

Issue #1006, opened 2025-04-18
WasmJs support


mani1232:
Since there is already support for Js I think it would be easy to add WasmJs

PR #1007, opened 2025-04-18
Feature/dependencies and WasmJs

mani1232:
#1006


DRSchlaubi (changes requested):
- Update gradle checksum
- There should be a shared target, like wasmJsShared
- Maybe extract some of the wasm/js shared compilation options into a shared function

DRSchlaubi:
Lgtm, will look into this further tmr

mani1232:
We also need https://github.com/kordlib/cache/issues/17

PR #1009, opened 2025-04-26
Feat/components v2

DRSchlaubi:


common/src/commonMain/kotlin/entity/DiscordMessage.kt
```diff
@@ -66,6 +66,7 @@
             "SuppressNotifications", shift = 12, kDoc = "This message will not trigger push and desktop notifications.",
         ),
         Entry("IsVoiceMessage", shift = 13, kDoc = "This message is a voice message."),
+        Entry("IsComponentsV2", shift = 15, kDoc = "Enable V2 of the Components API"),
```
lukellmann, line 69:
```suggestion
        Entry(
            "IsComponentsV2", shift = 15,
            kDoc = "allows you to create fully [component](https://discord.com/developers/docs/components/overview)-" +
                "driven messages.",
        ),
```
also `docUrl` is out of date, it should be https://discord.com/developers/docs/resources/message#message-object-message-flags

common/src/commonMain/kotlin/entity/DiscordMessage.kt
```diff
@@ -87,7 +88,7 @@
         Entry("Rich", stringValue = "rich", kDoc = "Generic embed rendered from embed attributes."),
         Entry("Image", stringValue = "image", kDoc = "Image embed."),
         Entry("Video", stringValue = "video", kDoc = "Video embed."),
-        Entry("Gifv", stringValue = "gifv", kDoc = "Animated gif image embed rendered as a video embed."),
+        Entry("Gifv", stringValue = "gifv", kDoc = "Animated gif media embed rendered as a video embed."),
```
lukellmann, line 91: why all this renaming from image to media?

lukellmann (changes requested):
review of common

rest/src/commonMain/kotlin/builder/component/ComponentContainerBuilder.kt
```diff
@@ -0,0 +1,68 @@
...
+ * @see ContainerBuilder
+ */
+public interface ComponentContainerBuilder {
+    @KordInternal
+    public fun addComponent(component: ContainerComponentBuilder)
+
+    /**
+     * Adds a separator with specified [spacing].
+     */
+    public fun separator(spacing: SeparatorSpacingSize): Unit = separator {
```
zTrap, line 21: `divider: Boolean` is necessary here. Also, `divider` and `spacing` should be nullable with `null` as default value, because they can be omitted and `true` and `Small (1)` will be used respectively

rest/src/commonMain/kotlin/builder/component/v2.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    override fun build(): DiscordComponent = DiscordChatComponent(
+        type = ComponentType.MediaGallery,
+        items = Optional(items.map { it.build() }),
+    )
+}
+
+public class SeparatorBuilder : ContainerComponentBuilder {
+    private var _spacing: Optional<SeparatorSpacingSize> = Optional.Missing()
+
```
zTrap, line 66:
```suggestion
private var _divider: OptionalBoolean = OptionalBoolean.Missing

public var divider: Boolean? by ::_divider.delegate()
```

rest/src/commonMain/kotlin/builder/component/v2.kt
```diff
@@ -0,0 +1,90 @@
...
+    public var spacing: SeparatorSpacingSize? by ::_spacing.delegate()
+
+    override fun build(): DiscordComponent = DiscordChatComponent(
+        type = ComponentType.Separator,
+        spacing = _spacing
```
zTrap, line 71:
```suggestion
        spacing = _spacing,
        divider = _divider,
```

rest/src/commonMain/kotlin/builder/component/v2.kt
```diff
@@ -0,0 +1,90 @@
...
+        UnfurledMediaItem(url),
+        _description,
+        _spoiler,
+    )
+}
+
+public class MediaGalleryBuilder : ContainerComponentBuilder {
+    public val items: MutableList<MediaGalleryItemBuilder> = mutableListOf()
+
+    public fun item(url: String, builder: MediaGalleryItemBuilder.() -> Unit) {
```
zTrap, line 54:
```suggestion
    public fun item(url: String, builder: MediaGalleryItemBuilder.() -> Unit = {}) {
```

rest/src/commonMain/kotlin/builder/component/ComponentContainerBuilder.kt
zTrap, line 21:
```suggestion
    public fun separator(
        spacing: SeparatorSpacingSize? = null,
        divider: Boolean? = null,
    ): Unit = separator {
        this.spacing = spacing
        this.divider = divider
    }
```

DRSchlaubi: I don't really see the point of that, you can just call the seperator function with the builder yourself

zTrap: I agree, but there is no point to have partial member function. My opinion is "all or nothing" :D

ptitjes:
Hey @DRSchlaubi ! Nice PR. I had to make a few tiny changes to make it work here:
- in the `common` module, regenerate code and dump API again
- in the `rest` module, dump API again
- make `DiscordComponent.description` both Optional and nullable [1]

You can see the details in the last three commits here:
https://github.com/ptitjes/kord/commits/feat/components-v2-didier/

[1] When deserializing the incoming messages in the Gateway, the `description` field sent back by Discord was `null`, and this was creating a deserialization error.

gdude2002 (changes requested):
I haven't had time to look into things in detail yet, but the presence of `AbstractMessageCreateBuilder` and `AbstractMessageModifyBuilder` is a strong indictment of Kord's development philosophy, imo.

This approach makes no sense from a logical viewpoint, and will significantly complicate things for API consumers - now instead of having a single base type that already worked everywhere, I need to mix between the interface and abstract type depending on whether I want to use components or not, and I need to go back over every single piece of code that currently uses the interface to make sure it doesn't now actually need the abstract type.

It's OK to break things when breakages are necessary! That's why we have versioning!

lukellmann:
> I haven't had time to look into things in detail yet, but the presence of `AbstractMessageCreateBuilder` and `AbstractMessageModifyBuilder` is a strong indictment of Kord's development philosophy, imo.

These types aren't new and weren't even touched in this PR, they have been there since #891

> This approach makes no sense from a logical viewpoint, and will significantly complicate things for API consumers - now instead of having a single base type that already worked everywhere, I need to mix between the interface and abstract type depending on whether I want to use components or not, and I need to go back over every single piece of code that currently uses the interface to make sure it doesn't now actually need the abstract type.

I couldn't find anything you can do with the `Abstract*` classes that you can't do with the interfaces. Did you need to use the `Abstract*` classes in some code? You shouldn't have to.

lukellmann (changes requested):

NotJansel:
tested around a bit and so far there was no error during the entire testing that could have been related to components v2

honkling:
Is this PR only targeting Components V2 in messages?
I tried out the branch for my discord bot because I'd like to use a string select component in my modal, but it seems like I can only attach action rows to my modal (which limits me to text inputs only), and labels aren't supported.

NotJansel:
@honkling to my knowlede yes, as the modal stuff got added pretty much right after this pr was ready for review/merge

honkling:
Hmm. Then, should components v2 support for modals be expected in a future PR? Or will this PR expand scope to accommodate modals?

PR #1010, opened 2025-04-26
Fix issues with application emojis

DRSchlaubi:


lukellmann (changes requested):
Thanks

lukellmann (approved):
Thanks for the quick fix!

lukellmann:
See https://github.com/kordlib/kord/compare/2002d0c0ec1a0336f589b68a0d2ca2af1ce73177...eaf5dbe1755a4b2005098c6a931465266fed0b65 for a cleaner diff of ABI changes.

PR #1011, opened 2025-05-23
Common ReactionEvent interface

MayaChen350:
I added an interface for `ReactionAddEvent` and `ReactionRemoveEvent` to remove code duplication and allow better Generics

Thing is, in one of my projects, I ended up doing a class for both of them because they're not bound by anything I could use as generic.

```kt
/** Abstraction class making using both type of events easier by sharing common properties and methods. **/
class ReactionEvent(
    private val addEvent: ReactionAddEvent? = null,
    private val removeEvent: ReactionRemoveEvent? = null
) {
    val emoji = addEvent?.emoji ?: removeEvent!!.emoji
    val guild = addEvent?.guild ?: removeEvent!!.guild
    suspend fun getMessage() = addEvent?.getMessage() ?: removeEvent!!.getMessage()
    suspend fun getUserAsMember() = addEvent?.getUserAsMember() ?: removeEvent!!.getUserAsMember()
    suspend fun getUser() = addEvent?.getUser() ?: removeEvent!!.getUser()
    suspend fun getRole(id: Snowflake) =
        addEvent?.getGuildOrNull()?.getRole(id) ?: removeEvent!!.getGuildOrNull()!!.getRole(id)
}
```

So basically, with that PR I'm grouping both Reaction events classes together to make things easier with an `interface`
Also I removed what I thought was unnecessary to keep after since both `ReactionAddEvent` and `ReactionRemoveEvent` had very similar code

Pretty simple, but efficient hopefully!


DRSchlaubi:
For some reason I can't approve this PR for the CI, but you need to run `gradlew apiDump`

DRSchlaubi:
Is it a binary incompatible change to move functions to a parent class?

MayaChen350:
I don't think so, but just to make sure, what do you mean by that?

DRSchlaubi:
that code compiled against a previous version will continue to work

and it is a change since the function signature changes, in this case there should be a deprecated version on the class, however that will make migrating kinda hard

/cc @lukellmann

MayaChen350:
Actually now that I think about it, I think the signatures would change since they're linked to the interface now but still are actually linked to the class somehow? Would they be the same if the methods would just be overriding the interface's instead?

core/src/commonMain/kotlin/event/message/ReactionAddEvent.kt
```diff
@@ -2,41 +2,26 @@ package dev.kord.core.event.message
...
-    public val userId: Snowflake,
-    public val channelId: Snowflake,
-    public val messageId: Snowflake,
-    public val guildId: Snowflake?,
-    public val emoji: ReactionEmoji,
+    public override val userId: Snowflake,
+    public override val channelId: Snowflake,
+    public override val messageId: Snowflake,
+    public override val guildId: Snowflake?,
+    public override val emoji: ReactionEmoji,
```
lukellmann, line 18: could remove `public` here

core/src/commonMain/kotlin/event/message/ReactionEvent.kt
```diff
@@ -0,0 +1,51 @@
...
+import dev.kord.core.behavior.MessageBehavior
+import dev.kord.core.behavior.UserBehavior
+import dev.kord.core.behavior.channel.MessageChannelBehavior
+import dev.kord.core.entity.*
+import dev.kord.core.entity.channel.MessageChannel
+import dev.kord.core.event.Event
+import dev.kord.core.supplier.getChannelOf
+import dev.kord.core.supplier.getChannelOfOrNull
+
+public interface ReactionEvent : Event, Strategizable {
```
lukellmann, line 15: should be `sealed` and override `withStrategy` to have the right return type

core/src/commonMain/kotlin/event/message/ReactionRemoveEvent.kt
```diff
@@ -2,58 +2,21 @@ package dev.kord.core.event.message
...
-    public val userId: Snowflake,
-    public val channelId: Snowflake,
-    public val messageId: Snowflake,
-    public val guildId: Snowflake?,
-    public val emoji: ReactionEmoji,
+    public override val userId: Snowflake,
+    public override val channelId: Snowflake,
+    public override val messageId: Snowflake,
+    public override val guildId: Snowflake?,
+    public override val emoji: ReactionEmoji,
```
lukellmann, line 14: same here

lukellmann:
> Is it a binary incompatible change to move functions to a parent class?

no, it isn't

MayaChen350:
Hi? This was kinda left out idk

NoComment1105 (approved):

NoComment1105 (approved):
Thanks! Glad we can get this in

MayaChen350:
Thank you! 

Thinking about it though, adding a common class for only that for a new release would be kind of weird... maybe (I originally made that PR because it was a fix to my specific case)

There's a chance other classes might benefit from a similar treatment (of course while not linking classes that wouldn't make sense together in a patern match)

I'll check soon and do a new PR if possible

NoComment1105:
I would agree and I look forward to reviewing what you come up with :)

PR #1023, opened 2025-09-25
Rename README.md to README.

RAY-D433:


DRSchlaubi:
I don't see a benefit of calling it README. Therefore this is closed

PR #1062, opened 2026-04-16
fix: make UnfurledMediaItem.contentType nullable

Lapis256:
I confirmed an issue where REST response deserialization can fail when sending a Component V2 message containing an image.

According to the official `discord-api-spec`, `UnfurledMediaResponse.content_type` is nullable; however, the current implementation defines `UnfurledMediaItem#contentType` as non-null, which is causing the error.

Reference:
https://github.com/discord/discord-api-spec/blob/eb04c76f20e0b160ce520ad793efcc9696e9bbd2/specs/openapi.json#L37743-L37784

This change updates `UnfurledMediaItem#contentType` to align with the actual response and the official specification.

## Observed response (excerpt)

```json
{
  "components": [
    {
      "type": 17,
      "components": [
        {
          "type": 9,
          "accessory": {
            "type": 11,
            "media": {
              "id": "...",
              "url": "...",
              "proxy_url": "...",
              "width": null,
              "height": null,
              "content_type": null
            }
          }
        }
      ]
    }
  ]
}
```

## Error (excerpt)

```text
kotlinx.serialization.SerializationException: descriptor for kotlin.String was not nullable but null mark was encountered
    at dev.kord.common.entity.optional.Optional$OptionalSerializer.deserialize(Optional.kt:164)
    ...
    at dev.kord.common.entity.UnfurledMediaItem$$serializer.deserialize(DiscordComponent.kt:156)
    ...
    at dev.kord.rest.route.ValueJsonMapper.deserialize(Route.kt:28)
    at dev.kord.rest.request.KtorRequestHandler.handle(KtorRequestHandler.kt:67)
    ...
```


NoComment1105 (approved):
Looks good to me! Thanks!

Issue #1013, opened 2025-07-21
Voice requires encryption


Wikijito7:
As stated [here](https://discord.com/developers/docs/change-log#voice-endtoend-encryption-dave-protocol), Discord will enforce E2EE when a user (or bot) is trying to connect to a Voice Channel. This would had a grace period on which clients could connect using the legacy non encrypted voice. It seems that the grace time is over and now Discord is returning an error when trying to connect to a VC without encryption.

It seems that we'll have to implement DAVE, Discord's Audio and Video End-to-End Encryption, if we want to connect to a VC. Is it on the roadmap?

I'll add the log aswell.

```log
[DefaultDispatcher-worker-6] ERROR dev.kord.voice.gateway.DefaultVoiceGateway - 
java.lang.IllegalStateException: Voice Gateway (689108711452442667) closed: 4016 Unknown encryption mode.
    at dev.kord.voice.gateway.DefaultVoiceGateway.handleClose(DefaultVoiceGateway.kt:233)
    at dev.kord.voice.gateway.DefaultVoiceGateway.start(DefaultVoiceGateway.kt:112)
    at dev.kord.voice.gateway.DefaultVoiceGateway$start$1.invokeSuspend(DefaultVoiceGateway.kt)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
    at kotlinx.coroutines.internal.ScopeCoroutine.afterResume(Scopes.kt:35)
    at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:101)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:829)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

Wikijito7:
[...]

PR #1063, opened 2026-04-27
Added DAVE support

Wikijito7:
Added DAVE support to Kord lib

This PR is a little sloppy, but it works. I've tested it and it seems to accept DAVE correctly.

I'm open to modify anything, feel free to add comments with whatever needs to be modified further :).

This PR is based on [Discord4J's PR](https://github.com/Discord4J/Discord4J/pull/1344) where they implemented DAVE support too.

Solves #1013.

### DIFF
--- a/voice/build.gradle.kts
+++ b/voice/build.gradle.kts
@@ -5,6 +5,10 @@ plugins {
     `kord-publishing`
 }
 
+repositories {
+    maven("https://maven.lavalink.dev/releases")
+}
+
 dependencies {
     api(projects.common)
     api(projects.gateway)
@@ -18,4 +22,14 @@ dependencies {
 
     implementation(libs.libsodiumBindings)
     compileOnly(libs.jna)
+
+    // libdave-jvm for DAVE protocol — versions defined in libs.versions.toml
+    implementation(libs.libdave.api)
+    implementation(libs.libdave.impl.jni)
+
+
+    testImplementation(libs.bundles.test.jvm)
+    testImplementation(libs.kotlinx.coroutines.test)
+    testImplementation(libs.mockk)
+    testRuntimeOnly(libs.bundles.test.jvm.runtime)
 }
--- a/changes.diff
+++ b/changes.diff
@@ -1,0 +1,926 @@
+diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
+index 1ad1ff84..18a71990 100644
+--- a/gradle/libs.versions.toml
++++ b/gradle/libs.versions.toml
+@@ -25,8 +25,8 @@ ksp = "2.3.5" # https://github.com/google/ksp
+ kotlinpoet = "2.2.0" # https://github.com/square/kotlinpoet
+ 
+ # tests
+-junit-jupiter = "6.0.2" # https://github.com/junit-team/junit5
+-junit-platform = "1.13.4"
++junit-jupiter = "5.11.2" # https://github.com/junit-team/junit5
++junit-platform = "1.11.2"
+ mockk = "1.14.9" # https://github.com/mockk/mockk
+ kbson = "0.4.0" # https://github.com/mongodb/kbson
+ 
+diff --git a/voice/src/main/kotlin/VoiceConnection.kt b/voice/src/main/kotlin/VoiceConnection.kt
+index c94e27dd..4761a29d 100644
+--- a/voice/src/main/kotlin/VoiceConnection.kt
++++ b/voice/src/main/kotlin/VoiceConnection.kt
+@@ -5,6 +5,7 @@ import dev.kord.common.entity.Snowflake
+ import dev.kord.gateway.Gateway
+ import dev.kord.gateway.UpdateVoiceStatus
+ import dev.kord.voice.dave.DaveProtocol
++import dev.kord.voice.dave.NoOpDaveProtocol
+ import dev.kord.voice.encryption.strategies.*
+ import dev.kord.voice.gateway.VoiceGateway
+ import dev.kord.voice.gateway.VoiceGatewayConfiguration
+@@ -77,7 +78,7 @@ public class VoiceConnection internal constructor(
+         nonceStrategy: @Suppress("DEPRECATION") NonceStrategy, connectionDetachDuration: Duration,
+     ) : this(
+         data, gateway, voiceGateway, socket, voiceGatewayConfiguration, streams, audioProvider, frameInterceptor,
+-        frameSender, dev.kord.voice.dave.NoOpDaveProtocol, connectionDetachDuration, nonceStrategy,
++        frameSender, NoOpDaveProtocol, connectionDetachDuration, nonceStrategy,
+     )
+ 
+     public constructor(
+@@ -86,7 +87,7 @@ public class VoiceConnection internal constructor(
+         frameInterceptor: FrameInterceptor, frameSender: AudioFrameSender, connectionDetachDuration: Duration,
+     ) : this(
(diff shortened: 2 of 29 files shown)


NoComment1105 (commented):
Thanks for the PR. I've had an initial, cursory look at this (as I've not much spare time right now) and on the face of it, it seems pretty good! Only things I'll flag up is that there is already an existing `feature/dave` branch, could you rebase onto that please? And there seems to be some random full qualifications of clases, can these be removed? They seem unnnecessary (example: `dev.kord.voice.dave.NoOpDaveProtocol` in `VoiceConnection.kt`

You say you "managed to make it work"... does that indicate it's not fully tested yet and need stability testing or just a way of words?

Also to merge this, commits must be verified, could you look into that please? Organisation restriction :P

Once again thanks, I look forward to your response :D

Wikijito7:
Hi! Thanks for the quick response!

I'll take a look at the build failing (seems weird, but I had some problems migrating the code from one computer onto this one, as the patched didn't want to _patch_ correctly). Also, the full qualification thingie is totally not intended, let me take a look at it tomorrow.

About the dave branch, ye, let me take a look at that branch and I'll try to rebase it. After I do that, would you like me to use that one as a base instead of main?

Also, I kinda expressed the "managed to make it work" bad. It works, although I found a bug after I created the PR, but I couldn't test it out as far as I would like, tbf, probably there might be some edge case or hidden bug that I need to refine, but the bot joins, plays and works with DAVE enabled.

Finally, about the verification thingie... _fair enough_. I'll take a look at it, I've changed computer recently and I didn't do it yet.

psmarko:
Hey, thanks for the PR.

I gave it my first lookover and I understand this is a primarily a JVM implementation. That is expected but it has introduced more JVM-only usages than before. It would ease my mind a little if the voice module were converted into a multiplatform module, [like all the others](https://github.com/kordlib/kord/blob/e3ceb0f138e1d64cfcb74aa7e3d8f7ff576212be/rest/build.gradle.kts#L2), and all the JVM stuff (dependencies and code) is thrown into the JVM target. I bring this up since Kord is multiplatform oriented, and this PR, as is, takes voice a few steps back in that regard.

There are also several new usages of the Java stdlib for crypto and concurrency. This is not ideal.

I have not yet reviewed much logic nor other design.

Would you want to work on making this multiplatform friendly? If not, I can look into it myself. 
I appreciate the work here, cheers.

voice/build.gradle.kts
```diff
@@ -18,4 +22,19 @@ dependencies {
...
+
+    // libdave-jvm for DAVE protocol — versions defined in libs.versions.toml
+    implementation(libs.libdave.api)
+    implementation(libs.libdave.impl.jni)
+
+    // Native libraries for libdave — include all platforms, loader picks the right one at runtime
+    runtimeOnly(libs.libdave.natives.darwin)
+    runtimeOnly(libs.libdave.natives.linuxX64)
+    runtimeOnly(libs.libdave.natives.linuxAarch64)
+    runtimeOnly(libs.libdave.natives.winX64)
```
psmarko: Kord should not bundle native artifacts. It is the user that's responsible for declaring platform-specific natives. These should rather be mentioned in the voice README.

```suggestion
```

psmarko (changes requested):

voice/build.gradle.kts
Wikijito7: Yea, fair. Gonna take a look and add it on the readme! Let me also see if this would break something of the current implementation.

Wikijito7:
Hi psmarko! Thank you for the review!
> I gave it my first lookover and I understand this is a primarily a JVM implementation. That is expected but it has introduced more JVM-only usages than before. It would ease my mind a little if the voice module were converted into a multiplatform module, [like all the others](https://github.com/kordlib/kord/blob/e3ceb0f138e1d64cfcb74aa7e3d8f7ff576212be/rest/build.gradle.kts#L2), and all the JVM stuff (dependencies and code) is thrown into the JVM target. I bring this up since Kord is multiplatform oriented, and this PR, as is, takes voice a few steps back in that regard.

Yep, currently this supports JVM-only. I have to be honest, this is the only library that I found that could work with Kotlin's JVM capabilities, but I didn't take a deep look. If you know of any Kotlin-first library that wraps the C++ binary, I wouldn't mind changing into it. I tried making it myself but couldn't dedicate that much time with work-stuff.

> There are also several new usages of the Java stdlib for crypto and concurrency. This is not ideal.

Those usages are related with how the libdave-jvm work. I can change it, but I might need a little bit of guidance of which one you might like more.
 
> Would you want to work on making this multiplatform friendly? If not, I can look into it myself. I appreciate the work here, cheers.

I can try it, why not! Tbh I haven't worked on a Kotlin Native project yet, but I wouldn't mind trying out. Let me know if you have any preference regarding the previous points :).

psmarko:
> this is the only library that I found that could work with Kotlin's JVM capabilities

It is one of the few, yes. My concern was there is little multiplatform scaffolding for future platform implementations. The roadmap for voice includes providing our own libdave mpp wrapper, so I wouldn't worry about the actual mpp implementation. Though I do want to make sure it is possible to add one in the future without breaking API (where possible).

> Those usages are related with how the libdave-jvm work.

While implementing the last encryption mode, I ran into AES256 not being exposed in the libsodium bindings we use for encryption. Thus I left it unimplemented. Is AES256 needed for DAVE? If so, I don't mind if the AES implementation stays for the time being, especially when not exposed publicly- just throw it into the JVM target.

`java.io.Closeable` can be replaced with a `close` function in `DaveProtocol`. This is how it is done in other places.

For concurrency usages, that's fine if it stays in the JVM target.

That's all the usages I spotted.

Just to reiterate, the goal is just to keep the voice common module platform-agnostic. A JVM implementation using libdave-jvm (`LibdaveDaveProtocol`) in the JVM target is fine.

Let me know if there's any issues turning it into a multiplatform module, no clue if there will be difficulties (though I would hope not).

Cheers.

Wikijito7:
Hi again, 

I've done a partial migration of the module to a kmp-format, jvmMain, module. So far, it works. The thing is, I'm having a issue with voice on a particular flow, that may or may not happen but definitely is a bug. When the bot is kept alone on a VC playing (the other user leave) and rejoins, the bot voice stops working, as the epox changed. Moving the bot or reconnecting solves the error.  This is a new error introduced with how DAVE works.

I'll take a look at this error, try to fix it and then I'll update the PR with these changes, that way you can re-review it and tell me if this new reorder is closer to what you expect as a kmp-ready module.

  > The roadmap for voice includes providing our own libdave mpp wrapper, so I wouldn't worry about the actual mpp implementation.

 Got it, then we can work with this jvm-only variant for the mean time until your multiplatform-ready variant is finished, then we can migrate to that one.

Wikijito7:
Sorry guys for the long-pause. Work's been crazy, today I remembered this was still undone.

I've uploaded some of the changes we talked, but the KMP migration was left undone. For the time being, I cannot dedicate the time I would like to finish everything perfectly.

Right now, this version is working fine. Probably part of the code can be done better, but it works as is.

Again, thank you for reviewing the previous version. Feel free to edit or comment whatever could be done better, and I'll try to take a look at it, although probably you'll be faster editing it by yourself if needed.

See you in the comments

voice/build.gradle.kts
Wikijito7: Removed, tested on my local bot, adding these libraries as dependencies on the consumer-side keeps everything working.

psmarko:
Just wanted to note that I've seen your update. Will find time to review in the near future, thanks.

PR #1072, opened 2026-08-08
allow nullable DiscordChannel.applicationId

pranav-narayanan:
In response to this breaking change: https://docs.discord.com/developers/change-log#channel-application_id-is-now-nullable

Discord made the `application_id` field in a channel object nullable, so in two places the type needed to be updated from `OptionalSnowflake` to `OptionalSnowflake?`.

Have tested deserialization with missing, null, and present `application_id` locally, they all now work.


NoComment1105:
Please run `./gradlew dumpLegacyAbi`

DRSchlaubi (approved):

